PR #9, opened 2012-11-28
Allow forceY on multiChart

pkendall64:
This patch allows you to call forceY on the lines in multiChart which means each Y axis can have the same bounds, or at least both start at 0.


liquidpele:
PR has conflicts, considering the age I doubt it'll be fixed so I'll close this and someone will have to open a new one.

PR #15, opened 2012-12-17
Adding d3.event to scatter event listeners

jonaldomo:
Adding d3.event to scatter event listeners to maintain consistency between models.  In the historical bar chart they are included, but not the line which uses the scatters events.


robinfhu:
This is not necessary because d3.event can be accessed in any function downstream.

PR #40, opened 2013-01-22
Fix bug of updateFisheye when you filtered out some data

awakia:
Hi,

This patch fixes a bug when you filtered out some datapoints by clicking legends, and moving mouse around the chart.


robinfhu:
This code has made it into scatterChart.js

PR #46, opened 2013-01-27
Add nvd3 scope to CSS rules for generic tags

csulok:
scope the svg css to only nvd3 charts


robinfhu:
This has been implemented in the latest nvd3

Issue #98, opened 2013-03-25
Creating pie chart not working - Uncaught TypeError: Object #<Object> has no method 'map' 


gaqzi:
Hi,

This line [`state.disabled = data[0].map(function(d) { return !!d.disabled });`](https://github.com/novus/nvd3/blob/a52eddf761e80b4bee8da23cbe3fecfe3189c4ca/src/models/pieChart.js#L60) introduced by commit a52eddf761e80b4bee8da23cbe3fecfe3189c4ca isn't letting my pie charts render. 

If I comment that line out and run the following code it'll create the charts as it should. (Basically the example from nvd3.org)

My code:

    function exampleData() {
      return [
      {
        key: "Cumulative Return",
        values: [
          {
            "label" : "CDS / Options" ,
            "value" : 29.765957771107
          } ,
          {
            "label" : "Cash" ,
            "value" : 0
          } ,
          {
            "label" : "Corporate Bonds" ,
            "value" : 32.807804682612
          } ,
          {
            "label" : "Equity" ,
            "value" : 196.45946739256
          } ,
          {
            "label" : "Index Futures" ,
            "value" : 0.19434030906893
          } ,
          {
            "label" : "Options" ,
            "value" : 98.079782601442
          } ,
          {
            "label" : "Preferred" ,
            "value" : 13.925743130903
          } ,
          {
            "label" : "Not Available" ,
            "value" : 5.1387322875705
          }
        ]
      }
      ];
    }
    
    nv.addGraph(function() {
      var chart = nv.models.pieChart()
         .x(function(d) { return d.label })
         .y(function(d) { return d.value })
         .showLabels(true);

        d3.select('svg')
          .datum(exampleData()) 
          .transition()
          .duration(1200)
          .call(chart);

        return chart;
    });

liquidpele:
Should be fixed in latest :)

PR #106, opened 2013-04-05
Fix Nvd3 Issue #98 Object #<Object> has no method 'map' on pieChart

corneadoug:
Change data[0] to data for data.map for the pieChart


robinfhu:
This is really out of date.

PR #120, opened 2013-04-22
"JS clean + add extra serie to test multi series on linePlusBarWithFocusChart

areski:
...rWithFocusChart

This also show that there is some filter issues when we have multiple series.


liquidpele:
Not valid PR anymore since that file was removed.

PR #165, opened 2013-06-18
Bug Fix - Prevents tooltip from showing with Date data type in line charts

danramosd:
Tooltips do not work when using Date objects for X or Y values with line charts.  The root of the problem comes from trying to add a "jitter" to these points.  In firebug I receive the following warning:

`Unexpected value NaN parsing cx attribute.`

My fix checks for a Date type first before applying the jitter which removes the NaN warning and fixes the tooltip.

(Sorry for the large chunk of changes in this commit I couldn't figure out why git read the change this way.  I only made changes to lines 10,031 - 10,035 )

### DIFF
--- a/nv.d3.js
+++ b/nv.d3.js
@@ -10028,8 +10028,11 @@ nv.models.scatter = function() {
                 // *Injecting series and point index for reference
                 /* *Adding a 'jitter' to the points, because there's an issue in d3.geom.voronoi.
                 */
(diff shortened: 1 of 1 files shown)


liquidpele:
I think this may already fixed by adding the jitter after the x() and y() calls instead of before, but I don't have any example to test against so closing until we get that based on latest development branch.

PR #172, opened 2013-06-27
Fixes #89 (scatterPlot voronoi issue)

### LINKED ISSUE #89: [Scatter] Browser reports error on mouse over
Uncaught TypeError: Cannot read property 'values' of undefined nv.d3.min.js:5564
pointPaths.on.on.on.series nv.d3.min.js:5564
i

                if (needsUpdate) return 0;
                var series = data[d.series],
                    point  = series.values[d.point];

FellowMD:
For a scatter plot with less than 4 points, the voronoi stuff doesn't work (apparently), so a few extra points are added.
But this should only happen when there are less than 4 points, so I adjusted the if condition.
Fixes #89

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -206,8 +206,8 @@ nv.models.scatter = function() {
           }
 
 
-          if(vertices.length) {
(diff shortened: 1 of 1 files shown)


liquidpele:
Closing pending reproduce steps or example.

PR #178, opened 2013-07-03
Adding Ordinal scales to scatterChart

saliksyed:
I've added the ability to have ordinal scales. The main modifications are that we need to conditionally check if the data is ordinal then build the separate scales accordingly (this is within scatter.js). Within scatterChart.js we need to re-assign the x and y scales after the data is set within the internal scatter object since the axes are regenerated if ordinal data is detected.


robinfhu:
Changeset is outdated, and I know there are workarounds to getting 'ordinal' scale behavior for scatter charts.  Ping me if you want to know how it can be done.

PR #204, opened 2013-08-03
Added url support to discreteBarChart point

campey:
If point has url, add 'a' container with 'xlink:href' within 'g'
(seems nicer to change 'g' to 'a' but this breaks resizing);
Move fill/stroke onto rect as they didn't get the 'g' style with 'a'
inbetwen;
Updated example to include some urls


liquidpele:
Still broken (see Robin's comment) and really stale, so closing.

PR #217, opened 2013-08-13
[multibar] Added bars transition 'end' event trigger

ghost:


liquidpele:
This user is deleted now and it's too stale to merge, so just closing this.

PR #235, opened 2013-08-24
Expose stack offset

### LINKED ISSUE #225: Stacked Column Percentage
It would be swell if the multiBarChart could display stacked columns with each column's segments scaled so that they add to 100%.

This image shows how that might look:
http://blog.visual.ly/wp-content/uploads/2012/08/StackedPercent.png?547b7b

rustybailey:
Fix #225 -
Thankfully, d3 has a percentage-based stacked column baked into the stack layout in the form of `stack.offset`. All you need to do is set the offset to expand and you get a chart like this:

screen shot 2013-08-24 at 8-24-13 7 12 01 pm

This PR exposes the offset options. The only thing remaining is to change the yAxis format to %, but I figured that it was best to leave the format up to the user, but feel free to add if you think it should be the default format.

### DIFF
--- a/src/models/multiBar.js
+++ b/src/models/multiBar.js
@@ -16,6 +16,7 @@ nv.models.multiBar = function() {
     , forceY = [0] // 0 is forced by default.. this makes sense for the majority of bar graphs... user can always do chart.forceY([]) to remove
     , clipEdge = true
     , stacked = false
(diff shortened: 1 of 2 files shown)


ankitav14:
I want to understand, how to use stack.offset.
As in where should i include that?

PR #260, opened 2013-09-08
Expose the 'controls' property on more chart models

hasaki:
The 'controls' object was already exposed on some models, this pull request exposes it on the others that have a controls object.

This would, for example, allow you to turn off right alignment (which keeps the controls from running into the legend) when dealing with a small chart.


liquidpele:
In dev branch already now.

PR #268, opened 2013-09-15
Rename Gruntfile to follow grunt convention

robertd:


ryan-ju:
Just want to point out that `Gruntfile` is no longer case sensitive, according to https://github.com/gruntjs/grunt/issues/685.  The only care you should take is to update `node-glob` to a newer version.

PR #269, opened 2013-09-16
Unifying Tooltip event functions + ScatterChart Tooltips

corneadoug:
I decided to arrange the events in order to group them the same way between models (Chart's scope and out of chart scope) when possible.

Also, I unified the type of Tooltip between scatterChart and scatterPlusLineChart.

Don't know if it was on purpose but in the examples/ directory, scatterChart.html line 76.
It was changing the look of the tooltip. Good example for that, but might lead to mistake when testing the charts.


robinfhu:
Too outdated.  We got rid of scatterPlusLineChart.

PR #270, opened 2013-09-16
Fix legend texts width computation

Marsup:
Hi,

On IE, Chrome and Firefox, `getComputedTextLength` seems to return 0 when the element is not visible, so I copied the workaround from the catch to this case.


robinfhu:
This logic has actually been implemented in legend.js already.

Marsup:
It wasn't at the time of that PR.

liquidpele:
I implemented quite a few of the pull requests manually in nvd3-community... thought I commented on all the ones I did though, sorry :)

PR #271, opened 2013-09-16
Fix scatter single point class removal

Marsup:
Hi,

I encountered this bug when the linechart was fed with a single value, and then updated it with more.
The nv-single-point class would stick and the linechart would have visible dots all over it.
I think it's the correct way to fix it but I have a doubt whether I should apply the class to `g` or something else, so advice welcome.


liquidpele:
Fixed using classd on the main wrap to avoid CSS changes.

Issue #291, opened 2013-09-30
LineChart: Only one dataset is shown when the datasets have the same key (or no key), tooltips are working though.


TN1ck:
When the data looks like this: `[{values: [...]}, {values: [...]}]`
or like this: `[{key: "key1," values: [...]}, {key: "key1", values: [...]}]`

Only one of the two datasets is shown. Here is a jsfiddle, you would expect the two charts would behave the same.

http://jsfiddle.net/3s7Te/1/

I've tested this with the newest version as well - same problem. I've only encountered the problem with lineCharts, barCharts work fine.

I've found the problem:
https://github.com/novus/nvd3/blob/master/src/models/line.js#L107

     var groups = wrap.select('.nv-groups').selectAll('.nv-group')
          .data(function(d) { return d }, function(d) { return d.key });

you can fix it with for example:

     var groups = wrap.select('.nv-groups').selectAll('.nv-group')
         .data(function(d) { return d }, function(d, i) { return d.key + i });

liquidpele:
I updated the jsfiddle to 1.7, it looks like it's still an issue... the "key" attribute of the data is the same for both series which causes the problem... seems silly but probably should be able to handle that. 

http://jsfiddle.net/3s7Te/9/

liquidpele:
should be fixed now.

liquidpele:
Had to revert due to https://github.com/novus/nvd3/issues/803, will look at robin's suggestion.

PR #292, opened 2013-09-30
Added index to key, so datasets with same keys will be shown at the same...

TN1ck:
... time

This will fix Issue 291: https://github.com/novus/nvd3/issues/291


liquidpele:
Manually applied this fix.

liquidpele:
Actually I reverted this due to https://github.com/novus/nvd3/issues/803

PR #296, opened 2013-10-02
replace axis d3 scale when scatterChart.js scale is replaced

squince:
Minor tweak to ensure that when replacing the scale for the scatterChart that the axis scale is also replaced.  Now mirrors ScatterChartWithLine.js

### DIFF
--- a/src/models/scatterChart.js
+++ b/src/models/scatterChart.js
@@ -151,6 +151,8 @@ nv.models.scatterChart = function() {
 
       //------------------------------------------------------------
       // Setup Scales
+      x = scatter.xScale();
+      y = scatter.yScale();
 
       x0 = x0 || x;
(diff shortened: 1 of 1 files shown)


robinfhu:
Not sure what problem this is supposed to solve. Can you explain in detail?

squince:
Wow, this is really old.  I've long since been using our own modified version of nvd3 in our dashboard.  As I recall, I had tried to use a feature to change the scale on a chart and it didn't behave the way I expected and the way other charts behaved.  As I recall, this was a bug that needed fixing.  I'd have to do some digging to remember exactly what the symptom was.

Your call on whether you want to take the time to look into this and merge it in.

robinfhu:
@squince cool.  If you've solved it in your own fork that is fine for now.  Closing.

PR #312, opened 2013-10-14
Create areas with upper/lower domain

jimerino:
I've added the possibility to create an area with a upper/lower domain. The lower value of the existing area is always the same for all the serie. Now you can define a lower value for the area for each point:

ae551a6a-3369-11e3-824b-dc80074f58c5

To activate this option only need to pass a 2-value array to the y value of the area:

sin.push({x: i, y: i % 10 == 5 ? null : [Math.sin(i/10),Math.sin(i/10)/2]});

That don't affect to the existing line or area. An area with a single y value is represented as usual with a same lower value for all the area (normally 0 base).

### DIFF
--- a/examples/lineChart.html
+++ b/examples/lineChart.html
@@ -41,8 +41,7 @@ svg {
 </div>
 
 <script src="../lib/d3.v3.js"></script>
-<script src="../src/intro.js"></script>
-<script src="../src/core.js"></script>
+<script src="../nv.d3.js"></script>
 <script src="../src/tooltip.js"></script>
 <script src="../src/utils.js"></script>
 <script src="../src/interactiveLayer.js"></script>
--- a/examples/lineChart.html
+++ b/examples/lineChart.html
@@ -41,7 +41,8 @@ svg {
 </div>
 
 <script src="../lib/d3.v3.js"></script>
-<script src="../nv.d3.js"></script>
+<script src="../src/intro.js"></script>
+<script src="../src/core.js"></script>
 <script src="../src/tooltip.js"></script>
 <script src="../src/utils.js"></script>
 <script src="../src/interactiveLayer.js"></script>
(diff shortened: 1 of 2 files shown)


liquidpele:
I'm curious what the use case for this is?

robinfhu:
I know what this change is trying to do.  Normally with the 'area' feature of lineCharts, the area extends all the way to the y-axis.  There is no way to make it customized like in the graphic.

The solution presented is not something I am a fan of.  It involves setting an array as the data Y value, like this:

      {x: 1,  y: [100,200]}

This is worth thinking about though.

liquidpele:
Agreed on not liking the y value being an array for a normal line, but it could be neat as a new model that shows the difference between two points like this:   

http://bl.ocks.org/mbostock/3894205   or maybe this   http://bl.ocks.org/mbostock/3884914

robinfhu:
Yeah those are some good examples.  I think the code we have for areaPaths in line.js can be rewritten to support this.

liquidpele:
@jimerino I looked at this again, and I think I'd merge this in if you made one change to it... instead of making y an array, can you modify so that the values just includes an extra attribute called ymin, like this?

    {x: 4, y: 6, ymin: 3} 

and of course ignore ymin if the series doesn't have the 'area' attribute set to true.    Would that work for you?   I just see that as being much more backward compatible and easier to read in the code. 

Also, it looks like you need to merge our development branch back into your branch and resolve any conflicts since github can't auto-merge this anymore.  Shouldn't be anything big though.

liquidpele:
Oh, actually you need to send this pull request to our development branch anyway, this one was sent to master so we can't merge this one... all pull requests have to go to development so that master stays stable :)

liquidpele:
Closing in favor of a newer PR that fixes the bugs in the last newer PR :p

PR #313, opened 2013-10-14
Custom valueFormatter when interactive guideline is used.

gardenermike:
The interactive guideline feature previously forced display of the value in the tooltip to fit a fixed convention, which was inconsistent with the configuration available elsewhere for tooltips and the y axis.  A configuration option was added to allow a custom valueFormatter to be passed, which we use to display the original data value rather than the percentage passed in the data.


robinfhu:
It's a little stale, but this change interests me.

liquidpele:
Really stale... good idea, but it would have to be rewritten....

This seems related:
https://github.com/novus/nvd3/issues/428

Ideally we could probably have a valueFormatter that worked either tooltip, and by default just used the tick formatting.

liquidpele:
All tooltips us the valueFormatter option now, on the tooltip object itself, and it uses the tickFormat of the axis as the default.  Cheers.

PR #324, opened 2013-10-22
Fixes issue 322 (scatterPlusLineChart legend cut off issue)

olemi:
I came up with this fix after comparing the source code of scatterChart.js and scatterPlusLineChart.js.


liquidpele:
Stale and description doesn't explain what the issue was... closing.

PR #325, opened 2013-10-22
Issue 323

olemi:
Basically the same code as in scatterChart.js


liquidpele:
too stale, closing this.

PR #329, opened 2013-10-29
Translate legend and separate tooltip format in line chart and stacked area chart

sulmanen:
- allow setting legendTranslateX and Y on
  -line chart
  -stacked area chart

- allow setting x and y value tooltip formatting independent of axis format for
  - line chart
  - stacked area chart

I've pushed these to production and they work (not on public web)


liquidpele:
I don't really like this implementation, I'm more for making legend use the width if it needs, or specifying the location via the 'top', 'right', stuff that's already in the code via the legendPosition option on pieChart.

PR #336, opened 2013-11-04
Add interative guideline for lineWithFocusChart

cjgu:
Also add test page for lineWithFocusChart.


liquidpele:
Too outdated... if you can re-submit based on the latest dev branch then we'd have another look!

PR #355, opened 2013-11-20
allow specifying the stroke-dasharray line style

snopoke:
See examples/lineChart.html for example.

dashed and dotted lines


liquidpele:
You can already do this with just CSS.... e.g on lineChart example:

.nv-series-4 { stroke-dasharray: 7,7; }

Issue #253, opened 2013-09-04
[multiChart] - multiple issues


AaronKulick:
I am highly dependent on the muliChart model as it is the most reliable way to get reliable graphs of which contain multiple bars and lines.  For simplicity and reusability, I use this model even when there only 1 series of a single type (line, area, bar) because the x-axis major ticks are not implemented the same across chart models.  

I believe it unfortunate, that this particular chart model has not received as much polish as the rest of the models and would love to seem the enhanced functionality and polish the basic/multi line and bar models have gotten.

Major features/issues off the top of my head:

* **hide/show type "line" via legend** - when on the right y-axis the line object is not properly disabled (the right y axis is hidden correctly, but the line remains).
* **graph components** - not styled because the model does not use the correct CSS classes (x.axis instead of nv-x.nv-axis).  I can provide a fix for the x and y axises.
* **forceY/X** - Adding forceY has been identified and (a workaround exists though not properly implemented as an option).
* **noData** - When no data is present show the no data message.  The fix for this is simple which I have implemented and I can provide a patch (which was borrowed one of the other models).
* **right and left y-axis major.tick lines/marks** -  these do no align or use the same number of ticks which causes multiple y-axis major.tick lines to be rendered on graph making it almost unreadable. My current workaround is to simply disable the right y-axis major.tick lines via CSS.  It would be ideal if the number of ticks was the same for the right y-axis as the left (or as an option).
* **x-axis major tick lines do not properly align** - when using an x-axis of dates the actual point on the line does not intersect with the x-axis major.tick line.

liquidpele:
This lists several issues... I made a lot of fixes to multichart, it shoudl be working pretty well now, some of the other issues were handled by other pull requests as well.   Closing and individual issues should be created for anything still broken, thanks!

PR #363, opened 2013-11-23
Multichart allow to hide (by disabled) all series of lines and multibar chart type

michalkutil:
Allow to hide/show all series of line and multibar chart types in multichart. Partially solves Issue #253, item: "hide/show type line via legend".


liquidpele:
This should be fixed already I think, if it's talking about some lines now hiding when you disable them in the legend on multichart.

PR #370, opened 2013-11-29
Allow specifying line points in scatterPlusLineChart.

nightscape:
Instead of providing an intercept and a slope you can specify arbitrary points which will be connected by lines.
For an example see test/ScatterChartTest.html


liquidpele:
scatterPlusLine was merged into just scatterChart

PR #371, opened 2013-12-03
Add interpolation option for lines in multiChart

AliAlbazaz:


liquidpele:
This is already implemented in latest.

PR #373, opened 2013-12-04
minor change to css to prevent breaking non-nvd3 SVG elements on the web page.

ktarplee:
nvd3 should not change the global SVG style but rather the nvd3 specific SVG style


liquidpele:
Already fixed in latest.

PR #381, opened 2013-12-12
Chart's 'controls' sub-component exposed

eugene-stepanov:
Chart's 'controls' sub-component exposed for several charts to be able
to override controls' properties and to be consistent with the other
charts all over the project


liquidpele:
Manually applied this, closing.

PR #382, opened 2013-12-12
Allow nvd3 to be installed via NPM and not just bower

nowells:
We use NPM as our package manager (moved away from bower) and it would be great to be able to NPM install nvd3.


liquidpele:
Looks like our package.json has most of these changes already, so I think it should be working?   Please check the latest development branch!

PR #387, opened 2013-12-16
left, bottom and right label positioning (+ some bugs)

sbuchok:
The ability to position the left label text is there through axisLabelDistance, however the right and bottom labels for a chart do not have this functionality. It has now been added (used the same way for all three)

xAxis.axisLabelDistance...
yAxis.axisLabelDistance...
y1Axis.axisLabelDistance...
y2Axis.axisLabelDistance...

I also came a cross a few errors while dealing with multiple charts on a page where sometimes (not all the time) there would be an error on mouse over of sections of a chart. This was due to the d3.event being undefined and then later it using d3.event.pageX or d3.event.pageY. I threw in checks to see if d3.event was truethy and if so continue on, if not, do nothing.

If you have any questions or comments, please let me know.

### DIFF
--- a/nv.d3.js
+++ b/nv.d3.js
@@ -1147,7 +1147,7 @@ nv.utils.optionsFunc = function(args) {
           var w = (scale.range().length==2) ? scale.range()[1] : (scale.range()[scale.range().length-1]+(scale.range()[1]-scale.range()[0]));
           axisLabel
               .attr('text-anchor', 'middle')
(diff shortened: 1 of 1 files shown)


liquidpele:
this PR edits the built file and not the source, please check latest development branch and send a new PR based on that if this issue isn't already fixed.  Also please see https://github.com/novus/nvd3#a-few-rules-for-pull-requests

Cheers!

PR #396, opened 2013-12-27
Made windowResize use addEventListener over window.onresize

jgrund:
With the current implementation, windowResize can only be used to add
an event listener not remove one. You can only remove all listeners by
setting ```window.onresize = null;```

This patch uses ```window.addEventListener(‘resize’, fun);``` instead so an
individual function can be removed when needed.

The basis for this is I'm trying to use nvd3 in a single page app, and would like to ensure event listeners are removed when I navigate to another page to avoid memory leaks.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -27,17 +27,22 @@ nv.utils.windowSize = function() {
 
 
 
-// Easy way to bind multiple functions to window.onresize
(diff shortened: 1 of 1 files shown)


moneytree-doug:
:+1:

liquidpele:
Already using addEventListener in latest, cheers.

Awakening-j:
+1

PR #398, opened 2013-12-28
stylesheets in less instead of css

capaj:
because it is less of a pain to write less.
I have modified gruntfile accordingly, but I don't know how to change makefile, I have no knowledge about how to write makefiles. SO if you would merge it, modify makefile accordingly


liquidpele:
The nvd3 css file is way to small/simple to need LESS or SASS imho.

joshjordan:
I agree with you @liquidpele

robinfhu:
At 600+ lines we can consider at least breaking up the CSS file and concacting it during the grunt release process.  I do it for a lot of my projects.  I'm thinking we can have 'pie.css', 'line.css', etc, for the various chart specific stuff.

capaj:
@robinfhu :+1:

liquidpele:
Yea that seems fine, I'll go ahead and break it up then.

PR #400, opened 2014-01-02
fix: legend key function overriden by pie.x();

tmszdmsk:


robinfhu:
We need to use legend.key(pie.x()), because pie chart data is structured in such a way that each data point represents a legend item.  legend.key() by default is this kind of function:

> function(d) { return d.key;}

But what if your pie data looks like this:

> [{ label: "series1", value: 50}, {label: "series 2", value: 50]

Then the legend.key() function must actually be:

> function (d) { return d.label}

Pie charts are kind of a weird case.

PR #404, opened 2014-01-07
Fixed a bug with scatterCharts not respecting changes to scales

kevinjdinicola:
Setting a scale/range/domain (say, an ordinal scale) on the scatterChart after creating it would set the xScale and yScale properties of scatter, but x and y in the chart() function of scatterChart are never updated with the latest scales that had been set on scatter.  Instead x and y retain the initial values given to them upon creation (linear scales typically).  I moved the code around so the scales behave more like that of line charts, and are pulled from scatter.xScale() / scatter.yScale() before drawing.


liquidpele:
This pull reqest has a ton of other stuff in it, is out of date, is to master instead of development, and includes the built files instead of only the source files.  Closing, and please re-submit any pull requests (one for each feature) base on the latest development branch, thanks!

PR #406, opened 2014-01-08
sparklinePlus chart Model - add logic to use showValue flag

bheithaus:
Looks like this if statement was missing to make use of the showValue flag, so I just added it in.

~ Thanks


liquidpele:
Manually applied in https://github.com/novus/nvd3/pull/890, closing this one.

PR #410, opened 2014-01-10
fix highlightZero: Firstly, previously d.__data__ was checked for zeros,...

benzitohhh:
... but the data is actually bound directly to d. Secondly, previously the css just set zeroline opacity higher, but this alone is not sufficient to visually distinguish it from the other lines, hence we explicitly set its color to black.

### DIFF
--- a/nv.d3.css
+++ b/nv.d3.css
@@ -235,6 +235,7 @@ svg .title {
 .nvd3 .nv-axis .zero line,
 /*this selector may not be necessary*/ .nvd3 .nv-axis line.zero {
   stroke-opacity: .75;
+  stroke: #000; /* Color zero-line so that it is distinguishable from the other lines. */
 }
 
 .nvd3 .nv-axis .nv-axisMaxMin text {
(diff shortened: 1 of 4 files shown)


liquidpele:
I'm not sure what axis.highlightZero even does, I tried to disable it on an example and didn't see any difference...

benzitohhh:
I guess that when axis.highlightZero==true, the zero axes should be distguised from the other axis lines (i.e darker color, or more bold etc).

But as you notice, the support is really inconsistent...

liquidpele:
@robinfhu can you comment on that option?

robinfhu:
You need to create a chart that has points around zero to really see this. But it's quite simple, all it does is make the axes at the zero point darker.

robinfhu:
This PR looks fine, just need to merge it to development.

robinfhu:
Attaching screenshot to show what I mean.
zero line scatter

Also, I recommend keeping the CSS change out.  The "zero" class is automatically applied, so if you want those darker lines override it in your own project.  

By default the zero lines don't stand out.

robinfhu:
I'll take care of merging this in, I think it's a good fix.

robinfhu:
Opened PR: https://github.com/novus/nvd3/pull/813

PR #422, opened 2014-01-22
Can optionally use an array of values in tooltip series content

Sigfried:
If tooltip `series[item#].value` is an Array, each value element gets its own `<td>`.


liquidpele:
I'm not sure how this would be used, when would value ever be an array?   

Also, the pull request is too out of date to merge and is to master instead of the development branch, so I'll close it but please feel free to re-submit based on the latest development branch if you think it's a worthwhile feature, cheers!

PR #426, opened 2014-01-28
scatterChart and sparklinePlus bug fixes

vkapartzianis:
I've fixed two bugs in the scatterChart model: data won't filter correctly when you enable fisheye, and pointer-events aren't initialized.

Fiddle http://jsfiddle.net/qggT8/1/ displays those problems and fiddle http://jsfiddle.net/qggT8/3/ has them fixed. Please contact me if you need any further details and thanks for your work.

### DIFF
--- a/src/models/scatterChart.js
+++ b/src/models/scatterChart.js
@@ -167,7 +167,7 @@ nv.models.scatterChart = function() {
       var g = wrap.select('g');
 
       // background for pointer events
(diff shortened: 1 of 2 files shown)


liquidpele:
fisheye was removed so this is too out of date and no longer valid, please check out the latest stuff and re-submit any pull requests you think are still valid with the changes, thanks!

PR #431, opened 2014-01-30
fix multiChart axis classes to match other models

comboy:
Would be nice to rename <code>yAxis1</code>/<code>yAxis2</code> to <code>y1Axis</code>/<code>y2Axis</code> to match linePlusBarChart, altough it's an API change that could break things for some people.

Two maybe a little related questions if you have time:
* what's the the tests status? <code>grunt lint</code> is throwing a lot of red at me but it could be my badly setup env
* are there currently any specific plans on how code could by DRYed a bit? I know it's difficult, especially keeping it simple and easy to dive in, but at least some parts seem to be begging for it.


liquidpele:
I manually applied this and tested the chart to make sure it still worked properly.

PR #434, opened 2014-02-02
Added text colour  so that tooltips display correctly when on pages with white text.

robingrindrod:
If a chart is displayed on a page with white text, the tool-tips become unreadable. I have added two lines of CSS to ensure that the tool-tip's text is displayed in black so that it can be read.


liquidpele:
Looks to be still an issue, merging.

PR #438, opened 2014-02-04
Updated to the latest version of d3.js and added support for AMD

### LINKED ISSUE #405: incompatible with latest version of d3
I had problems using the latest version of d3 with the latest version of nvd3, other people have similar problems which might be caused for the same reason.

Bs using the version included in the **/lib** directory the java script error disappeared.

http://stackoverflow.com/questions/17624111/nvd3-line-chart-uncaught-typeerror-cannot-read-property-x-of-undefined/20994982#20994982

rimunroe:
Fixes #405 and allows this to be used with the latest version of d3 and requirejs. Since d3 is now an amd module, it wouldn't export d3 to global if being run with requirejs, which would break nvd3.


liquidpele:
This is too out of date to merge, also it is to master and should be to the development branch and edits the built file instead of only the source files.   Closing for now, and feel free to re-submit this PR.   Sorry for the huge wait in response btw... we're cleaning up the backlog now...

PR #439, opened 2014-02-07
Remove chart when no data is available

HenningM:
I'm using nvd3 to visualize real time data, and I've noticed that the chart is not removed when there is no data available.
nvd3_nodata


liquidpele:
This was recently fixed in a separate pull request, closing this one.

PR #442, opened 2014-02-10
Layer

DavidSouther:
*This is not ready to merge*

As we've been talking internally at Novus for a couple months, we want to pull out the boilerplate code from many of the charts and impose a sane OO model on top. Some of this work is predicated on the ideas from http://davidsouther.com/2013/02/readable-d3/.

Short-term, I'm pulling out the repeated boilerplate code (margins, size, d3 wrap, axis). Mid-term, this will be a new internal architecture, with the same API. Long term, we should be able to make the entire library much more extensible and readable.

## Testing

From both branches (#development and #canvas), start a webserver in the root directory (python -m SimpleHTTPServer). Open a browser to the `test/` directories. Compare visually (simple refresh on save is fine).


liquidpele:
This is stale and way too big of a pull request anyway... closing.

PR #444, opened 2014-02-12
[428] Partial fix for 428, Updated line charts guildline mouse over eve...

cstephe:
...nt to use a settable custom formatter for the value field. Otherwise uses the xAxis formatter.  This is only available to line chart and only for value and only is applied to guidelines for now.


liquidpele:
should be fixed in latest code, cheers.

PR #446, opened 2014-02-16
Stroke-width on line charts

lindep:
Hi,
This is the first time contributing and mainly want to ask if this is a good way to accomplish the below and if it is an optimal change.    
Add functionality to the line chart to allow the path stroke-width to be changed.

Changed src/models/line.js, also changed the example line.html. 
Added a config setting for the stroke-width in the sin series in examples/line.html

Thanks.


liquidpele:
Each line has a unique class name, just apply the CSS like normal.

PR #447, opened 2014-02-16
Fixed placement of x axis label

luke3141:
The placement of the x axis label was off when a bar chart with 1 or 2 bars was used. This change adds in a  function to calculate it.

Ideally, `rangeExtent` should be used if present as it gives the most direct values to be used. 
Failing that, the calculation depends on the number of values in `scale.range()` as follows:
* `scale.range().length > 2`: use old method
* `scale.range().length == 2 AND scale.range[0] == 0`: use old method (presumably `scale.range() == [0, full_length]`)
* `scale.range().length == 2 AND scale.range[0] != 0`: use same method as for `scale.range().length > 2`
* `scale.range().length == 1`: use scale.range[0].

I have not found a case where the last two methods would be used (presumably it requires an ordinal data set in which case `scale.rangeExtent` is present). However, I've added it in for security - the old method didn't take into account every possible use scenario and I wanted to avoid the same mistake.

### DIFF
--- a/src/models/axis.js
+++ b/src/models/axis.js
@@ -68,6 +68,7 @@ nv.models.axis = function() {
       // scale.range is 1. Difficult to tell.
       var xPosition = scale.range()[0];
     }
+    return xPosition;
   }
 
   function chart(selection) {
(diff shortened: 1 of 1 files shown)


liquidpele:
Closing as duplicate of https://github.com/novus/nvd3/pull/585

Issue #368, opened 2013-11-27
2 line/scatter series with same x and y values gives error when using voronoi


sizeight:
Rendering a chart with the 2 series below causes a JS error with D3V3 but works fine with D2V2. The 2 series have the exact same 5th value of 1.07, and this causes the error.
Error message:
Uncaught TypeError: Cannot read property 'data' of undefined 
```
[
    {
        "key": "Series 1" ,
        "values": [[1030752000000,11.42],[1033344000000,12.08],[1036022400000,12.32],[1038614400000,11.37],[1041292800000,11.07]]
    },
    {
        "key": "Series 2" ,
        "values": [[1030752000000,11.5],[1033344000000,12.24],[1036022400000,12.6],[1038614400000,11.7],[1041292800000,11.07]]
    }
]
```

liquidpele:
This looks to still be an issue in the stackedAreaChart example if you use regular tooltips and not the guideline.  Interestingly I couldn't reproduce errors with the lineChart though...

liquidpele:
Okay, I think https://github.com/novus/nvd3/pull/880 will fix this, although I'd like a better solution eventually, please let me know if anyone still has issues with it!

Issue #330, opened 2013-10-29
Scatter chart error when multiple points have same value


mgcdanny:
Thanks for the awesome software.  

A simple example posted in the jsfiddle link below.  Basically if you create a scatter plot with two points that have the exact same x,y coordinates, then the hover over does not work, nor tooltip, and there is an error in the console:

"Uncaught TypeError: Cannot read property 'data' of undefined"

The issue seems to go away when using d3.v2.js instead of d3.v3.js

http://jsfiddle.net/mgcdanny/FzZTa/

Apologies in advance if I am over looking anything obvious. 

Thanks

liquidpele:
Not a pull request, it's a whole fork.  If you have issues with it, make a pull request or defect with example jsfiddle or whatever and I'll get it fixed up.  

https://github.com/nvd3-community/nvd3

seiyria:
Thanks for sharing @liquidpele - I'll have to look into this more later.

portante:
@liquidpele, I have not been able to solve this problem using the `development` branch from nvd3-community (`341da00`) or novus (`ec20666`).  I am going to submit a fix for this where overlapping points that are considered duplicates are merged to show a tooltip with all the data.

liquidpele:
Really?  The current code does apply a tiny random jitter (Math.random() * 1e-7) to the vertices, so I would have thought this would be fixed, or is the issue something else besides what one of the comments above describes?

portante:
@liquidpele, jitter like that is not guaranteed to solve the problem and because it is random, it might one time and then not for another, do you agree?  Also, if the points are within 1e-6 of each other, even in the same series, 1e-7 jitter won't distinguish from 1e-6 difference, unless I have my math wrong, right?

liquidpele:
Yea, that's true, we could handle this a lot better.

KidA78:
@liquidpele @portante 
I'm also running into this same issue, using the linechart, sometimes I get the above error.  
Using latest nvd3 (1.8.5-dev), and d3 3.5.x

Downgrading D3 to 3.3 definitely fixes it.  But is this the right solution?  I'm unclear whether this is a D3 or NVD3 issue.   @portante I came across a commit where you say this fixes the issue on your own fork of the project: https://github.com/portante/nvd3/commit/f70bc2a3b193322bad1fa75ceb926dcefc17576f

Is there any reason this hasn't been rolled up to fix this for others?
My other question:  is this something that should be fixed in D3 or NVD3?

[...]

PR #448, opened 2014-02-16
delete duplicates from vertices before passing to voronoi

luke3141:
This fixes the problem where a scatter plot (as well as other types of plots that use scatter, including lines that use scatter to determine the mouse-over areas) generates an error for recent versions of d3 (after about 3.3.7) and where there are overlapping points.

The problem is that `voronoi()` requires the vertices that it's passed to  not contain duplicates. (see: https://github.com/mbostock/d3/wiki/Voronoi-Geom#wiki-_voronoi).

This solves issues #330 and #368/#427. It may be a better fix than the latter as it fixes the underlying cause. I also found cases where 427 did not fix the problem.

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -221,6 +221,19 @@ nv.models.scatter = function() {
               [width + 10,-10]
           ]);
 
+	  // delete duplicates from vertices - essential assumption for d3.geom.voronoi
+
+	  var epsilon = 1e-6; // d3 uses 1e-6 to determine equivalence.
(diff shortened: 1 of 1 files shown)


LROCore:
Thank you.  This was killing me and this patch fixed my issue.  Hopefully it gets pulled into the main trunk.

liquidpele:
I'm re-opening this just to track this concept...  a better solution than https://github.com/novus/nvd3/pull/880 should be done, and this is probably part of that.. but we also need a way to get at the multiple point data on hover for the tooltips.

liquidpele:
This was actually done at some point by someone I think.

AgDude:
This was implemented in #1934

PR #452, opened 2014-02-19
Legend - Variable label padding (and symbol/circle size)

Robinson7D:
The circle was hardcoded to 5px radius =  10px in width. 
Previously 28 was added to each label to account for this 10 "plus some padding".
This allows the padding used to be less (or more) than the default 18.

I left the default values as they were so as not to disrupt the styling of old charts.

### DIFF
--- a/src/models/legend.js
+++ b/src/models/legend.js
@@ -11,6 +11,8 @@ nv.models.legend = function() {
     , color = nv.utils.defaultColor()
     , align = true
     , rightAlign = true
+    , labelPadding = 18
+    , symbolRadius = 5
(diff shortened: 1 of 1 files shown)


robinfhu:
Might be useful to add a suite of legend.js formatting options. This is related to https://github.com/novus/nvd3/pull/657

liquidpele:
Yep, agreed, the legend object could use some more options there.

robinfhu:
PR #827 created a legend.padding option. No option for circle size yet.

liquidpele:
@Robinson7D please re-submit a PR based on the latest development branch if you want, this one is too out of date to merge... also instead of changing the radius, it would probably be a better idea to allow custom symbols like what the scatter.js allows now.   Cheers.

Issue #48, opened 2013-01-29
Tooltips for pie chart are broken in IE 10


piterimov:
Tooltips for pie charts in Opera behave strangely (I used Opera 12). They appear and disappear when moving the mouse. Or appear when you click on a pie slice.
I do not know the reason for that, but it would be nice to fix it.
Thanks.

dsm0880:
I found a very easy fix for those who don't want to introduce new JavaScript. Basically I over road the css class .nvtooltip, since the position is set to absolute I simply added a top attribute of ~50px (top:50px) and that seemed to get the tooltip off of the graph so the cursor no longer moves over it. I hope this helps someone.

PR #453, opened 2014-02-20
Issue 48 - hacking around the missing pointer-events:none in internet ex...

liptga:
This is a fix to avoid the flickering of tooltips in pie charts in internet explorer 10 (https://github.com/novus/nvd3/issues/48). The base idea came from the https://github.com/kmewhort/pointer_events_polyfill repository. On leaving the pie I proof if the mouse is on the tooltip or not. If it is on the tooltip, then I prevents its hiding. See the flowcharts in the bug description. I tried to document my changes in the code.

### DIFF
--- a/src/models/pieChart.js
+++ b/src/models/pieChart.js
@@ -3,7 +3,6 @@ nv.models.pieChart = function() {
   //============================================================
   // Public Variables with Default Settings
   //------------------------------------------------------------
-
   var pie = nv.models.pie()
     , legend = nv.models.legend()
     ;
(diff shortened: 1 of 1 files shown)


liquidpele:
Possible fix for the IE10 pie chart tooltip issue?

liptga:
I would call it a workaround :) But it works fine.

liquidpele:
@robinfhu @portante what do you guys think?  I can manually implement this (since we can't auto-merge anymore) if we agree it looks good...

portante:
@liquidpele, can you post a PR so I can take a look?

liquidpele:
this is already a pull request though, or do you want to see a new one based on the current code?

src/models/pieChart.js
```diff
@@ -30,8 +29,29 @@ nv.models.pieChart = function() {
   //============================================================
   // Private Variables
   //------------------------------------------------------------
+  var internetExplorerBefore11 = function() {
```
robinfhu: Let's move this function to nv.utils

src/models/pieChart.js
```diff
@@ -30,8 +29,29 @@ nv.models.pieChart = function() {
...
+    }
+    return false;
+  }
+
+  var pre11InternetExplorer = internetExplorerBefore11();
 
   var showTooltip = function(e, offsetElement) {
+    var pieElementUnderCursor;
+	if ( pre11InternetExplorer ) {
+      //saving the reference to the pie, since we will need it later.
```
robinfhu: there's some bad tab spacing here

src/models/pieChart.js
```diff
@@ -39,8 +59,41 @@ nv.models.pieChart = function() {
...
+    //getting the element underneath
+    var elementUnderneath = document.elementFromPoint(d3.event.pageX, d3.event.pageY);
+    if ( elementUnderneath !== pieElement ) {
+      //we are already out of the pie, removing tooltips
+      dispatch.tooltipHide();
+      //remove pie hover
+      d3.select(findParentWithCSSClass( pieElement, "nv-slice" )).classed('hover', false);
+    }
+    else {
+	  //still in the tooltip and inside the pie. Restoring display value
```
robinfhu: bad tab spacing

src/models/pieChart.js
```diff
@@ -39,8 +59,41 @@ nv.models.pieChart = function() {
...
+      tooltip.style('display',originalStyle);
+    }
   };
 
+  var findParentWithCSSClass = function(node, searchedClassName) {
```
robinfhu: should be a utility function in nv.utils

robinfhu:
It's a little messy solution, but we do need it to work in IE10.  How come other tooltips don't need this type of hack on IE10?

liquidpele:
I think the main issue with IE10 with pie charts is that the tooltip shows up right under the mouse, which makes IE10 freak out... perhaps a better solution is to make sure the tooltip displays a bit away from the mouse?

robinfhu:
@liquidpele  +1.  I like that idea, it may be a more pragmatic solution.

liquidpele:
Okay, closing this and opening an issue for it to track for now.

PR #458, opened 2014-02-25
Fix: Legend. Make sure legend references most recent data.

jgrund:
If nvd3 is being used in a way where the same reference to data is not
kept, the legend can break.

The reason for this is event listeners on legend are only added on
enter; since click and dblclick reference the data from a closure, they
always reference the data value captured when the listeners were setup.

This patch does a lookup of data from series instead of referencing the
captured variable. By doing this, we always have the right data
reference.


liquidpele:
Manually applied this, thanks!

PR #459, opened 2014-02-26
Allow control of yAxis ticks in multiBarChart

smashtie:
Use yAxis ticks if already set rather than computing from available
space, in a similar manner to other models.


liquidpele:
This diff isn't readable, closing in favor of https://github.com/novus/nvd3/pull/601 which is a duplicate.

PR #460, opened 2014-02-26
Tooltips now positioned correctly when nested inside multiple scrollable divs

rhodri:
I had my NVD3 `<svg>` element nested inside a few `<div>` elements, and found that tooltips were displaying off the screen. This fix takes into account the `scrollLeft` and `scrollTop` properties of all ancestor elements of the chart, not just the immediate parent.

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -340,11 +340,23 @@ window.nv.tooltip.* also has various helper methods.
         container.innerHTML = content;
         body.appendChild(container);
         
-        //If the parent container is an overflow <div> with scrollbars, subtract the scroll offsets.
+        //Subtract the scroll offsets of all parent elements.
         if (parentContainer) {
-           pos[0] = pos[0] - parentContainer.scrollLeft;
-           pos[1] = pos[1] - parentContainer.scrollTop;
+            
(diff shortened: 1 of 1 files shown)


liquidpele:
Looks like something to still include, probably needs some unit tests around tooltip positioning too.

anthonyettinger:
we need this, we are having same issue with charts inside a scrollable div. The tooltip position only takes into account window scroll.

liquidpele:
Yup, I'll be digging into the tooltip code pretty soon, but this particular fix isn't really good... recursing up the dom for every tooltip adjustment is way less than ideal.  I have some ideas, most of the time will be just validating all the different scenarios.

anthonyettinger:
i would have expected the tooltip position to be positioned relative to the chart container not the window. Maybe provide an option to enable that?

liquidpele:
This should be fixed inside the latest development code now, closing... but please check it out and open any issues you might find.  Thanks!

anthonyettinger:
@liquidpele  I upgraded but I'm still experiencing the problem. is there something I have to do to enable it?

RenatoUtsch:
This is fixed by #1190.

PR #472, opened 2014-03-17
Added getX and getY default functions.

JacobMetaGeek:
The x and y getter/setter were throwing an error because there was no getY or getX defined.


liquidpele:
fixed in latest code already.

PR #473, opened 2014-03-18
Expose yScale in discreteBarChart model

damiencaselli:
Right now only linear log is set through `discreteBar` inheritance.

What this pull request does is simply add `yScale` to the `discretebar` object, so any `d3.scale` can be used.

graph using logarithmic scale


liquidpele:
Already fixed due to the new option inheritance stuff.

PR #497, opened 2014-04-14
Update d3 in bower.json

cebor:


robinfhu:
We've upgrade d3 in the development branch.

portante:
@robinfhu, is there a way for folks to get a sense of when a development branch release might be made?  Do we have any schedule posted?

robinfhu:
@portante  I haven't given it much thought.  Maybe you and @liquidpele  can help come up with a schedule?
Off the top of my head, I think we can target end of February. We should spend time making sure things are still working and writing up release notes. 

And in the future, you can do a release every month or so.

portante:
@robinfhu, seems reasonable to me.  Particular problems might warrant an early release, like the NPE in  the interactive guide stuff (duplicate points given to d3).

Can you enable milestones for the nvd3?

Or perhaps issue a v1.8.0-alpha release on the development branch so that we can put a stake in the ground?

Thanks!

robinfhu:
Milestones might be good. Never used them in github, let me take a look.

robinfhu:
@liquidpele, @portante , created a milestone. https://github.com/novus/nvd3/milestones

Feel free to add whatever bugs you'd like to get fixed to it. Once all those issues are fixed we'll release.

liquidpele:
Milestones are more for developing things on a roadmap, not just calling the next version number a milestone.  If we have something specific we want to change about nvd3 then it's worth looking at making a roadmap but we haven't discussed much around that yet.  

I think a basic release schedule is a good idea, say a release every 3 or 6 months (unless a serious issue warrants more of course).  Then depending on the kinds of changes we can version it, i.e. something that breaks the api bumps the second number (which is why I went from 1.6.0 to 1.7.0) and otherwise bump the third number.  Bumping up the first number would basically be a whole refactor.

PR #555, opened 2014-05-17
Pie chart: formatting value

alexanderfloh:
`value` should be formatted when used as label for segments.


robinfhu:
PR opened for this change. Closing.

PR #581, opened 2014-06-05
fix issue with pointPaths choking on undefined points after voronoi clip...

rDr4g0n:
Voronoi can clip points out, causing it to return undefined. When the output from voronoi is passed to d3 pointPaths, an undefined value causes an uncaught error to be thrown. The fix just returns M 0 0 for the undefined point.


robinfhu:
We have this change in development now.

PR #582, opened 2014-06-11
Tooltip positioning fixed. tooltip renders inside svg in a <g> container now

arunkjn:


liquidpele:
I looked over this when I was doing a lot of the refactoring in nvd3-community and decided that I liked the tooltips being html based, they're far easier to customize that way...

robinfhu:
Yeah, I add a lot of HTML and CSS to my tooltips, making it an SVG elem would make that harder.

arunkjn:
Well you can still have html based tooltips. That is irrelevant. What's important is that they should render inside the chart container itself, rather than body which they do currently. That way all quirks of positioning them are solved. Maybe you can use foreignobject tag to render them

robinfhu:
@arunkjn I actually think that we could simplify a lot of our tooltip problems if all we did was put a absolutely positioned DIV on the <body> and positioned it by finding:
> hoverElement.getBoundingClientRect()

and then adjusting that using
> window.pageYOffset

eddr:
Hi

I'm using multiBarChart and the tooltip is still in HTML with positioning problems

PR #585, opened 2014-06-12
Fixes x axis label position (v2.0.0-dev)

luke3141:
This is a rebased version of #447.

Formerly, when there was (eg) a bar chart with just 1 or 2 bars then the position was wrong. The reason for this is that scale.range[0] was the position of the first bar, and scale.range[1] the position of the 2nd. The label was being placed at scale.range[1]/2 which was about 1/4 to 1/3 through the graph.

The new version works as follows:
* If there is `scale.rangeExtent` (ie: the scale is ordinal [not continuous], eg: bar chart) then rangeExtent contains the upper and lower values of the range. we use this to calculate the mid point.
* If not, then if there are more than 2 points, we use the same method as previous - take the highest point, add one more interval (distance between consecutive scale points), and find the half way point between here and the start
* if there are only two points and the first one is zero, go half way between them.
* if there are only two points and the first one is not zero, then use the same method as if there are multiple points.
* if there is only one point, then use that.

### DIFF
--- a/src/models/axis.js
+++ b/src/models/axis.js
@@ -66,6 +66,38 @@ Axis.prototype.draw = function(data){
         scale = this.scale(),
         w = null;
 
+    function getXLabelPosition(scale) {
+	// This function calculates the midway position for the X
+	// label. It tries to use rangeExtent() if present. If not, it
+	// defers to range() values. Depending on the chart type, these
+	// can have different meanings and it can be difficult to
+	// calculate the full range of the chart. This should work in all
+	// cases.
+	var xPosition;
+	if ('rangeExtent' in scale) {
+	    xPosition = (scale.rangeExtent()[1] + scale.rangeExtent()[0]) / 2;
+	    return xPosition;
+	}
+
+	if (scale.range().length > 2) {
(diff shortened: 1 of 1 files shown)


robinfhu:
We're no longer working on refactor branch, this will need to be PR'd against development.

luke3141:
Sure, but you were when this PR was submitted 8 months ago. What would need to change for it be accepted next time?

robinfhu:
Create a PR against development branch.  Also, can you post some screenshots explaining what issue this would solve?

luke3141:
Can do but not immediately - I'll need to re-create the problem (presumably with dev branch). I will endeavour to do this in the next couple of weeks.

liquidpele:
Thanks luke, and sorry for the huge lag there.  Just so you know, novus added myself and another community member to their github org so even if robin has to drop off again due to re-prioritization we should still be able to handle things.

luke3141:
Much appreciated. This sounds like an excellent development to what has otherwise been a strong project.

liquidpele:
Hey luke, just checking in :)

liquidpele:
I think luke got hit by a bus.   I'll manually apply his fix in my own PR at some point I guess...

PR #591, opened 2014-06-19
Fixes custom x/y value getters for multiChart

fabrik42:
The multiChart model provides `x()` and `y()` methods like many other nv models, even with the same implementation. But there are some default settings missing, which causes the methods to throw an error: `Uncaught ReferenceError: getX is not defined`. I added the two missing variable declarations with a default method based on the conventions in other models.


liquidpele:
Fix was merged in from my fork, cheers.

PR #596, opened 2014-06-24
New chart type «Exploded Pie» declaration

Koochy:
This patch extends "Pie Chart" in order to get new chart type "Exploded pie" abillity.
Exploded Pie example
exploded pie


robinfhu:
Okay...that's kind of interesting.  Very 'explody' looking.  @liquidpele, what do you think? Worthy addition to NVD3?

liquidpele:
Neat but needs to be cleaned up, not using the new option function format.  I assume there changes are based off the older code.  Also I wonder what this would do donut...I did similar with donut using a single d3 call that might just work with pie too.... AnglePadding I think it was.

robinfhu:
Okay.  I'm leaning towards leaving this out of nvd3 pie, because it will just become another feature we will need to support in the future.

Issue #390, opened 2013-12-18
Reloading data breaks cumulative line chart


bittuamar:
The issue starts when data is reloaded or changed. It seems to break and throws the error saying "TypeError: Cannot read property 'y' of undefined".

I tried to go through the nvd3 code where the error occured and thanks to @cmaurer who helped me spot where the error was occuring. The method 'indexify' is causing the trouble. Can this be fixed ?

You can see the  error with a simple example in the plknr (by @cmaurer) (http://plnkr.co/edit/5QK8Lo?p=preview) that adds a datapoint every time the psdGraph function is called.  

Here are the screen shots

ss1 - first time clicked
![ss1](https://f.cloud.github.com/assets/3953467/1777129/864fc308-6820-11e3-821c-149e74e5d088.png)

- clicked again to load another set of data
![ss2](https://f.cloud.github.com/assets/3953467/1777130/8650624a-6820-11e3-8c8b-a4b232ec16b5.png)

Here is the error
![ss3](https://f.cloud.github.com/assets/3953467/1777131/86529538-6820-11e3-8992-6317fbc40174.png)

robinfhu:
Issue fixed via: https://github.com/novus/nvd3/commit/bea10a4b

PR #606, opened 2014-06-27
Fixes issue #390   by reading raw y instead of scaled y if d.display is null

johnraber:
Issue #390


robinfhu:
I fixed this issue in this commit: https://github.com/novus/nvd3/commit/bea10a4b

PR #609, opened 2014-06-30
Remove NPE in scatter.js updateInteractiveLayer()

nbehnam:
It appears that there may not be data associated with certain pointpaths which cause an NPE when attempting to get the length of the data.


robinfhu:
scatter.js now has this code:
 >if (!d || !d.data || d.data.length === 0)
 >                              return 'M 0 0';

PR #616, opened 2014-07-07
add support for component.io

antoinerousseau:
https://github.com/component/component


robinfhu:
What is component.io?  Another package manager?  Why do we need this?

antoinerousseau:
You don't *need* this. Besides, this PR is 7 months old, and component.io did not really get the expected success in between.

robinfhu:
@antoinerousseau thanks.

PR #627, opened 2014-07-18
Add ()'s so that instanceof check works as intended

MatrixFrog:
instanceof has a higher precedence than the ! operator, so this is needed, otherwise you're just calling 'true instanceof Array' or 'false instanceof Array' which are both always false.


robinfhu:
Thanks. This makes sense.

PR #629, opened 2014-07-24
Missed getX and getY declarations in multiChart

ssergio83:
Added the missed getX and getY declarations to avoid errors during the minification process in production environment


liquidpele:
Merged into latest from my fork, closing, thanks!

PR #632, opened 2014-07-28
Build nv.d3.js and nv.d3.min.js (fixed legend placement version)

duraraxbaccano:
Hi, I've rebuilt these destination files.


robinfhu:
Pull request has too many out of scope files.

PR #635, opened 2014-08-01
Bugfix/edgecase one item to chart

jjviscomi:


robinfhu:
Thanks, this looks like a good fix. Will just need to test it.

PR #657, opened 2014-09-04
Control legend items spacing a bit better

AnSavvides:
The way this currently works is that legend items have a fixed space between one another, however in some cases it might be necessary to customise that space. `legendExtraDistance` will do just that; it's `0` by default and can be set to any value that you want to extend the distance between legend items by.

You can see it in action in the `legend.html` example which I've also updated.

### DIFF
--- a/examples/legend.html
+++ b/examples/legend.html
@@ -23,7 +23,8 @@ nv.addGraph({
 
     var chart = nv.models.legend()
                 .width(width)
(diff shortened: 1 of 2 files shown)


robinfhu:
No a bad idea. I'd like to rename this 'legendPadding'.

robinfhu:
Opened new PR for this feature.

PR #661, opened 2014-09-10
Fix broken legend when data is changed

sylvinus:
Not a d3/nvd3 expert, but this is needed to make my custom legend behave (it uses dynamically-generated legend data, not directly the main graph's data).

This should (admittedly not the ideal way) fix errors some other users have also been getting:
http://stackoverflow.com/questions/23015095/nvd3-passing-wrong-event-on-statechange-when-clicking-legend-after-chart-is-re
("data always points to the initial data array").

### DIFF
--- a/src/models/legend.js
+++ b/src/models/legend.js
@@ -47,6 +47,10 @@ nv.models.legend = function() {
             dispatch.legendMouseout(d,i);
           })
           .on('click', function(d,i) {
+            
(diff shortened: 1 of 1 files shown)


robinfhu:
I'd like to test this first. Just need to understand what this solves.

robinfhu:
PR #802 should solve this issue.

PR #662, opened 2014-09-10
range of one element fix

DarthVictor:


robinfhu:
Cannot just update nv.d3.js without updating the original source file in src/models.

PR #687, opened 2014-10-09
Fix broken links in indentedtree.html

gb96:
Great work on nvd3. I'm using it at a green-tech start-up to chart electricity consumption...

1. Update link to nvd3,org home page.
2. Make all links to other examples relative, so they won't break so easily.


robinfhu:
I think this is fine. Thanks.

PR #696, opened 2014-10-14
Prefixed svg element in css to avoid affecting other svg elements from other libraries.

cirilomegg:
The svg classes on the css are not prefixed by nvd3 what causes other svgs not generated with nvd3 to be affected.

### DIFF
--- a/src/nv.d3.css
+++ b/src/nv.d3.css
@@ -143,7 +143,7 @@
  */
 
 
-svg {
+.nvd3 svg {
   -webkit-touch-callout: none;
   -webkit-user-select: none;
   -khtml-user-select: none;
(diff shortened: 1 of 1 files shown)


kdekooter:
This is a nasty issue. When will it be merged?

liquidpele:
This project is abandoned, see here instead:   https://github.com/nvd3-community/nvd3

robinfhu:
This is a good idea.  Just need to consider how this will affect our projects.

robinfhu:
I committed my own changeset, which leaves one of the svg styles in.  It seems we need the 'width:100%' and 'height:100%' styles to make it work with the examples.

Issue #703, opened 2014-10-20
Shadow DOM issues


naunga:
The tooltips do not render correctly due to styles being encapsulated and because the logic placing tooltip containers within the DOM does not account for the fact that chart's parent element may be within a shadow DOM. 

Encountered these issues while working on a Polymer element to encapsulate nvd3 charts. You can take a look at that project here: https://github.com/naunga/polychart-element

liquidpele:
You'll need to send us a pull request  that handles only the issue itself

PR #704, opened 2014-10-20
Shadow dom tooltips

naunga:
Proposed fix for issue #703.

Provides better compatibility with the shadow DOM. Reworked logic for tooltip container placement which accounts for charts that exist within a shadow DOM. The pie chart objects have been reworked to use the tooltip model vs. the original nv.tooltip.show function.

This fork will be used in a production-project shortly to accommodate a Polymer element that wraps nvd3.


robinfhu:
Way too big of a change.  Closing for now, feel free to use this in your fork.

PR #744, opened 2015-01-01
Added Rounded Top Bars 

moneytree-doug:
:3


moneytree-doug:
And removed deprecated files

PR #757, opened 2015-02-08
Added ability to specify line width

jstrength:
Sometimes it is nice to be able to have lines of different width. To use this feature you just need to add a width property, as I've done in the example files. I had to remove the stroke-width style from the nv-line rule and added some logic to line.js to choose 1.5 as default unless the width property exists. I also removed the nv-line.hover rule because it wasn't being used anywhere and was modifying stroke-width.

### DIFF
--- a/examples/line.html
+++ b/examples/line.html
@@ -89,7 +89,8 @@
             {
                 values: cos,
                 key: "Cosine Wave",
-                color: "#2ca02c"
+                color: "#2ca02c",
+                width: 3
             }
         ];
     }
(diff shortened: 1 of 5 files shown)


robinfhu:
Hmmm, not sure I support this idea fully yet.  Line width is really a styling issue, which should be handled in CSS.  You can specify a specific line by looking for the 'series-0' class being applied to each line.

robinfhu:
Oh looks like it was merged.  Okay, we'll take it.

PR #759, opened 2015-02-09
Updating package.json so that on postinstall it will do bower install.

robinfhu:
Simple configuration change.


joshjordan:
Should we be doing `bower cache clean` here as well? https://github.com/auth0/auth0.js/issues/12

robinfhu:
@joshjordan  I don't know.  Because travis-ci seems to start on a clean environment, it doesn't seem necessary

Issue #754, opened 2015-02-06
Interactive Bisect code broken for d3 3.5


robinfhu:
These tests currently break in the development branch:

     ✗ accessor in index mode - existing item
	expected 1 to equal 7

      ✗ accessor in index mode - inbetween item 1
	expected 1 to equal 7

      ✗ accessor in index mode - inbetween item 2
	expected 1 to equal 8

Question: why doesn't this code work with d3 v3.5?

portante:
I think because of https://github.com/mbostock/d3/wiki/Arrays#d3_bisector.

Bisector now considers the given function as an accessor if it has one parameter and as a comparator if it has two, and we always use two.

Will post a proposed fix.

robinfhu:
Yes I just saw: https://github.com/mbostock/d3/commit/3c7cc81b6233289263525b581d635ad9f8d9dd37

This screws up my logic.  Yeah would love to see a fix.

portante:
@robinfhu, if you have a better solution, I'm all ears, but it appears that D3 broke the API contract as of v3.4.4.

portante:
@robinfhu, looks like we can close this one now.

PR #762, opened 2015-02-09
Move minimum D3 version to 3.4.4

portante:
As of D3 commit:

https://github.com/mbostock/d3/commit/3c7cc81b6233289263525b581d635ad9f8d9dd37

We are no longer able to use an array accessor that accepts two
parameters (the second being the index of the first parameter in the
given array), since d3.bisector() will consider it a comparison
function instead of an accessor.

There does not appear to be a way to get the index value of the element
being considered in the context of the accessor.

As a result, this commit proposes we drop the interactiveBisect()
support for accessors accepting a second parameter being the index of
the first in a given array.

We now wrap all x-accessors passed to interactiveBisect() hiding the
second parameter, using an internal comparator instead.

This commit also attempts to get the coverage to 100% for
interactiveBisect().

See issue #754 as well.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "nvd3",
-  "version": "1.7.1",
   "homepage": "http://www.nvd3.org",
   "authors": [
     "Bob Monteverde",
@@ -22,7 +21,7 @@
   ],
   "license": "Apache License, v2.0",
   "dependencies": {
-    "d3": "^3.3.13"
+    "d3": "^3.4.4"
   },
   "ignore": [
     "**/.*",
@@ -37,11 +36,5 @@
     "*.xml",
     "*.json",
     "*.md"
-  ],
-  "resolutions": {
-    "d3": "^3.3.13"
-  },
-  "devDependencies": {
-    "d3": "^3.3.13"
-  }
+  ]
 }
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Novus has decided to keep the library in sync with what is in nvd3-community mas
 
 **D3 Versioning Note**
 
-NVD3 should work with the latest d3.js (version 3.5), but I did notice that the interactive guideline tooltip was broken for d3.js 3.5.  Changing it to d3.js 3.3.13 fixes it. This will need to be investigated.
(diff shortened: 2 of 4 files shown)


bower.json
```diff
@@ -1,6 +1,5 @@
 {
   "name": "nvd3",
-  "version": "1.7.1",
```
liquidpele: version should stay there, the gruntfile reads it from here!

portante: But I thought bower version numbers don't mean anything.  See https://github.com/bower/bower.json-spec

robinfhu:
I'm okay with this actually.

robinfhu:
Originally I wanted to find a way to support d3 3.3 and below, but now I think it's not worth it.

bower.json
liquidpele: They don't, but the gruntfile reads in the bower json to get the version, so the version is stored only in a single place :)

liquidpele:
I'm fine with the D3 version bump, people should use close to latest anyway.  However, you should probably add a comment explanation in there to help when we re-read that code a year from now.

bower.json
portante: It looks to me like it reads package.json, no?

liquidpele: Agh... In my defense I have not had my coffee yet!

robinfhu:
Okay, so there's one small hiccup with this change.  In our code, we actually use function(d,i){return i;} for our accessor a lot.  The original d3 bisect code passed in the 'index' to the accessor function.  I need a way to get that index value.

joshjordan:
Please also update the `README` with this change.

bower.json
liquidpele: :coffee: :coffee: :coffee:

robinfhu:
I am going to try something slightly controversial.  I'm going to just write my own bisect left function.

liquidpele:
So actually, I'm a little confused about this one anyway.  What exactly wasn't working?  I was doing my testing with d3 3.5 and never saw any issues.

portante:
@robinfhu, is that really worth it?  What do we gain?

@joshjordan, okay, will do.

bower.json
portante: No defense necessary, all this stuff is quite confusing.

robinfhu:
@portante @liquidpele , thought about it some more, I think you are right.  In some of our code, we've been relying on using the data index for the x-accessor.  We do function(d,i){ return i;} a lot.

But I realize now we can modify our code to manually add an index to each piece of data, so I can do something like: function(d,i){ return d._myIndex;}

So basically, I think everything is fine :-)

robinfhu:
Thanks @portante

Resseguie:
There were still a few version references out of sync after this merge. See #789

liquidpele:
I merged the fix.

PR #772, opened 2015-02-10
Initial Node module support.

jjzazuet:
Hi. Like the title says.

I see that the NPM registry only has version 1.1.15 beta.

So I think these changes should allow others to `require()` version 1.7+ of the library.

Please let me know if there's anything else I should specify. Thanks!

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1,7 +1,6 @@
 
-// set up main nv object on window
-var nv = window.nv || {};
-window.nv = nv;
+// set up main nv object
+var nv = {};
 
 // the major global objects under the nv namespace
 nv.dev = false; //set false when in production
@@ -120,6 +119,10 @@ nv.addGraph = function(obj) {
     }
 };
 
-if ('undefined' !== typeof exports && 'undefined' !== typeof module) {
-    module.exports = nv;
+if (typeof(module) !== 'undefined' && typeof(exports) !== 'undefined') {
+  module.exports = nv;
+}
+
+if (typeof(window) !== 'undefined') {
+  window.nv = nv;
 }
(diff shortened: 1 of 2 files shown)


src/core.js
```diff
@@ -118,4 +118,8 @@ nv.addGraph = function(obj) {
     if (!nv.render.active) {
         nv.render();
     }
-};
\ No newline at end of file
+};
+
+if ('undefined' !== typeof exports && 'undefined' !== typeof module) {
```
robinfhu: Do: typeof(exports) !== 'undefined'

robinfhu:
@jjzazuet , so after you made these changes, were you able to load nvd3 into a node.js program?

jjzazuet:
At least I was in my application. Node users must be aware of the fact that both `nvd3` and `d3` itself rely heavily on the `window` object. Let me find some time to put together a test case tomorrow. In the meantime, please hold on this request. Cheers!

jjzazuet:
Ack. I think it might take more time to try to mock the `window` object inside a pure node environment.

In the meantime, I just changed the object export logic as per your recommendations. However, I'm worried about this statement:

    var nv = window.nv || {};

Just to clarify, during script load time, is it possible we may encounter a preexisting instance of `nv` inside the `window` global?

Thanks for your time!

joshjordan:
Why do we need to mock the window? How is nvd3 useful outside of a browser?

jjzazuet:
@joshjordan Right. My exact same thought. And the answer is, it's really not. So in that case, my current commit works inside my `gulp/browserify` based application with the packaging changes I made. The only thing left to consider is, why `var nv` is assigned to an existing `window` property or a new object. That's what I'd like to understand.

Thanks again for your time and help!

Issue #279, opened 2013-09-20
Property 'arguments.callee' should not be used.


stickwithchrist:
In the render() function:

```javascript
nv.render = function render(step) {
  step = step || 1; // number of graphs to generate in each timeout loop

  nv.render.active = true;
  nv.dispatch.render_start();

  setTimeout(function() {
    var chart, graph;

    for (var i = 0; i < step && (graph = nv.render.queue[i]); i++) {
      chart = graph.generate();
      if (typeof graph.callback == typeof(Function)) graph.callback(chart);
      nv.graphs.push(chart);
    }

    nv.render.queue.splice(0, i);

    if (nv.render.queue.length) setTimeout(arguments.callee, 0);
    else { nv.render.active = false; nv.dispatch.render_end(); }
  }, 0);
};
```

This (arguments.callee) has been deprecated and only renders type errors in strict mode.

liquidpele:
Seems to still be an issue... callee is indeed depreciated:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments/callee

robinfhu:
We can use the solution @stickwithchrist posted.  I am cool with that.  We can even use this opportunity to rewrite this logic.

robinfhu:
I'll take care of this.

PR #775, opened 2015-02-11
Fixing issue #279: removing arguments.callee, which is deprecated.

robinfhu:
Also adding some better comments explaining nv.addGraph

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -99,7 +99,7 @@ nv.render = function render(step) {
         nv.render.queue.splice(0, i);
 
         if (nv.render.queue.length) {
-            setTimeout(renderLoop, 0);
+            setTimeout(renderLoop);
         }
         else {
             nv.dispatch.render_end();
@@ -107,7 +107,7 @@ nv.render = function render(step) {
         }
     };
 
-    setTimeout(renderLoop,0);
+    setTimeout(renderLoop);
 };
 
 nv.render.active = false;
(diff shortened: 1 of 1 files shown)


robinfhu:
@liquidpele please review.

src/core.js
```diff
@@ -97,18 +98,38 @@ nv.render = function render(step) {
 
         nv.render.queue.splice(0, i);
 
-        if (nv.render.queue.length) setTimeout(arguments.callee, 0);
+        if (nv.render.queue.length) {
+            setTimeout(renderLoop, 0);
```
joshjordan: We can omit the second param when you want `0`

src/core.js
```diff
@@ -97,18 +98,38 @@ nv.render = function render(step) {
...
+        }
         else {
             nv.dispatch.render_end();
             nv.render.active = false;
         }
-    }, 0);
+    };
+
+    setTimeout(renderLoop,0);
```
joshjordan: Ditto.

src/core.js
```diff
@@ -97,18 +98,38 @@ nv.render = function render(step) {
...
+or 
+
+nv.addGraph(<generate Function>, <callback Function>)
+
+The generate function should contain code that creates the NVD3 model, sets options
+on it, adds data to an SVG element, and invokes the chart model. The generate function
+should return the chart model.  See examples/lineChart.html for a usage example.
+
+The callback function is optional, and it is called when the generate function completes.
+*/
```
joshjordan: This is great work, but let's try to keep completely separate pieces of work in separate commits.

robinfhu: Yep will keep in mind

Issue #770, opened 2015-02-10
multiBarChart or multiChart with focus


vishal1shukla2:
Is it possible to have focus area like lineWithFocusChart & linePlusBarWithFocusChart. I would like to propose this feature if this is not possible already.

liquidpele:
It is not currently implemented, but feel free to code it up and send us a pull request!   We rely on a lot of community enhancement submissions :)

vishal1shukla2:
Sure. I am not expert JS developer, but I managed to get this working with just couple of small glitches, I am sure these would be pretty small fix for other contributors. Is it possible to send pull request for code in such state?

joshjordan:
Please do send the pull request. We can help you get it into an acceptable
state.

vishal1shukla2:
Pull request #778. Thanks.

PR #778, opened 2015-02-11
Enhancement #770 - MultiBarWithFocusChart feature in multibarChart model...

vishal1shukla2:
Just one small in my local example, need to call chart.update() in callback after chart is rendered to make context area visible. Though this is not reproducible in the examples in the code.

### DIFF
--- a/src/models/multiBarChart.js
+++ b/src/models/multiBarChart.js
@@ -1,4 +1,3 @@
-
 nv.models.multiBarChart = function() {
     "use strict";
 
@@ -7,19 +6,29 @@ nv.models.multiBarChart = function() {
     //------------------------------------------------------------
 
     var multibar = nv.models.multiBar()
+        , multibar2 = nv.models.multiBar()
         , xAxis = nv.models.axis()
         , yAxis = nv.models.axis()
+        , x2Axis = nv.models.axis()
+        , y2Axis = nv.models.axis()
         , legend = nv.models.legend()
         , controls = nv.models.legend()
+        , brush = d3.svg.brush()
         ;
 
-    var margin = {top: 30, right: 20, bottom: 50, left: 60}
+    var margin = {top: 30, right: 20, bottom: 250, left: 60}
+        , margin2 = {top: 0, right: 30, bottom: 20, left: 60}
         , width = null
         , height = null
+        , height2 = 100
         , color = nv.utils.defaultColor()
         , showControls = true
         , controlLabels = {}
         , showLegend = true
+        , focusEnable = false
+        , focusShowAxisY = false
+        , focusShowAxisX = true
+        , focusHeight = 50
         , showXAxis = true
         , showYAxis = true
         , rightAlignYAxis = false
@@ -33,10 +42,13 @@ nv.models.multiBarChart = function() {
         }
         , x //can be accessed via chart.xScale()
         , y //can be accessed via chart.yScale()
+        , x2
+        , y2
         , state = nv.utils.state()
+        , brushExtent = null
         , defaultState = null
         , noData = "No Data Available."
-        , dispatch = d3.dispatch('tooltipShow', 'tooltipHide', 'stateChange', 'changeState', 'renderEnd')
+        , dispatch = d3.dispatch('tooltipShow', 'tooltipHide', 'brush', 'stateChange', 'changeState', 'renderEnd')
         , controlWidth = function() { return showControls ? 180 : 0 }
         , duration = 250
         ;
@@ -46,6 +58,11 @@ nv.models.multiBarChart = function() {
     multibar
         .stacked(false)
     ;
+
+    multibar2
+        .stacked(false)
+    ;
+
     xAxis
         .orient('bottom')
         .tickPadding(7)
@@ -57,7 +74,14 @@ nv.models.multiBarChart = function() {
         .orient((rightAlignYAxis) ? 'right' : 'left')
         .tickFormat(d3.format(',.1f'))
     ;
+    x2Axis
 
+        .orient('bottom')
+        .tickPadding(5)
+    ;
+    y2Axis
+        .orient('left')
+    ;
     controls.updateState(false);
 
     //============================================================
@@ -110,7 +134,8 @@ nv.models.multiBarChart = function() {
             var availableWidth = (width  || parseInt(container.style('width')) || 960)
                     - margin.left - margin.right,
                 availableHeight = (height || parseInt(container.style('height')) || 400)
-                    - margin.top - margin.bottom;
+                    - margin.top - margin.bottom - (focusEnable ? focusHeight : 0) ,
+                availableHeight2 = focusHeight - margin2.top - margin2.bottom;
 
             chart.update = function() {
                 if (duration === 0)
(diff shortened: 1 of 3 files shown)


robinfhu:
Remove everything from the build/* folder from this PR please.

robinfhu:
@vishal1shukla2 , haha, no, I did not mean actually delete those files from the git repo.  We don't want you to checkin any changes to build/nv.d3.js or build/nv.d3.css for now.  The only files that should be changing are the multiBarChart files.

vishal1shukla2:
Sorry for messing it up @robinfhu, had to reset the head and commit just 3 files. Please let me know if this should be done in some other, better way.

liquidpele:
The pull request still has stuff from under build being removed...

vishal1shukla2:
Fixed the commits to have just 3 files.

vishal1shukla2:
Note that focus functionality works only for numeric/date based x-scale. If x-axis domain is NaN, it will simply disable focus.

robinfhu:
So I've looked at this now, and it's a lot of code we will need to support going forward. And it's for a feature I think doesn't make much sense with multiBarChart. The focus feature is useful for charts with lots of data points, like a line chart, and you want to zoom in.  That use case is extremely rare with multiBarCharts.  I have a feeling this was built for a project you specifically needed?

@liquidpele thoughts?

vishal1shukla2:
I agree with your thoughts that focus feature does not make sense when we have limited data-points in X-axis. My feeling is, one of the common use-cases is observing top N values of some dimension over the period of time - in form of stacked bar chart.

robinfhu:
Took a screenshot of this feature for reference.
screen shot 2015-02-16 at 8 07 22 am

So I looked at the example, and it's as I expected.  You did a good job implementing it, and it works.  I just think the feature is overkill right now and adds a lot of code we would need to support later on.  I'd like to keep this out of NVD3 unless others feel there is a strong desire for it.  

@vishal1shukla2 thanks for the PR.  Feel free to keep this code in your fork and use it in whatever projects you have!

robinfhu:
@vishal1shukla2 what you've built here though is a great example of just how flexible NVD3 is.  You've shown that it's really easy to take the existing chart models and extend it to do whatever you want.  This is a great example for the community.

liquidpele:
I dunno, I kind of like having the focus feature on as many things as it can work on, plus it's already on linePlusBar, so there is already precedent for it being with bars.  

@portante what do you think?

robinfhu:
LinePlusBar uses a single bar series, whereas multiBar can have more than one bar on an X-tick.

liquidpele:
I'm not sure why that changes anything... either focus is useful for bars or not.  Personally I think it's useful to focus in, especially if it can then show the values over the bars when the bars get thicker due to the focus.  It's also useful if people want to for instance zoom into different parts and export the svg to a PNG or something similar.

src/models/multiBarChart.js
```diff
@@ -163,22 +188,58 @@ nv.models.multiBarChart = function() {
...
             var wrap = container.selectAll('g.nv-wrap.nv-multiBarWithLegend').data([data]);
             var gEnter = wrap.enter().append('g').attr('class', 'nvd3 nv-wrap nv-multiBarWithLegend').append('g');
             var g = wrap.select('g');
 
-            gEnter.append('g').attr('class', 'nv-x nv-axis');
-            gEnter.append('g').attr('class', 'nv-y nv-axis');
-            gEnter.append('g').attr('class', 'nv-barsWrap');
-            gEnter.append('g').attr('class', 'nv-legendWrap');
-            gEnter.append('g').attr('class', 'nv-controlsWrap');
+            var focusEnter = gEnter.append('g').attr('class', 'nv-focus');
```
robinfhu: If focusEnabled is false, these elements still get generated and appended to the DOM.

src/models/multiBarChart.js
```diff
@@ -429,8 +674,25 @@ nv.models.multiBarChart = function() {
             rightAlignYAxis = _;
             yAxis.orient( rightAlignYAxis ? 'right' : 'left');
         }},
+        xTickFormat: {get: function(){return xAxis.xTickFormat();}, set: function(_){
```
robinfhu: Why were these options created?  Normally we just use xAxis.tickFormat().  Though, I think the goal here is to set both the chart and the focus chart's axis formatting

src/models/multiBarChart.js
```diff
@@ -429,8 +674,25 @@ nv.models.multiBarChart = function() {
...
+            multibar2.x(_);
+        }},
+        y: {get: function(){return multibar.y();}, set: function(_){
+            multibar.y(_);
+            multibar2.y(_);
+        }},
         barColor:  {get: function(){return multibar.barColor;}, set: function(_){
             multibar.barColor(_);
+            multibar2.barColor(_);
             legend.color(function(d,i) {return d3.rgb('#ccc').darker(i * 1.5).toString();})
```
robinfhu: This code was duplicated by you in line 239, just curious as to why

robinfhu:
@liquidpele , I see what you're saying, but let me try to explain my thought process.  I've always thought the focus feature for being useful for 'time series' related data, which line charts are mostly used for.  I tend to see multiBarChart as being used for discrete series'.   This is apparent when you realize that the focus feature only works on a specific start and end range.

The bar chart that you see in the linePlusBarChart is really a historicalBarChart with lines overlaid on it. A historicalBarChart is basically a line chart but with each point replaced with a vertical bar.  That's why it makes sense to have the focus feature for linePlusBarChart, because it's used for time series'.

As a sidenote, if the focus feature is so useful, perhaps we should create it as a separate component which can be plugged into various charts.  That way we aren't duplicating so much code.

robinfhu:
Also, @liquidpele , you bring up good points about how people might find it useful to select certain bars so that they can export it.  Users of nvd3 should be aware that it's relatively easy to hook in external widgets to the chart.  So you can create an HTML button that updates the data and redraws the chart.

portante:
@robinfhu, like the idea of having a focus option as a separate plug-in, seems quite useful.

liquidpele:
Me too... easier said than done though...  I think it would take some major refactoring.

liquidpele:
Me too... easier said than done though...  I think it would take some major refactoring.

src/models/multiBarChart.js
vishal1shukla2: Had used couple of charts(lineWithFocus & linePlusBarChart) as reference. So just tried to keep is consistent to the way it is implemented in linePlusBarChart. If it must be changed, let me know.

src/models/multiBarChart.js
vishal1shukla2: As mentioned in last comment, found similar implementation in lineWithFocusChart. This can be safely removed as well, if it is not required to have these functions externalised.

src/models/multiBarChart.js
vishal1shukla2: This was missed after certain trial/errors. Thanks for pointing it out. Please confirm earlier to changes as well, so I can add the commit to PR.

src/models/multiBarChart.js
robinfhu: I still have the opinion that this feature does not make much sense with this chart. So don't worry too much about fixing these

mila76:
I try myself to create this but no time to fish...
Now i can try to use this, thanks.
I think is a useful feature, sometime bar is so small and a chart link the screenshoot is so more readable

liquidpele:
I'd like to take another look at this....   @vishal1shukla2 can you rebase this against latest?

amalahussein:
+1 on the need for this feature--I'm working on a very specific feature where I need a focus element for a multibar chart---kudos for adding this feature @vishal1shukla2! Hope to see this merged in soon.

leongaban:
Please implement this in the original repo! We need this feature, nvd3 is the plug and play version of d3.

liquidpele:
ah!  This is the other one I accidentally closed when I deleted the dev branch.  Leaving this closed, and I created a new PR of it pointing to master.

PR #782, opened 2015-02-12
Meteor Integration

bobiblazeski:
Hello guys,

As you might know, the Meteor community is widely using nvd3. We had many different wrapper packages all doing the same thing, so I'm submitting a PR that integrates Meteor packaging directly into your repository: at the end of this merge process, in case you'll find yourself comfortable with it, we'll get automated new Meteor package releases with every new release of nvd3.

I've made this PR as lean as possible so as not to take up too much space within your repository: basically it adds only a package.js file among the ones already present for other package managers.

All you have to do after accepting this PR would be:

Create an account at https://meteor.com/ (click SIGN IN, then Create account). After you've done that, please let me know the name of the account, and we'll add you as a maintainer of the nvd3 organization (which has been already reserved to make things simpler for you).

Activate your repo on autopublish.meteor.com

Head to autopublish.meteor.com, sign in with your GitHub account, locate this repository and enable it for autopublish by toggling the checkbox you'll find on the right side of the repository item. You can get an idea about how Meteor Autopublish works by having a quick look at this page, but basically it creates a web hook to let us know when you push new tags. When you'll release new versions, autopublish.meteor.com will automatically package the the new version of the library also for Meteor!

In case you're not comfortable with autopublish.meteor.com maintaining access rights to your repos, please head to GitHub Application Settings and revoke permissions for Meteor Autopublish.

Please note that you will not be able to toggle your repository until you'll have the package.js file on your master branch.

Alternatively, in case you're not willing to give read/write rights to your public repositories to anyone else but you (which is something appreciable), we will be available to provide a step by step guide about manually creating the required webhook.

Please also note that we've already published the current version of the package on Atmosphere https://atmospherejs.com/nvd3/nvd3 under the name nvd3:nvd3. If you think another name would be a better fit, just let us know!

Thanks in advance for your availability and collaboration and congratulation for the awesome project!

### DIFF
--- a/GruntFile.js
+++ b/GruntFile.js
@@ -1,4 +1,5 @@
 var version = '1.7.1';
+var oldVersion = '1.7.1';
 module.exports = function(grunt) {
     var _pkg = grunt.file.readJSON('package.json');
 
@@ -35,6 +36,18 @@ module.exports = function(grunt) {
                 files: {
                     'build/nv.d3.min.js': ['build/nv.d3.js']
                 }
+            }
+        },
+        replace: {
+            version: {
+                src: [
+                    'package.js'
+                ],
+                overwrite: true,
+                replacements: [{
+                    from: oldVersion,
+                    to: version
+                }]
             }
         },
         jshint: {
@@ -108,9 +121,10 @@ module.exports = function(grunt) {
     grunt.loadNpmTasks('grunt-contrib-copy');
     grunt.loadNpmTasks('grunt-contrib-cssmin');
     grunt.loadNpmTasks('grunt-karma');
+    grunt.loadNpmTasks('grunt-text-replace');
 
     grunt.registerTask('default', ['concat', 'karma:unit']);
-    grunt.registerTask('production', ['concat', 'uglify', 'copy', 'cssmin']);
+    grunt.registerTask('production', ['concat', 'uglify', 'copy', 'cssmin', 'replace']);
     grunt.registerTask('release', ['production']);
     grunt.registerTask('lint', ['jshint']);
 };
--- a/GruntFile.js
+++ b/GruntFile.js
@@ -1,5 +1,4 @@
 var version = '1.7.1';
-var oldVersion = '1.7.1';
 module.exports = function(grunt) {
     var _pkg = grunt.file.readJSON('package.json');
 
@@ -45,8 +44,8 @@ module.exports = function(grunt) {
                 ],
                 overwrite: true,
                 replacements: [{
-                    from: oldVersion,
-                    to: version
+                    from: /(version?\s?=?\:?\s\')([\d\.]*)\'/gi,
+                    to: '$1' + _pkg.version + "'"
                 }]
             }
         },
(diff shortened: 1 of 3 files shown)


robinfhu:
Thanks, didn't realize the Meteor.JS community used nvd3 so much.  
Since this adds a new file to our root folder, many people not familiar with Meteor will be confused as to what it's for.  Can you add some comments to the package.js file to explain what it's for.  

Also, how does the 'oldVersion' variable in the gruntfile impact things?

GruntFile.js
```diff
@@ -37,6 +38,18 @@ module.exports = function(grunt) {
...
         },
+        replace: {
+            version: {
+                src: [
+                    'package.js'
+                ],
+                overwrite: true,
+                replacements: [{
+                    from: oldVersion,
+                    to: version
```
liquidpele: Actually the version up there I think was from a merge of older stuff and shouldn't be there... the version should come from the _pkg.version, i.e. from the package.json.   The package.js can probably do the same thing, no?

bobiblazeski:
@robinfhu Thanks for the quick reply. As you might see there is already 7 versions of nvd3 at atmosphere js https://atmospherejs.com/?q=nvd3 

package.js is the description for the meteor packing system, I'll add some comments to keep things clear.

oldVersion is added only because I didn't know how to replace the version in the grunt build.
I'm just doing string replacement into package.js to update the version.  I don't use grunt so I can't recommend anything better but maybe using something like this https://github.com/Dogfalo/materialize/blob/a6243e866437ef5fc5d0b0d3d452102443fcfd3a/Gruntfile.js#L409-422 
called by grunt release --oldver 1.0.0 --newver 1.0.1 might suit you. 

@liquidpele I'll switch to  _pkg.version, if grunt.option( "oldver" ), is inconvenient to you

liquidpele:
Can package.js not read the version from package.json directly instead of a fragile string replace hack?

bobiblazeski:
@liquidpele  No  meteor uses package.js as its packaging mechanism http://docs.meteor.com/#/full/packagejs ,  it doesn't need npm package.json.
From my understanding of your build process your changing version at package.json and then run grunt production. If you have problem with string replacement I could use template, but that would add another file.

liquidpele:
would you ever be using the package. As file when the package. Json file wasn't there?

liquidpele:
even a clever regex replace would be better than the string her place because then you wouldn't be relying on keeping the old version correct

splendido:
Hey guys, sorry to chime in...
Would be something like [this](https://github.com/MeteorPackaging/autopublish.meteor.com/wiki/Caveat#useful-regexp) acceptable?

liquidpele:
Year I think that would be fine.... it removes the issue of keeping track of the older version and it matching what was in that file and all that

dohomi:
@bobiblazeski @splendido @dandv 

The current project is ```1.8.1``` but in Atmosphere the last version is ```1.7.1```. Shouldn't this update be automatically published to ```Meteor```? I'm not familiar with the script could you have a look if something is missing? Thanks!

bobiblazeski:
@dohomi I'm afraid I don't know how automatic update works. I  wait a little for help from @splendido @dandv    if it takes too long will update version manually

splendido:
I'm not sure we ever registered nvd3 on autopublish.meteor.com
...anyway, at the moment we're having some problem with our bot user, so I'd say manual update is the only option (at least temporarly..)

Issue #785, opened 2015-02-12
ScatterChart is missing options binding


jamesgarfield:
All other charts use `chart.options = nv.utils.optionsFunc.bind(chart);` to bind the options function to the chart object, but scatterChart is missing this.

liquidpele:
Verified, looks like an oversight from when I was redoing stuff, fixed by @jamesgarfield in the referenced pull request above.

PR #786, opened 2015-02-12
Add options function binding to scatterChart

jamesgarfield:
Refs #785

Every other chart has this. Needed to have chart.options() be the standard way to access a chart's options.


liquidpele:
tis beautiful, thanks for the pull!

Issue #787, opened 2015-02-12
Logic for 'noData' on charts is duplicated everywhere


jamesgarfield:
The only difference across all charts is what the test for there being no data looks like.

This can be extracted into a utils method.

liquidpele:
No need to open an issue for non-bugs if you're already planning on doing the pull request... we'll just track the pull request in this case :)

PR #788, opened 2015-02-12
Extract noData functionality

jamesgarfield:
Ref #787 

After fixing the chart clearing bug, I thought I should capitalize on my current familiarity of the 'noData' logic to extract it into it's own utils function.

Each chart still has it's own tests for 'noData', but I've also added one in the utils unit tests for safety.


src/utils.js
```diff
@@ -592,3 +592,39 @@ Calculate the available width for a chart.
...
     return nv.utils.sanitizeWidth(width, container) - margin.left - margin.right;
 }
+
+/*
+Clear any rendered chart components and display a chart's 'noData' message
+*/
+nv.utils.noData = function(chart, container) {
+    var opt = chart.options(),
+        margin = opt.margin(),
+        noData = opt.noData(),
```
liquidpele: Lets do noData = opt.noData() || "No Data Available", and set the default of noData to null in the rest of the models... that way the default string is only defined in one place.

src/utils.js
```diff
@@ -592,3 +592,39 @@ Calculate the available width for a chart.
...
+        .attr('y', y)
+        .text(nv.utils.identity);
+}
+
+/*
+Identity function
+*/
+nv.utils.identity = function(i) { 
+    return i; 
+}
```
liquidpele: This function seems silly to be broken out by itself when it's only called in once place, just make it an anonymous function above like the original code was doing.

jamesgarfield: I disagree about the utility of this, but I'm happy to leave it for another time.

liquidpele:
Looks good!

Issue #760, opened 2015-02-09
Interactive guideline tooltip was broken for d3.js 3.5


joshjordan:
Moving this comment here from the `README` so we can track it as task.

robinfhu:
Yeah this is causing the 'development' branch tests to fail.  Still thinking of a solution...

robinfhu:
Solved by https://github.com/novus/nvd3/pull/762

ashgillman:
This bug is still referenced by readme.md

robinfhu:
@ashgillman thanks.  We should remove it when we update the readme for the next release.

PR #789, opened 2015-02-12
d3 version fix in README and package.json

Resseguie:
Fixes a syntax typo for D3 peerDependencies in package.json and makes it the same as specified in bower.json .

Also combines duplicate (and wrong) version dependencies in README.

Removes the reference to Interactive guideline tooltip bug that was fixed per #760 that is also related to D3 version.


liquidpele:
Looks good.

PR #767, opened 2015-02-09
Expose utils methods of container sizing and available dimensions

portante:
This is an attempt to make it easier for callers to figure out chart sizing, while reducing duplicate code of such calculations.

The angularjs-nvd3-directives project issues https://github.com/angularjs-nvd3-directives/angularjs-nvd3-directives/issues/56, and https://github.com/angularjs-nvd3-directives/angularjs-nvd3-directives/issues/11, show that initial need.

PR #791, opened 2015-02-12
Fix the tests to use a div

portante:


liquidpele:
Seems good to me... Was there something in particular you were questioning?

portante:
The change to use a `div` came from @robinfhu in #767, and I thought it was good idea.  There was another FIXME that I left in there in #767 that appears to be merged regarding the use of `focusHeight` when `focusEnabled` is not checked.

PR #790, opened 2015-02-12
Fix stackedAreaChart yAxis tick format

floribon:
We cannot override yAxis.tickFormat as it keeps a reference to the initial yAxisTickFormat option.

It is not consistent with other charts, such as Line, where it does work.

It also used to work in the old nvd3 live example page: http://nvd3.org/livecode/index.html#codemirrorNav

The code related to the tickFormat on this chart has changed since the newer versions and bring this regression, which this commit fix.

However this erase the use of yAxisTickFormat, which I am not sure if it is obsolete or supported.

Test case without the fix: http://jsfiddle.net/floribon/xu1qz34u/

PR #795, opened 2015-02-16
Stacked Area: yAxis formatting bug (percentage format sticks)

robinfhu:
Pull Request #790 introduced another bug, where if you switch to 'Expanded'
    and then switch back to 'Stacked', the yAxis retains the percentage format.
    Fixed bug and added unit test.


liquidpele:
Looks good.

PR #798, opened 2015-02-17
Proposal only!  Fix voronoi tooltips to show all siblings for duplicates

portante:
This is a proposal for a fix.  It solves the voronoi problem of duplicate points by eliminating them, but saving the duplicates as siblings that are also displayed in the tooltip.  This should solve problems like #330 once and for all.

For now, we only do that for `lineCharts`.

Please ignore the two debugging commits, and the base change to `lines.js`.

### DIFF
--- a/src/models/lineChart.js
+++ b/src/models/lineChart.js
@@ -54,6 +54,13 @@ nv.models.lineChart = function() {
             y = yAxis.tickFormat()(lines.y()(e.point, e.pointIndex)),
             content = tooltip(e.series.key, x, y, e, chart);
 
+        if (e.sibs) {
+            for (var i = 0; i < e.sibs.length; i++) {
+                x = xAxis.tickFormat()(lines.x()(e.sibs[i].point, e.sibs[i].pointIndex)),
+                y = yAxis.tickFormat()(lines.y()(e.sibs[i].point, e.sibs[i].pointIndex)),
+                content += tooltip(e.sibs[i].series.key, x, y, e.sibs[i], chart);
+            }
+        }
         nv.tooltip.show([left, top], content, null, null, offsetElement);
     };
 
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -141,35 +141,34 @@ nv.models.scatter = function() {
             g.attr('clip-path', clipEdge ? 'url(#nv-edge-clip-' + id + ')' : '');
 
             function updateInteractiveLayer() {
+                // Always clear needs-update flag regardless of whether or not
+                // we will actually do anything (avoids needless invocations).
                 needsUpdate = false;
 
                 if (!interactive) return false;
 
-                var eventElements;
-
-                var vertices = d3.merge(data.map(function(group, groupIndex) {
-                        return group.values
-                            .map(function(point, pointIndex) {
-                                // *Adding noise to make duplicates very unlikely
-                                // *Injecting series and point index for reference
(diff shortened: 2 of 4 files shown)


portante:
@liquidpele, @robinfhu, what do you guys think?

Perhaps too much mechanism?

src/models/lineChart.js
```diff
@@ -54,6 +54,13 @@ nv.models.lineChart = function() {
             y = yAxis.tickFormat()(lines.y()(e.point, e.pointIndex)),
             content = tooltip(e.series.key, x, y, e, chart);
 
+        if (e.sibs) {
```
robinfhu: I'd rename it e.siblings

src/models/scatter.js
```diff
@@ -189,13 +213,19 @@ nv.models.scatter = function() {
...
+                        .filter(function(d, i) {
+                            // Filter out dummy points
+                            return vertices[i][4] !== null;
+                        })
+                        .map(function(d, i) {
+                            return {
+                                'data': bounds.clip(d),
+                                'series': vertices[i][2],
+                                'point': vertices[i][3],
+                                'sibs': (vertices[i].length >= 6) ? vertices[i][5] : null
```
robinfhu: There's heavy usage of magic numbers, which I realize has always been in the code.  But would be nice if it was put in variable

robinfhu:
One thing I like is that this didn't really add much code, it was more of a refactor.  The change should be transparent to end users...most people shouldn't notice any differences.  The difference is that there is now a 'siblings' data property available for the tooltips.

This should be tested pretty thoroughly too.

Nice work.

src/models/lineChart.js
```diff
@@ -54,6 +54,13 @@ nv.models.lineChart = function() {
...
+            for (var i = 0; i < e.sibs.length; i++) {
```
liquidpele: This is a good place to use "for in" instead of having to use the length property, this also makes it able to handle objects if you wanted to switch to that later.

src/utils.js
```diff
@@ -567,31 +567,79 @@ nv.utils.initSVG = function(svg) {
...
-};
+    if (height < 0) {
+        debugger;
+    }
+    if (height > 0) {
+        console.log("height = ", height);
+        return height;
+    }
+    else {
+        var ch = parseInt(container.style('height'));
```
liquidpele: You need to specify the radix for parseInt, i.e. parseInt(containter.style('height'), 10)

It's also probably a good idea to use isNaN afterward for a sanity check too.

src/utils.js
```diff
@@ -567,31 +567,79 @@ nv.utils.initSVG = function(svg) {
...
-};
+    if (width < 0) {
+        debugger;
+    }
+    if (width > 0) {
+        console.log("width = ", width);
+        return width;
+    }
+    else {
+        var cw = parseInt(container.style('width'));
```
liquidpele: same as other parseInt

liquidpele:
So from what I can tell, it looks like the issue this solves is that the tooltips would only return data about a single point if there were multiple points in the same location?

robinfhu:
This PR still looks incomplete, is it in progress?

portante:
@robinfhu, yes, just behind on a number of other things at work.

portante:
@robinfhu, @liquidpele, just getting back to this, probably too late for 1.8.1.

PR #811, opened 2015-02-20
Make it work under atom-shell

alfa07:


portante:
@alfa07, can you resubmit this PR with no `build/*` files changed?

src/core.js
```diff
@@ -1,3 +1,4 @@
+d3 = require('d3');
```
robinfhu: What happens if we include this in the browser normally and 'require' isn't defined? Won't this error out?

robinfhu:
Please explain what this solves. Also, this breaks the tests.

src/core.js
liquidpele: I assume they're actually tryign to use requirejs there?   There is another pull request for AMD, we should probably just make it loadable for requriejs and AMD so people don't have to use a shim.

Issue #781, opened 2015-02-12
I'd like to make a sunburst graph.   They are purty.  


liquidpele:
I'd like to create a sunburst model.  I think these things are supa sweet.  

http://bl.ocks.org/kerryrodden/477c1bfb081b783f80ad

robinfhu:
That surprisingly didn't take a lot of code.  It needs tooltips I think.  Go for it!

liquidpele:
Okay, so just to voice my thoughts, next up is to add the chart with legend and tooltips.   As part of this, I'm going to make a version of the tooltip that just follows next to the mouse and set that to work on pie/donut charts too.  It makes the most sense for those graph types that don't have a specific coordinate to appear in relation to, and should resolve the tooltip issues with piecharts even in IE since the mouse should never be over the tooltip in that case.

liquidpele:
This is completed.  It's basic but easy to extend as needed.

robinfhu:
:+1:

PR #812, opened 2015-02-22
Base of the sunburst chart, make color functions a little more sane and ...

liquidpele:
...have ability to use them as a normal ordinal scale too.

https://github.com/novus/nvd3/issues/781

### DIFF
--- a/src/models/sunburst.js
+++ b/src/models/sunburst.js
@@ -150,6 +150,7 @@ nv.models.sunburst = function() {
         width:      {get: function(){return width;}, set: function(_){width=_;}},
         height:     {get: function(){return height;}, set: function(_){height=_;}},
         id:         {get: function(){return id;}, set: function(_){id=_;}},
+        duration:   {get: function(){return duration;}, set: function(_){duration=_;}},
 
         // options that require extra logic in the setter
         margin: {get: function(){return margin;}, set: function(_){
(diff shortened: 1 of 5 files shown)


liquidpele:
This is pretty basic, the only change that would affect anythign else is with the color functions.  Looking into things more, it looks like you're *supposed* to use an ordinal scale for colors, instead of the (d, i) functions that we've been using, and the sunburst was doing so based on the name of the data instead of any kind of index, so I modified the functions to just use a scale if i isn't passed to the function (meaning get the color by value instead of by index).

src/models/sunburst.js
```diff
@@ -0,0 +1,168 @@
...
+                        x.domain(xd(t));
+                        y.domain(yd(t)).range(yr(t));
+                        return arc(d);
+                    };
+                };
+            }
+
+        });
+
+        renderWatch.renderEnd('pie immediate');
```
robinfhu: should probably say 'sunburst immediate' just for consistency

liquidpele: doh

robinfhu:
This is pretty amazing work @liquidpele .
sunburst

Since it doesn't really affect other charts go ahead an add it.  Also see if you can add a unit test that renders the whole chart, so we get some code coverage on it.  Really cool.

robinfhu:
BTW @liquidpele , have you used the code coverage tool built into the nvd3 build process?
code coverage

You can see it by going to the 'coverage/' folder after running 'grunt'.

liquidpele:
Added some unit tests so it has some level of coverage per robin, merging now.

PR #813, opened 2015-02-22
Issue #410. Replacing this.__data__ with bound variable 'd'. Removing  highlightZero option.

robinfhu:
Adding unit test to check for the 'zero' tick in axes.
Issue #410


liquidpele:
Actually I kind of think we should remove this option altogether and just always apply the zero class to the line, then people can just override with CSS as they need to...

robinfhu:
@liquidpele  I agree actually.  highlightZero is true by default anyway, so I doubt anyone ever turns it off.

robinfhu:
@liquidpele removed the highlightZero option.

PR #818, opened 2015-02-26
Fix donut chart labels.

### LINKED ISSUE #801: Labels are broken on donut chart example
Not sure when it happened, but the donut chart labels, when displayed outside the slices, are all overlapping and on the top.

mgold:
Fixes #801. Only set start and end angles if they're specified. Explicitly equate with `false` because `0` is a valid angle. `false`, on the other hand, is not a valid angle until it gets coerced to `0`. The problem was that the labels were being put on an arc of zero angular length.

### DIFF
--- a/src/models/pie.js
+++ b/src/models/pie.js
@@ -77,11 +77,11 @@ nv.models.pie = function() {
             var arc = d3.svg.arc().outerRadius(arcRadius);
             var arcOver = d3.svg.arc().outerRadius(arcRadius + 5);
 
-            if (startAngle) {
+            if (startAngle !== false) {
                 arc.startAngle(startAngle);
                 arcOver.startAngle(startAngle);
             }
-            if (endAngle) {
+            if (endAngle !== false) {
                 arc.endAngle(endAngle);
                 arcOver.endAngle(endAngle);
             }
--- a/src/models/pie.js
+++ b/src/models/pie.js
@@ -205,7 +205,9 @@ nv.models.pie = function() {
                 }
 
                 if (donutLabelsOutside) {
-                    labelsArc = d3.svg.arc().outerRadius(arc.outerRadius()).startAngle(startAngle).endAngle(endAngle);
+                    labelsArc = d3.svg.arc().outerRadius(arc.outerRadius());
+                    if (startAngle === false) labelsArc.startAngle(startAngle);
+                    if (endAngle === false) labelsArc.endAngle(endAngle);
                 }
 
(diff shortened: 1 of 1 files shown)


src/models/pie.js
```diff
@@ -205,7 +205,9 @@ nv.models.pie = function() {
                 }
 
                 if (donutLabelsOutside) {
-                    labelsArc = d3.svg.arc().outerRadius(arc.outerRadius()).startAngle(startAngle).endAngle(endAngle);
+                    labelsArc = d3.svg.arc().outerRadius(arc.outerRadius());
+                    if (startAngle === false) labelsArc.startAngle(startAngle);
+                    if (endAngle === false) labelsArc.endAngle(endAngle);
```
liquidpele: shouldn't this be !== and not === ?

liquidpele: Actually, the logic above that that affects the actual arcs themselves just does:

if(startAngle) { ... } and if(endAngle) { ... } so we should probably just do the same here.

mgold: Yes, it's !==. Not sure how it worked before. Sorry about that.

The logic above will fail if the user specifies and angle of zero. Should I change that too?

src/models/pie.js
```diff
@@ -205,7 +205,9 @@ nv.models.pie = function() {
                 }
 
                 if (donutLabelsOutside) {
-                    labelsArc = d3.svg.arc().outerRadius(arc.outerRadius()).startAngle(startAngle).endAngle(endAngle);
+                    labelsArc = d3.svg.arc().outerRadius(arc.outerRadius());
+                    if (startAngle !== false) labelsArc.startAngle(startAngle);
+                    if (endAngle !== false) labelsArc.endAngle(endAngle);
```
liquidpele: Well, the logic should match, so we can either set these to be if(startAngle) or change the other locations to check for !== false, up to you.

mgold:
Okay, done. And is succinct as `if (startAngle)` is, it's more important to accept zero.

liquidpele:

mgold:
No problem! I'm going to see if I can hunt down the 100% and make it respond to filters.

liquidpele:
Well, the "100%" is just supposed to be a text title the same as "woot" on the other one, I added that a while back.  What did you have in mind to make it do besides just be text?

mgold:
See #819. Changing the title to some other string literal would certainly be the easiest solution, unless you want dynamically updating titles.

PR #821, opened 2015-02-26
Added a tooltip to sunburst.  

liquidpele:
I tried to re-use the existing tooltip code, but it's way too specific for x-y coordinates and having a mouse following version is stupid simple anyway since it doesn't need any of the extra complexity.

### DIFF
--- a/src/models/sunburstChart.js
+++ b/src/models/sunburstChart.js
@@ -1,0 +1,157 @@
+nv.models.sunburstChart = function() {
+    "use strict";
+
+    //============================================================
+    // Public Variables with Default Settings
+    //------------------------------------------------------------
+
+    var sunburst = nv.models.sunburst();
+
+    var margin = {top: 30, right: 20, bottom: 20, left: 20}
+        , width = null
+        , height = null
+        , color = nv.utils.defaultColor()
+        , tooltips = true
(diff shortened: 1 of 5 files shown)


src/models/sunburstChart.js
```diff
@@ -0,0 +1,146 @@
...
+        , noData = null
+        , duration = 250
+        , dispatch = d3.dispatch('tooltipShow', 'tooltipHide', 'stateChange', 'changeState','renderEnd')
+        ;
+
+    //============================================================
+    // Private Variables
+    //------------------------------------------------------------
+
+    var tooltip = d3.select('body')
```
robinfhu: Honestly, I really like this way of doing tooltips.

robinfhu:
No real objections, keep up the awesome work!

liquidpele:
Cool, merging in.

liquidpele:
@robinfhu what do you like about that tooltip code?  I'm planning to convert the pie chart to similar so I was just curious :)

robinfhu:
First, I've always supported the idea of putting the div on the body, because it is much easier to absolutely position it.  All you need to do then is position the top and left using:

hoverElement.getBoundingClientRect().top + window.pageYOffset
hoverElement.getBoundingClientRect().left

Issue #799, opened 2015-02-18
PieChart tooltips flicker on and off in IE10


liquidpele:
It looks like IE10 has an issue where the tooltip will flicker when the mouse is over the tooltip. 

This was fixed in a pull request here:  https://github.com/novus/nvd3/pull/453

However, the pull request changed a lot of code and it's probably better to review the tooltip position for pieCharts so that it doesn't display under the mouse so that the issue is just avoided altogether instead of having a hack just for IE10.  This issue is to track this bug in the meantime.

liquidpele:
Should be fixed.

PR #822, opened 2015-02-27
Change pie/donut charts to use a mouse following tooltip

liquidpele:
Also cleaned up the code a little, and changed the labelsOutside options to just be a single option and added depreciated stubs for the old options that gives console warnings.  Also changed it so depreciated warnings show up even if not in dev mode since they're pretty important.

Current charts that override tooltipContent will see their tooltips break though.   There isn't a lot I can do about that since the way it was passing the data was stupid because it was picking out parts instead of just passing the whole event object.

### DIFF
--- a/src/models/pieChart.js
+++ b/src/models/pieChart.js
@@ -15,10 +15,13 @@ nv.models.pieChart = function() {
         , legendPosition = "top"
         , color = nv.utils.defaultColor()
         , tooltips = true
-        , tooltip = function(key, y, e, graph) {
-            return '<h3 style="background-color: '
-                + e.color + '">' + key + '</h3>'
-                + '<p>' +  y + '</p>';
+        , tooltipContent = function(evt) {
+            return '<table>' +
+                '<tbody><tr><td class="legend-color-guide">' +
+                '<div style="background-color: ' + evt.color + '"></div></td>' +
+                '<td class="key"> ' + pie.x()(evt.data) + '</td>' +
+                '<td class="value">' + pie.y()(evt.data) + '</td>' +
+                '</tr></tbody></table>';
         }
         , state = nv.utils.state()
         , defaultState = null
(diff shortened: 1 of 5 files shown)


liquidpele:
This also should fix https://github.com/novus/nvd3/issues/799 since the tooltip won't appear under the mouse anymore.

src/models/pieChart.js
```diff
@@ -193,13 +184,23 @@ nv.models.pieChart = function() {
...
+    });
+
+    pie.dispatch.on('elementMousemove.tooltip', function(evt) {
+        tooltip.style('top', (event.pageY-15)+"px").style("left",(event.pageX+10)+"px");
     });
 
-    pie.dispatch.on('elementMouseout.tooltip', function(e) {
-        dispatch.tooltipHide(e);
+    pie.dispatch.on('elementClick.tooltip', function(evt) {
+        console.log("caught click! ", evt);
```
robinfhu: remove console.log

src/models/pieChart.js
```diff
@@ -193,13 +184,23 @@ nv.models.pieChart = function() {
...
+        tooltip.html(tooltipContent(evt))
+            .style('display', 'inline-block');
```
robinfhu: is it necessary to set the display to inline-block here?

src/models/pieChart.js
```diff
@@ -193,13 +184,23 @@ nv.models.pieChart = function() {
...
+    });
+
+    pie.dispatch.on('elementMouseout.tooltip', function(evt) {
+        tooltip.style('display', 'none');
+    });
+
+    pie.dispatch.on('elementMousemove.tooltip', function(evt) {
+        tooltip.style('top', (event.pageY-15)+"px").style("left",(event.pageX+10)+"px");
```
robinfhu: the 15 and 10 numbers here could be options like 'tooltip distance'

src/models/pieChart.js
robinfhu: Oh nevermind, this is to make the tooltip appear.

src/models/pieChart.js
liquidpele: ugh, I've got to start searching for those first!

src/models/pieChart.js
liquidpele: I considered that but I intentionally didn't since I'd like to get into some of this other tooltip code.  Ideally I'd like all the charts to use a tooltip object that has it's own options like the legend.

robinfhu: I see. Thanks!

robinfhu:
Nice, saw the behavior. Great improvement and hopefully will reduce the tooltip issues!
piechart tooltip

Issue #823, opened 2015-02-27
xAxis.axisLabel not centered for multiBarHorizontalChart


vbsql7:
The left-hand vertical text is the issue. It shows up, but at the very top of the chart instead of being nicely centered like the bottom axis label is.

In the code, the xAxis and yAxis are opposite because this is a 'horizontal' bar chart. 

![screen shot 2015-02-27 at 3 31 01 pm](https://cloud.githubusercontent.com/assets/9969712/6420641/cd7733de-be95-11e4-820c-492db178acd3.png)

```
<!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8">
    <link href="nv.d3.css" rel="stylesheet" type="text/css">
    <script src="d3.min.js" charset="utf-8"></script>
    <script src="nv.d3.min.js"></script>

    <style>
        text {
            font: 12px sans-serif;
        }
        svg {
            display: block;
        }
        html, body, #chart1, svg {
            margin: 5px;
            padding: 0px;
            height: 100%;
            width: 100%;
        }
    </style>
</head>
<body>

<div id="chart1" >
<svg></svg>
</div>
<script>

// Create fake data stream
function generateData ( ) {
  var numberOfPoints = 10,
      graphData = [],
      j;

    series = [];
    for (j = 0; j < numberOfPoints; j++) {
      graphData.push({label: 'Item ' + String(j), value: (Math.random() * 100)});
    }
  
  return graphData;
}

// Format data stream as JSON
var dataToGraph = [
  {
    key: 'Series 1',
    color: '#1f77b4',
    values: generateData()
  }
]

// Create chart 
var chart;
var svg;

svg = d3.select('#chart1 svg');

nv.addGraph(function() {

  chart = nv.models.multiBarHorizontalChart()
  .x(function(d) { return d.label })
  .y(function(d) { return d.value })
  .noData('Not enough data to graph')
  .barColor(['lightsalmon', 'indianred'])
  .duration(250)
  .showValues(true)
  .showControls(false)
  ;

  // Add Y axis label
  chart.yAxis
    .axisLabel('Y axis label')
    .tickFormat(d3.format(',.0f'))
    ;

  chart.xAxis.axisLabel('Things'); // not centered, but present and floats with resize

  svg.datum(dataToGraph).call(chart);

  nv.utils.windowResize(chart.update);

  chart.dispatch.on('stateChange', function(e) { nv.log('New State:', JSON.stringify(e)); });

[...]

PR #824, opened 2015-02-27
Fixes https://github.com/novus/nvd3/issues/823

liquidpele:
by getting the max value in scale instead of assuming [0] was the max when calculating the x offset for the axis label.

I did a little code cleanup too, based on warnings in my IDE.  

Also set the example to use axis labels there.

### DIFF
--- a/src/models/axis.js
+++ b/src/models/axis.js
@@ -69,6 +69,7 @@ nv.models.axis = function() {
             axisLabel.exit().remove();
 
             var xLabelMargin;
+            var axisMaxMin;
             var w;
             switch (axis.orient()) {
                 case 'top':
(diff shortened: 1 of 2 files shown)


src/models/axis.js
```diff
@@ -84,7 +85,7 @@ nv.models.axis = function() {
                         .attr('y', 0)
                         .attr('x', w/2);
                     if (showMaxMin) {
-                        var axisMaxMin = wrap.selectAll('g.nv-axisMaxMin')
```
robinfhu: Why did you remove 'var'?

src/models/axis.js
```diff
@@ -217,12 +217,13 @@ nv.models.axis = function() {
                      if(labelPadding > width) width = labelPadding;
                      });
                      */
+                    console.log('debug: ', scale.range());
```
robinfhu: remove console.log

robinfhu:
I think you need to add back 'var axisMaxMin', otherwise you create a global variable.

src/models/axis.js
liquidpele: meant to move the declaration to the top of the switch statement.  It was doing duplicate declarations.  Fixed now.

robinfhu:
Thanks!

Issue #771, opened 2015-02-10
Tooltips: nvtooltip-pending-removal div is not removed on multiBarChart


robinfhu:
On the development branch, when I open the file examples/multiBarChart.html, I mouse over the chart and I see the following:

![screen shot 2015-02-10 at 3 24 23 pm](https://cloud.githubusercontent.com/assets/3769532/6135754/082d0b76-b139-11e4-8e99-88473630f776.png)

When I inspect the DOM, there are lots of divs with 'nvtooltip-pending-removal' classed to them.

robinfhu:
@petersondrew can you please take a look?

RenatoUtsch:
I know that the bar charts use the old tooltip code, not the code that are used by the charts with interactive guideline (like the line chart).

petersondrew:
Hmm, I'll take a look when I get into the office tomorrow.

robinfhu:
Thanks. This is happening for pretty much any chart with the regular hover tooltip.

petersondrew:
Alright, either I'm going crazy, or we're not comparing apples to apples. I just synced roadranger/nvd3/development with novus/nvd3/development, and when I open the multiBarChartTest.html under test I do not get this behavior at all, on chrome, ff, or IE11.

What browser are you testing in? Do you have any local changes, particularly with the nvd3 css file, that may be contributing?

petersondrew:
Just in case it's somehow relevant, I appear to be on d3 3.5.4 locally.

robinfhu:
@petersondrew , is it possible you have changes in roadranger/nvd3/development that were not merged in to our branch?  You may have a different environment.

petersondrew:
I don't believe so. I cloned to roadranger/nvd3 yesterday and only committed the changes for #765. This morning I did a pull request from novus/nvd3 to sync my development branch with upstream.

I'm going to pull novus/nvd3 down as well and test there.

petersondrew:
@robinfhu I just pulled novus/nvd3 and on development branch I do not see any problems with the tooltips on the multibarchart.html page. I tried under /examples and /test.

Can you please clone development into a clean folder, do a bower install, and see if you see the same issues?

robinfhu:
@petersondrew , the reason is because you need to run 'grunt' once before loading the test pages.

Grunt creates build/nv.d3.js.  In development, we don't checkin changes to that file, so you are using an older version of that.

petersondrew:
[...]

PR #825, opened 2015-02-27
Issue 771 Tooltip pending removal fix

robinfhu:
Issue #771 fix.
I brought back some code that was removed from nv.tooltip.cleanup()

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -492,10 +492,20 @@
         // Find the tooltips, mark them for removal by this class (so others cleanups won't find it)
         var tooltips = document.querySelectorAll('.nvtooltip');
         if (tooltips) {
+
             nv.dom.write(function() {
+                var removeQueue = [];
                 for (var i = 0; i < tooltips.length; i++) {
+                    removeQueue.push(tooltips[i]);
                     tooltips[i].className = 'nvtooltip-pending-removal';
                 }
+
+                window.setTimeout(function() {
+                    while(removeQueue.length) {
+                        var toRemove = removeQueue.pop();
+                        toRemove.parentNode.removeChild(toRemove);
+                    }
+                },200);
             });
         }
     };
--- a/test/mocha/tooltips.coffee
+++ b/test/mocha/tooltips.coffee
@@ -1,0 +1,14 @@
+describe 'NVD3', ->
+    describe 'Tooltips', ->
+        it 'removes tooltip from document', (done)->
+            d3.select('body')
+            .append('div')
+            .classed('nvtooltip', true)
+
+            nv.tooltip.cleanup()
+
+            window.setTimeout ->
+                d3.select('.nvtooltip').empty().should.be.true
+                d3.select('.nvtooltip-pending-removal').empty().should.be.true
+                done()
+            , 500


robinfhu:
@petersondrew @liquidpele please take a look.

petersondrew:
The idea behind my changes, and the reason that code was the way it was, is because the interactive layer re-uses that container div. Setting the css class just makes it invisible. (I even added tests for it)

Could you try perhaps changing the classname that the interactive layer uses to nvtooltip-hidden or something perhaps? If the class is applied in the interactive layer, and the css changed, then you can add this removal code back in as it will only remove those with nvtooltip-pending-removal and not nvtooltip-hidden.

Thoughts?

robinfhu:
@petersondrew let me take a look. Thanks.

robinfhu:
@petersondrew I see this line of code:

           var hidden = document.querySelector("#" + id + ".nvtooltip-pending-removal");
            if (hidden)
                hidden.className = "nvtooltip " + (classes? classes: "xy-tooltip");

This confuses me.  I'm trying to understand what this does but I can't.

Right now with my code change the behavior is exactly what I want.

liquidpele:
So wait, just glancing at this, it looks like it'll remove all tooltips, what if there are multiple graphs on the screen with tooltips being shown?

liquidpele:
var hidden = document.querySelector("#" + id + ".nvtooltip-pending-removal");
        if (hidden)
            hidden.className = "nvtooltip " + (classes? classes: "xy-tooltip");

It's finding anything with the .nvtooltip-pending-removal class with ALSO has an id attribute with the value of the id variable.  Then if it finds at least one, it changes the class attribute to be "nvtooltip xy-tooltip".

liquidpele:
Like I said in the defect, I think it might be better to just save the dom object outside the tooltip call function, and then when we need to find it again we can just do like tooltip_dom.node() and manipulate it as needed... that would remove all the unnecessary id and class selection stuff, and would make it so each chart had it's own encapsulated tooltip instead of them stepping over each other.

robinfhu:
@liquidpele Oh, I see what you are saying.  I was basically copying old functionality, which behaved like I expected.  

Can @liquidpele  or @petersondrew PR a solution?

liquidpele:
I'll try to look at it tonight

robinfhu:
@liquidpele  thanks :+1: :-)

liquidpele:
yea, so I looked at it last night, seems that nothing except the interactive guideline uses the tooltip object, everything else just uses a function to display the tooltip... so I'm giving this a little more thought, but I'm still working on it.

robinfhu:
This is the last remaining issue I think we should solve. After this, we can release what's in development.

liquidpele:
Just an update, I've got a chart using the tooltip object now instead of the "old" tooltips, just a few little style kinks to work out, but I've modified a lot so any other tooltip updates will likely just get overwritten by this.

robinfhu:
SInce @liquidpele has taken over this, I am closing this PR.

pimlinders:
@liquidpele it would be really nice if we could provide custom content when using "useInteractiveGuideline" as well. Note that on version 1.8.1 when "useInteractiveGuideline" is set to the true the guidelines remains on the last x axis position no matter the mouse position.

liquidpele:
You can do chart.interactive guideline.tooltip.content generator()

Please open an issue for any bugs :)

pimlinders:
@liquidpele you rock! Looked further into the mouse hover issue, seems to be something with d3 version 3.3.13. Updating d3 resolved the issue.

liquidpele:
Oh yea, we require 3.4.4+ now due to some issues we saw with previous versions... the readme is only updated with that on the development branch and the 1.8.1alpha though :)

PR #828, opened 2015-02-28
Issue 555: Pie label formatting (also removing 'labelFormat')

robinfhu:
For issue #555

### DIFF
--- a/src/models/pie.js
+++ b/src/models/pie.js
@@ -273,7 +273,7 @@ nv.models.pie = function() {
                         var percent = (d.endAngle - d.startAngle) / (2 * Math.PI);
                         var labelTypes = {
                             "key" : getX(d.data),
-                            "value": getY(d.data),
+                            "value": valueFormat(getY(d.data)),
                             "percent": labelFormat(percent)
                         };
                         return (d.value && percent > labelThreshold) ? labelTypes[labelType] : '';
--- a/src/models/pie.js
+++ b/src/models/pie.js
@@ -350,6 +350,7 @@ nv.models.pie = function() {
             labelsOutside=_;
             nv.deprecated('donutLabelsOutside', 'use labelsOutside instead');
         }},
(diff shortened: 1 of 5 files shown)


robinfhu:
Hold on, I plan to remove 'labelFormat' from pie.js...

robinfhu:
@liquidpele this should be ready now.

src/models/pie.js
```diff
@@ -341,7 +348,11 @@ nv.models.pie = function() {
         // depreciated after 1.7.1
         donutLabelsOutside: {get: function(){return labelsOutside;}, set: function(_){
             labelsOutside=_;
-            nv.deprecated('pieLabelsOutside', 'use labelsOutside instead');
+            nv.deprecated('donutLabelsOutside', 'use labelsOutside instead');
+        }},
+        labelFormat: {get: function(){ return valueFormat;}, set: function(_) {
```
liquidpele: Can you add a comment like the others so we know what the version was before it was deprecated?  Otherwise it's easy to lose track and never remove the things...

robinfhu: Yep.

liquidpele:
So, glancing at the code, valueFormat was previously not even used anywhere, but labelFormat was, so why are we opting to use valueFormat?   Plus, since labelType is an option, labelFormat keeps the same prefix and seems clearer to me.   Thoughts?

robinfhu:
@liquidpele , valueFormat is the standard for other charts like discreteBar and multiBar, so it's for consistency.  

The person who opened issue #555 wanted to use valueFormat, and he was correct in pointing that out.

robinfhu:
@liquidpele , basically, in all other charts, 'valueFormat' is used for formatting labels that appear inside the chart.  So it should be the same for pieChart.

liquidpele:
Ah okay, sounds good then.

robinfhu:
Thanks :-)

Issue #816, opened 2015-02-25
Focus Chart Path Transition Wiggle


eschlon:
Trying to build a Line with Focus chart and noticed an annoying wiggle when scrubbing through the data (essentially [this](http://bost.ocks.org/mike/path/) problem). The issue is not present on the example on nvd3.org (and is not present locally when I grab the js used there). It is present in both master and dev branches here. You can replicate it by just loading the lineWithFocusChart.html example in either branch, zooming in and scrubbing through the chart.

robinfhu:
Yeah something changed in the line with focus code during all our merges.  You're right, it used to never do that.

liquidpele:
This was probably due to me combining the linePlusBar and linePlusBarWithFocus charts... maybe the latter had a specific interpolation set?

robinfhu:
But why would that affect the lineWithFocus chart?

liquidpele:
Oh whoops I got confused.  No idea then, but lineplusbar does it too anyway...

robinfhu:
Did a git bisect, and this started happening after @fshao816 added the 'renderWatch' functionality.  Looking into this...

robinfhu:
The only solution I have is to turn of transition duration for the lines.  Let me pull request this.

robinfhu:
@eschlon PR opened.

PR #829, opened 2015-02-28
Issue #816:  lineWithFocusChart, fixing the 'wiggling'

robinfhu:
Change the lines.duration to 0, to remove transitions and prevent weird things from happening when changing the focus area.

Issue #816


robinfhu:
Taggin @eschlon

liquidpele:
Tested this out and seems to work well :)

PR #838, opened 2015-03-03
Fixing long standing bug in scatter.js, where the nv-single-point class ...

robinfhu:
...would never go away even if the data was updated with more than one point.

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -94,12 +94,7 @@ nv.models.scatter = function() {
                 .range(sizeRange || _sizeRange_def);
 
             // If scale's domain don't have a range, slightly adjust to make one... so a chart can show a single data point
-            if (x.domain()[0] === x.domain()[1] || y.domain()[0] === y.domain()[1]) {
-              singlePoint = true;
-            }
-            else {
-              singlePoint = false;
-            }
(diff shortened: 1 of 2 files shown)


src/models/scatter.js
```diff
@@ -94,7 +94,13 @@ nv.models.scatter = function() {
                 .range(sizeRange || _sizeRange_def);
 
             // If scale's domain don't have a range, slightly adjust to make one... so a chart can show a single data point
-            if (x.domain()[0] === x.domain()[1] || y.domain()[0] === y.domain()[1]) singlePoint = true;
+            if (x.domain()[0] === x.domain()[1] || y.domain()[0] === y.domain()[1]) {
+              singlePoint = true;
+            }
+            else {
+              singlePoint = false;
+            }
```
joshjordan:
```
if (condition)
{ true }
else { false }
```
is a bit of a convoluted pattern.

Could simply be:
```
singlePoint = (x.domain()[0] === x.domain()[1] || y.domain()[0] === y.domain()[1]);
```

robinfhu: Oh yeah, thanks!

robinfhu:
@liquidpele can we get this merged? Thanks.

PR #841, opened 2015-03-04
Pie title supports resizing.

mgalloue:
Adjust automatically the font-size of the title, with respect to the current donut chart size. 
- It's not necessary to define a size in a css. 
- It supports resizing.

### DIFF
--- a/examples/donutChart.html
+++ b/examples/donutChart.html
@@ -18,6 +18,11 @@
             height: 350px !important;
             width: 350px !important;
         }
+		#test3 {
+			height: 500px !important;
+            width: 500px !important;
+		}
+		
         html, body {
             margin: 0px;
             padding: 0px;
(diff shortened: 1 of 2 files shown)


examples/donutChart.html
```diff
@@ -18,6 +18,11 @@
             height: 350px !important;
             width: 350px !important;
         }
+		#test3 {
```
robinfhu: Can we fix the indenting?

robinfhu:
The problem with your solution is it's not one size fits all.  If I make the donut title just a few characters bigger, the autosizing logic doesn't work:

screen shot 2015-03-04 at 11 17 25 am

robinfhu:
The code isn't taking into account the length of the text.  I am going to close this.  Feel free to reopen if you have any updates!

PR #842, opened 2015-03-04
Pie title supports resizing.

mgalloue:
Adjust automatically the font-size of the title, with respect to the current donut chart size.

- It's not necessary to define a size in a css.
- It supports resizing.
- it depends on the title length
- it takes into account the donut ratio.

### DIFF
--- a/examples/donutChart.html
+++ b/examples/donutChart.html
@@ -18,6 +18,11 @@
             height: 350px !important;
             width: 350px !important;
         }
+		#test3 {
+			height: 500px !important;
+            width: 500px !important;
+		}
+		
         html, body {
             margin: 0px;
             padding: 0px;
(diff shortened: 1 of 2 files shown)


robinfhu:
Will look at this in about an hour.

robinfhu:
Thanks!

mikyladewitt:
Hello,

How do we use this? This feature is perfect for a problem I am trying to solve, but I don't see how to turn it on.

Thanks!

liquidpele:
If you just want to resize the inner text, you can modify the CSS, ala the class name on it:  .nv-pie-title

mikyladewitt:
@liquidpele ah I may have misunderstood which title this was referring to. I was looking for a built in way to modify the chart title, which I just realized is actually angular-nvd3 specific. Sorry for the confusion!

PR #844, opened 2015-03-04
Updated version strings to 1.8.1-alpha tag.

rkofman:
Since there's a tagged release, I figured it's necessary to update the version strings to make sure developers that are riding the alpha wave aren't confused (if they use the built files, their teammates would think they're using 1.7.1 -- causing all sorts of havoc and debugging hell).

Not sure how to request that this change get pulled into the 1.8.1-alpha tag, but that's certainly part of the intent.


rkofman:
Closing, will re-open against the "development" branch.

robinfhu:
Yes, against dev plz

robinfhu:
Yeah sorry, I forgot to do this.

rkofman:
Unfortunately, I seem to be failing to generate the correct diff: when doing `grunt release`, off the 1.8.1-alpha tag, the diff in the minimized js and css files seems to be more than just the version string comment.

Would you mind just doing it for me?

robinfhu:
Yep, looking at it now.

robinfhu:
This is done. The release tag was force moved to the newer version.

rkofman:
thanks!

PR #847, opened 2015-03-05
Line supports line pattern style.

mgalloue:
Possibility for a line to have a line pattern style. 
The user can choose to define a style or not.

There is 11 styles available:
- None
- Solid
- Dense
- Medium
- Sparse
- DashShort
- DashMedium
- DashLong
- DashDot
- DashDotDot
- DashDashDot

### DIFF
--- a/examples/line.html
+++ b/examples/line.html
@@ -90,7 +90,8 @@
                 values: cos,
                 key: "Cosine Wave",
                 color: "#2ca02c",
-                strokeWidth: 3
+                strokeWidth: 3,
+                style: "Sparse"
             }
         ];
     }
--- a/src/css/lines.css
+++ b/src/css/lines.css
@@ -39,8 +39,50 @@
     fill-opacity: 0;
 }
 
-
-
 .nvd3 .nv-indexLine {
     cursor: ew-resize;
+}
+
+.nvd3 .nv-line-pattern-None {
+stroke-opacity:0 !important;
+}
+
+.nvd3 .nv-line-pattern-Solid {
+stroke-opacity: 1;
+}
+
+.nvd3 .nv-line-pattern-Dense {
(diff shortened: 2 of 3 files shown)


mgalloue:
An other approach is to customize directly the stroke-dasharray, there is no need to add properties on css and the user is free to create the line pattern he want but it may be less intuitive.

robinfhu:
This is a good idea, but I think the wrong approach. 

Users can already write their own CSS to customize how their lines look, this feature doesn't really add anything to NVD3.

It also bloats our CSS file with a bunch of different configurations which will need to be maintained.  I'm closing this PR for that reason, but feel free to post your thoughts.

liquidpele:
@mgalloue can you make a PR to only have a single extra attribute on the series data called "dashed" which can be true or false, and then set a class to make it a simple dashed line?  Beyond that people should write their own CSS based on the unique classes already put on the line containers, but I think a dashed line is something so commonly used that it's worth putting in that functionality...

liquidpele:
Re-opening this for now, pending an update to the functionality.  

Also, thanks for contributing :)

mgalloue:
Sure, I agree with you but how I can customize my line without any modifications on nvd3 side? Here I put a new class, how I can do that out of NVD3?
         groups
-                .attr('class', function(d,i) { return 'nv-group nv-series-' + i })
+                .attr('class', function(d,i) { return 'nv-line-pattern-'+ d.style + ' nv-group nv-series-' + i })

Also were can I post my thoughts without send a PR?

Thanks,

robinfhu:
@mgalloue , we need the ability to set a custom class for each series.  So your PR has the right idea.  Let me make a modification to it.

robinfhu:
Opened PR #848  @mgalloue

PR #850, opened 2015-03-06
Refactor tooltips for great good

liquidpele:
This is the massive refactor of the tooltips.   This all started due to us needing to keep track of the tooltip dom element, but all the charts were still using a single function which made it hard to track, so I converted them all to using the tooltip object that the interactive guideline uses.  

A few things to note:

1. This might break anyone using a customized tooltip, because I changed the data format to be more generic in many cases (needed extra info that it wasn't giving and then tried to make that consistent). 
2. best I can tell, stackedArea tooltips were already broken, so I fixed those. 
3. The tooltip object is put onto the chart object, so you can easily modify your tooltips like you do the axis or legend. 
4. I deprecated the "tooltips" and "tooltipContent" options on the charts themselves. 

I'm sorry I didn't add tests around this, but I burnt myself out a little bit so I'm kind of sick of tooltips :/


liquidpele:
Shit, stupid github defaulting to master branch -_-

PR #851, opened 2015-03-06
Refactor tooltips for great good

liquidpele:
Trying this again

This is the massive refactor of the tooltips.   This all started due to us needing to keep track of the tooltip dom element, but all the charts were still using a single function which made it hard to track, so I converted them all to using the tooltip object that the interactive guideline uses.  

A few things to note:

1. This might break anyone using a customized tooltip, because I changed the data format to be more generic in many cases (needed extra info that it wasn't giving and then tried to make that consistent). 
2. best I can tell, stackedArea tooltips were already broken, so I fixed those. 
3. The tooltip object is put onto the chart object, so you can easily modify your tooltips like you do the axis or legend. 
4. I deprecated the "tooltips" and "tooltipContent" options on the charts themselves. 

I'm sorry I didn't add tests around this, but I burnt myself out a little bit so I'm kind of sick of tooltips :/   

I know it's a lot of changes, so feel free to just try it out and test it yourself along with the code review... get the "fix_tooltips" branch from my personal repo and build it.


robinfhu:
I'm definitely interested in looking at this and testing it thoroughly.  Looks like some unit tests broke, but i can help fix those.

robinfhu:
Some testing guidelines: 
* make sure to check tooltips in all browsers. 
* check all the example pages.  Look for edge cases like charts in scroll windows.

robinfhu:
At first glance, I like what I see. Tooltips badly needed an overhaul. 
One advantage of this technique is that it makes tooltips so much more customizable. 

Also noticed you removed the X and Y axis tooltips from scatterChart.  Some people might balk at that, but I never cared for those.  They were hard to maintain, and it is better UX to put the info in one main tooltip.

robinfhu:
Had a feeling this would be a problem.  I just checked the lineChartTest.html file, regular tooltips aren't positioned right.  I think window.pageYOffset needs to somehow get taken into account.
tooltip behavior 1

robinfhu:
The tooltip never disappears after leaving the stacked area chart.
stacked area behavior 1

robinfhu:
Tooltips seem broken in stacked area charts too.  This was from examples/stackedAreaChart.html
stacked area behavior 2

joshjordan:
Do we consider custom tooltips (that is, those that might be broken by this
change) to be a part of the public API?

robinfhu:
The public API was kept, but a deprecated warning was added for it.  It's on us to communicate the API change.

src/tooltip.js
```diff
@@ -32,16 +29,21 @@
...
-            ,distance = 50   //Distance to offset tooltip from the mouse location.
-            ,snapDistance = 25   //Tolerance allowed before tooltip is moved from its current position (creates 'snapping' effect)
+            ,   distance = 25   //Distance to offset tooltip from the mouse location.
+            ,   snapDistance = 0   //Tolerance allowed before tooltip is moved from its current position (creates 'snapping' effect)
             ,   fixedTop = null //If not null, this fixes the top position of the tooltip.
             ,   classes = null  //Attaches additional CSS classes to the tooltip DIV that is created.
-            ,   chartContainer = null   //Parent DIV, of the SVG Container that holds the chart.
+            ,   chartContainer = null   //Parent dom element of the SVG that holds the chart.
+            ,   hidden = true  // start off hidden, toggle with hide/show functions below
+            ,   hideDelay = 400  // delay before the tooltip hides after calling hide()
```
robinfhu: hideDelay is a nice feature, however I think defaulting it to 0 is better, since it more closely matches how tooltips behave now (which is to disappear immediately).  But we should let people know the option exists.

liquidpele: it is required for certain charts or else you get flickering of the tooltip some other charts like bar charts or pie charts we could set it 0

src/models/pieChart.js
```diff
@@ -185,21 +178,20 @@ nv.models.pieChart = function() {
...
     });
 
     pie.dispatch.on('elementMousemove.tooltip', function(evt) {
-        tooltip.style('top', (event.pageY-15)+"px").style("left",(event.pageX+10)+"px");
-    });
-
-    pie.dispatch.on('elementClick.tooltip', function(evt) {
-        tooltip.html(tooltipContent(evt))
-            .style('display', 'inline-block');
+        tooltip.position({top: d3.event.pageY, left: d3.event.pageX})();
```
robinfhu: I would definitely check this in other browsers, especially IE.  event.pageY seems to have different standards across browsers.

src/tooltip.js
robinfhu: Which charts have the flickering issue?

liquidpele:
The tooltips on stackedarea are a little broken due to this bug:

https://github.com/novus/nvd3/issues/330

If you disable the lines that look like they overlap points of other lines, then it'll start working right.

robinfhu:
@liquidpele regarding stacked area charts, wow you are right, I forgot about that bug.  I think that's why I created the newer tooltip back then!

joshjordan:
Right, but part of communicating that change is versioning properly. If
we're introducing backwards-incompatibility, we should be bumping a major
version.

robinfhu:
@joshjordan yeah, this change is probably big enough to be a version bump, unless we think it's worth it to support the old functionality?

liquidpele:
I added an example page for the tooltip... I couldn't get it to show the tooltip in the wrong place though?  Is there some trick to making it happen?

src/tooltip.js
liquidpele: linePlusBarChart was the worst, there are spaces between the bars that cause it.  However, even the line charts were jumpy if you didn't take care to go from one point to another or the points were far enough apart.  

I had to do a similar thing a long time ago when crating a tooltip over jqplot graphs... due to tiny gaps here and there it's a lot safer to keep a small delay there for hiding.  I bumped up the default a bit since I thought it looked smoother, if you like it hiding faster we can bump it down some though, we just can't really remove it entirely.

robinfhu:
@liquidpele I can't figure it out in tooltip.html either.  Try opening test/lineChartTest.html.  Maybe the div needs to be floated left?

robinfhu:
@liquidpele , I have a suspicion as to what the issue is.

All the charts where tooltips are broke have one thing in common: they use scatter.js as a base.  You're not using event.pageY/X in scatter.js.

liquidpele:
Okay, I think things should work now :ghost:

robinfhu:
@liquidpele yes it works!

robinfhu:
@liquidpele can you figure out why the stackedarea tooltips don't disappear?

liquidpele:
Fixed.

robinfhu:
@liquidpele great!

robinfhu:
Be nice to get a second set of eyes on this. Taggin @portante

portante:
I'll take a look at this later today, @robinfhu, @liquidpele.

robinfhu:
@liquidpele  Found a bug in IE 11.   When using the interactive guideline tooltip, and mousing over one of the points, the tooltip flickers.
line chart ie bug

liquidpele:
Is that a new bug?  I tried not to change the guideline stuff...

robinfhu:
It may not have been introduced by your change.  It used to not happen though, I checked version 1.2.  It's worth looking into.

liquidpele:
It was happening in 1.7.0, I pushed up, basically setting point-events to none on the points if the scatter has the "interactive" option set to false (when is done automatically if you turn on the interactive guideline).   Some of these options are confusing :/

robinfhu:
@liquidpele thanks for the IE fix.  Things are looking good now. 

How do you feel about bumping the major version of NVD3 after this merge?

portante:
@robinfhu, meaning all other changes will be for the next major version?  And we'll have a branch for this line?

liquidpele:
I think either 1.8.0 or 2.0.0 is fine, but I usually do api changes in the second number, major changes in the first, and bug fixes in the last.

robinfhu:
@portante @liquidpele , how does this look?  I'm good with merging this, but just wanted to get a second set of eyes.

liquidpele:
If no one objects in the next 12 hours lets merge it.

robinfhu:
The countdown begins!

portante:
@liquidpele, @robinfhu, so I am behind and there are a lot of details in this patch set that I don't quite understand at first blush.  So don't wait for me, I'll muddle through it over time.

robinfhu:
@liquidpele good job :+1:

RenatoUtsch:
If you removed the code that displays the old tooltips, what good does keeping tooltip and tooltipContent as deprecated will do? They don't work anymore, do they?

robinfhu:
@RenatoUtsch  it produces an error warning for the user triggering them to update their code.

RenatoUtsch:
Understood.
@liquidpele great job.

RenatoUtsch:
Another thing, now with shadow DOM, the line and stacked area charts are working perfectly, but in the cumulative line chart the tooltip is not disappearing and in the others the tooltip isn't even shown.

I'll look at why this is happening, but now it should be much easier to fix.

robinfhu:
@RenatoUtsch please checkout https://github.com/novus/nvd3/releases/tag/v1.8.1-alpha and help test the tooltip functionality. Thanks!

RenatoUtsch:
@robinfhu will do that!

RenatoUtsch:
Do I report the bugs I find here or in another place?

liquidpele:
Open new issues.

RenatoUtsch:
@liquidpele Something strange is happening with Shadow DOM.

While the line, cumulative line and stacked area charts are working perfectly, all the other charts stopped being interactive. What did you do differently with the other charts that you didn't with these three?

liquidpele:
there isn't really any difference...  for pie and bars, the tooltip should follow the mouse instead of position next to the point, but that's about it.  If you can reproduce it in an online example so I can see, then open an issue and I'll take a look.

RenatoUtsch:
I can reproduce it but with a polymer wrapper. Is that a problem for you?

liquidpele:
Dunno, open an issue and let's see.

HedCET:
hi i am using nvd3 from bower & using nvd3 graphs inside polymer elements

```
    nv.tooltip.cleanup = function() {

// linto
$("* /deep/ .nvtooltip").remove();
// linto

        // Find the tooltips, mark them for removal by this class (so others cleanups won't find it)
        var tooltips = document.getElementsByClassName('nvtooltip');
        var purging = [];
        while(tooltips.length) {
            purging.push(tooltips[0]);
            tooltips[0].style.transitionDelay = '0 !important';
            tooltips[0].style.opacity = 0;
            tooltips[0].className = 'nvtooltip-pending-removal';
        }

        setTimeout(function() {

            while (purging.length) {
                var removeMe = purging.pop();
                removeMe.parentNode.removeChild(removeMe);
            }
        }, 500);
    };
```

here the problem is tooltip don't automatically hide with out this line of code ```$("* /deep/ .nvtooltip").remove();```
teste

liquidpele:
@HedCET this is a merged pull request, please open issues for bugs you find and make sure the bugs are against the latest development branch code.  Thanks.

RenatoUtsch:
@HedCET I already opened an issue:
https://github.com/novus/nvd3/issues/877

Anyways, I am currently developing a polymer wrapper to make it easy to use nvd3 from polymer. You can look at it here:
https://github.com/RenatoUtsch/polynvd3/tree/development

anthonyettinger:
How do I specify the `chartContainer` to use? By default the tooltips use window offset and don't take into account a div containing the chart with a scrollbar.

RenatoUtsch:
I actually fixed this in #1190, but they didn't accept the pull request and now it has merge conflicts. Don't know if I'll take the time to fix them.

anthonyettinger:
The docs say there is an `tooltip.chartContainer` option to use, but I have no idea how to use it.

RenatoUtsch:
You set this option when you create your chart. You normally access something like chart.interactiveLayer.tooltip.chartContainer() or chart.tooltip.chartContainer() to set it.

anthonyettinger:
I tried passing in chart options with this property:

```{ tooltip: {
chartContainer: '#my-chart'
}}```

RenatoUtsch:
This doesn't seem like nvd3. Are you using a wrapper like angular-nvd3?

I updated the pull request to solve the merge conflicts.

anthonyettinger:
Yeah we are using angular-nvd3

RenatoUtsch:
Which chart are you using? If it uses the Interactive Layer you have to access it first.

Issue #845, opened 2015-03-05
The scatterChart example no longer displays shapes other than circles


liquidpele:
The scatterChart example should be displaying multiple types of shapes, including a custom-built shape defined in the scatterChart.html example page, but it's only displaying circles for me when I build latest and view it.

PR #852, opened 2015-03-06
Fix scatter shapes

liquidpele:
Shapes show up again in scatter chart.


liquidpele:
Fixes https://github.com/novus/nvd3/issues/845

Issue #145, opened 2013-05-30
d3.time.scale compatibility


ninjudd:
It seems to me that `d3.time.scale` cannot be used with nvd3.

For example, this seems like it should work to me, but it doesn't:

```
chart.xAxis.scale(d3.time.scale());
```
It seems to ignore the tickFormat (add possibly the tickValues) completely.

The reason I want to use `d3.time.scale` is because it seems to have good handling of automatic time ticks (https://github.com/mbostock/d3/wiki/Time-Scales#wiki-ticks). Specifically, it looks like it automatically aligns tick marks to human friendly intervals:

```
1-, 5-, 15- and 30-second.
1-, 5-, 15- and 30-minute.
1-, 3-, 6- and 12-hour.
1- and 2-day.
1-week.
1- and 3-month.
1-year.
```

Is there any way to get nvd3 working with d3 time scales? Is there something I'm missing? I'm happy to help out if someone can point me in the right direction.

heliodor:
I posted a pretty complete time series example on JsFiddle: http://jsfiddle.net/3r88bgjw/ Hope it helps.

It shows how to: 1) set the x scale to a time scale so default ticks are round values, 2) format the tooltip, 3) put tick marks on the x axis for usability, 4) use bars instead of line so any gaps are visible, 5) use multi-format so that time labels are short and fit nicely.

AlonAmsalem:
amazing, thanks!

mb-dev:
Thanks!

liquidpele:
@robinfhu Short term we should add that jsfiddle code as an example since it shows off a lot... long term we should probably merge in the main changes to historicalBar.  Time series makes a lot of sense for that chart, and line chart too.

liquidpele:
The jsfiddle example was merged into the project so there is an example there to work from at least, and w'ell try to make using a time series easier using that as a starting point.  Closing this story, and if anyone wants to help out send us a pull request :)

milimetric:
The bar timeseries chart was not too useful in my case, but adapting code from the other responses I was able to make the lineWithFocusChart work well:

        chart.lines.xScale(d3.time.scale());
        chart.lines2.xScale(d3.time.scale());

        var timeFormat = d3.time.format('%Y-%m-%d');
        chart.xAxis.tickFormat(timeFormat);
        chart.x2Axis.tickFormat(timeFormat);

[...]

PR #856, opened 2015-03-09
Add time series example 

liquidpele:
Took the jsfiddle example from https://github.com/novus/nvd3/issues/145#issuecomment-76990255 and made an example in the project with it.

### DIFF
--- a/examples/TimeSeries.html
+++ b/examples/TimeSeries.html
@@ -1,0 +1,134 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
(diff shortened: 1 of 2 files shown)


robinfhu:
Nice example. But the axis tick marks code is not part of the chart update cycle. So if you resize the page, the tick marks won't resize with the chart.

liquidpele:
Yea I know, this is just the first step, I want to incorporate the tick marks into the axis code but didn't have time right now.

robinfhu:
Okay, the this should serve as a good example to point people to.

PR #857, opened 2015-03-09
Add possibility to have different slice size on a pie

mgalloue:
You can have the possibility to customize each slice size, by define an inner and an outer arc radius for each slice (value between 0 and 1). 
You can still use the existing donut ratio if you want.

### DIFF
--- a/src/models/pie.js
+++ b/src/models/pie.js
@@ -49,9 +49,11 @@ nv.models.pie = function() {
                 , container = d3.select(this)
                 ;
             if (arcsRadius.length === 0) {
+                var outer = radius - radius / 5;
+                var inner = donutRatio * radius;
                 for (var i = 0; i < data[0].length; i++) {
-                    arcsRadiusOuter.push(radius - radius / 5);
-                    arcsRadiusInner.push(donutRatio * radius);
+                    arcsRadiusOuter.push(outer);
+                    arcsRadiusInner.push(inner);
                 }
             } else {
                 arcsRadiusOuter = arcsRadius.map(function (d) { return (d.outer - d.outer / 5) * radius; });
--- a/src/models/pie.js
+++ b/src/models/pie.js
@@ -1,15 +1,15 @@
-nv.models.pie = function() {
+nv.models.pie = function () {
     "use strict";
 
     //============================================================
     // Public Variables with Default Settings
     //------------------------------------------------------------
 
-    var margin = {top: 0, right: 0, bottom: 0, left: 0}
+    var margin = { top: 0, right: 0, bottom: 0, left: 0 }
         , width = 500
         , height = 500
-        , getX = function(d) { return d.x }
-        , getY = function(d) { return d.y }
+        , getX = function (d) { return d.x }
+        , getY = function (d) { return d.y }
         , id = Math.floor(Math.random() * 10000) //Create semi-unique ID in case user doesn't select one
         , color = nv.utils.defaultColor()
         , valueFormat = d3.format(',.2f')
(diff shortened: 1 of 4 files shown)


src/models/pie.js
```diff
@@ -321,57 +346,65 @@ nv.models.pie = function() {
...
+        } 
+        },
+        y: { get: function () { return getY; }, set: function (_) {
+            getY = d3.functor(_);
+        } 
+        },
+        color: { get: function () { return color; }, set: function (_) {
+            color = nv.utils.getColor(_);
+        } 
+        },
```
liquidpele: Can you do }} so so the syntax matches the rest of the project?

src/models/pie.js
```diff
@@ -139,26 +162,26 @@ nv.models.pie = function() {
...
-                        .attr("d", arc);
+                        .attr("d", arcs[i]);
                 }
-                dispatch.elementMouseout({data: d.data, index: i});
+                dispatch.elementMouseout({ data: d.data, index: i });
             });
-            ae.on('mousemove', function(d, i) {
-                dispatch.elementMousemove({data: d.data, index: i});
+            ae.on('mousemove', function (d, i) {
+                dispatch.elementMousemove({ data: d.data, index: i });
```
liquidpele: There are a ton of these spacing changes... they really distract from your actual code changes, would it be possible to submit the PR without all the spacing changes?

liquidpele:
I checked out the example from your fork, REALLY COOL!

liquidpele:
You'll probably want to link to your example page in /examples/site.html so that you can get to it from the main example page.

robinfhu:
Oh that's an interesting feature.
screen shot 2015-03-09 at 11 59 58 am

src/models/pie.js
```diff
@@ -41,11 +42,22 @@ nv.models.pie = function() {
...
-                ,container = d3.select(this)
+                , availableHeight = height - margin.top - margin.bottom
+                , radius = Math.min(availableWidth, availableHeight) / 2
+                , arcsRadiusOuter = []
+                , arcsRadiusInner = []
+                , container = d3.select(this)
                 ;
+            if (arcsRadius.length === 0) {
+                for (var i = 0; i < data[0].length; i++) {
+                    arcsRadiusOuter.push(radius - radius / 5);
```
robinfhu: because you are pushing constant values in to the arrays, pre-calculate 'radius - radius/5', store it in a variable, and just push the variable

src/models/pie.js
```diff
@@ -41,11 +42,22 @@ nv.models.pie = function() {
...
+                    arcsRadiusInner.push(donutRatio * radius);
```
robinfhu: pre calculate donutRatio*radius

src/models/pie.js
```diff
@@ -321,6 +344,7 @@ nv.models.pie = function() {
 
     chart._options = Object.create({}, {
         // simple options, just get/set the necessary values
+        arcsRadius: { get: function () { return arcsRadius; }, set: function (_) { arcsRadius = _; } },
```
robinfhu: add this new option feature to documentation.html

liquidpele:
@robinfhu @mgalloue do you think instead of a whole separate array/option, that perhaps it would be easier to just add attributes to the main data object like this?

    var testdata1 = [
        { key: "Updated", y: 0, innerRadius: 0.5, outerRadius: 1 },
        { key: "Pending", y: 100, innerRadius: 0.65, outerRadius: 0.95 }
    ];

That way the radius data is also sent to the tooltip and available for other stuff more easily...  just a thought, I'm not saying it *has* to change...

mgalloue:
@liquidpele I like the fact it's a separate option because I see it like an advanced option, but if you want to use it easily for other stuff, it also makes sense...

src/models/pie.js
mgalloue: sure, thanks

liquidpele:
okay that's fine it is your future so I will leave the decision up to you

liquidpele:
@robinfhu I'm cool with merging whenever you are.

PR #875, opened 2015-03-12
Fix legendPosition('right') for pie/donut charts. 

liquidpele:
This is a pull request from https://github.com/nvd3-community/nvd3/pull/71 that needs to go into novus too.


liquidpele:
Hmm, I wonder why it can't automatically merge it?   That's weird...

PR #888, opened 2015-03-16
Basic support in touch browser

zocoi:
- Enable interactiveLayer mouseHandler on touch event
- Fixed position of tooltip in touch browser, because webkit does not support `transform` which makes tooltips to appear on the top left corner

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -249,8 +249,9 @@
                         // using tween since some versions of d3 can't auto-tween a translate on a div
                         .styleTween('transform', function (d) {
                             return translateInterpolator;
-                        })
-                        // not using tween for webkit touch device
+                        }, 'important')
+                        // Safari has its own `-webkit-transform` and does not support `transform` 
+                        // transform tooltip without transition only in Safari
                         .style('-webkit-transform', new_translate)
                         .style('opacity', 1);
                 }
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -250,6 +250,8 @@
                         .styleTween('transform', function (d) {
                             return translateInterpolator;
                         })
+                        // not using tween for webkit touch device
+                        .style('-webkit-transform', new_translate)
                         .style('opacity', 1);
                 }
 
--- a/src/interactiveLayer.js
+++ b/src/interactiveLayer.js
@@ -141,6 +141,7 @@ nv.interactiveGuideline = function() {
             }
 
             svgContainer
+                .on("touchmove",mouseHandler)
                 .on("mousemove",mouseHandler, true)
                 .on("mouseout" ,mouseHandler,true)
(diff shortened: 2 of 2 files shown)


zocoi:
ping @liquidpele @robinfhu for reviewing

src/tooltip.js
```diff
@@ -250,6 +250,8 @@
                         .styleTween('transform', function (d) {
                             return translateInterpolator;
                         })
+                        // not using tween for webkit touch device
+                        .style('-webkit-transform', new_translate)
```
liquidpele: Are you sure this doesn't remove the tooltip move animation for non-mobile webkit browsers?

liquidpele:
Yea, that looks like it's going to break chrome on the desktop... the transform and -webkit-transform will fight:      https://jsfiddle.net/w2zd8bdf/

zocoi:
@liquidpele I made an example of the transition here and looks like `-webkit-transform` was preceding and got overriden by `transform`: http://jsfiddle.net/w2zd8bdf/14/ , it could be because the smooth transition keep updating the style which made it cascading.

This is still a hack until we introduce browser detection to do handle events and optimize animation properly. What do you think?

liquidpele:
Yea it will work fine as long as the transform comes after the -webkit-transform in the style attribute, but I'm not sure we can rely on that... 

Besides that, the tooltips don't seem to have any issues in chrome 41 on my android 4.4.4, can you specify the exact platform/version where the transform: translate(..) isn't working?

zocoi:
it's Safari on any platform: desktop, iOS (iphone, ipad)...

liquidpele:
Ugh... Confirmed.  Safari is becoming the new IE  -_-

Okay, I played with this some and it does appear to add the tween to the end of the style block, but I still think it's a bit sketchy since that's d3 behavior and I'm not sure they absolutely define it that way (i.e. it might just be how they did it, but that could change at any time).    

The styleTween() function can take a third parameter, the string 'important' to add the !important flag to the CSS item in the style attribute, that would make sure that the transform took priority over the -webkit-transform for chrome/opera and safari should fall back to the -webkit-transform    ... the downside to that being that then it would be slightly harder to override tooltip position manually if someone wanted to do that.   @robinfhu what do you think?

liquidpele:
@robinfhu what's your opinion about maybe setting the styleTween to do !important there?

robinfhu:
Just starting to read this thread, give me a sec.

robinfhu:
I think it would be rare that someone would want to override the tooltip position.  So the 'important' flag doesn't bother me.

Also, the inline comment should explain in detail what the 'webkit-transform' is for, otherwise we might accidentally remove it in the future.

zocoi:
updated with comment and important flag

liquidpele:
Cool thanks for contributing!

PR #896, opened 2015-03-17
support tooltip offsets from pointer

admorphit:
add support for configuring tooltip offset

### DIFF
--- a/examples/documentation.html
+++ b/examples/documentation.html
@@ -793,12 +793,6 @@ as well as used to print out examples using those example inputs.
             examples: [{top: 200, left: 300}, {left: 50}],
             example_object: 'chart.tooltip'
         },
-        tooltipOffset: {
-            desc: "For tooltip: sets the top/left offset for the tooltip from the pointer.  Should be given an object with 'left' and/or 'top' attributes.  You can override just one, just like the 'margin' option on charts.",
-            default: "Starts off with {top: 0, left: 0}",
-            examples: [{top: 200, left: 300}, {left: 50}],
-            example_object: 'chart.tooltip'
-        },
         contentGenerator: {
             desc: "For tooltip: Function that generates the tooltip content html.  This replaces the 'tooltipContent' option that was on most charts.  Please note that the data passed this function is usually different depending on the chart, so you'll probably need to console.log() the input object.  Also, the data passed is always a single object now, so previous functions written for the tooltipContent option will have to be adjusted accordingly.",
             default: "See contentGenerator function in tooltip.js",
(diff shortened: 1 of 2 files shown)


zocoi:
+1

liquidpele:
Looks good but please add the new option to examples/documentation.HTML in the json definitions.

admorphit:
Added
```
offset: {
            desc: "For tooltip: sets the top/left offset for the tooltip from the pointer.  Should be given an object with 'left' and/or 'top' attributes.  You can override just one, just like the 'margin' option on charts.",
            default: "Starts off with {top: 0, left: 0}",
            examples: [{top: 200, left: 300}, {left: 50}],
            example_object: 'chart.tooltip'
        }
```
to `examples/documentation.html`, but it's appearing under `StackedAreaChart` because it's the first one in the link to have an `offset` property. Is there an easy way to get the offset to display under the tooltip section?

liquidpele:
oh, if that same option name is used elswhere then they'll kind of conflict in the docs... maybe rename it to tooltipOffset ? 

As long as the name in the json matches the option name in chart._options then it should automatically show up in the docs for that chart object.

admorphit:
Hate the idea of having to do something like `this.chart.tooltip.tooltipOffset` just because the documentation can't handle it.

Implemented the changes though would recommend updating the docs to be more flexible.

liquidpele:
haha... tell you what, if you want to just call it offset and leave it out of the docs for now I'll make the docs more flexible.  I had been meaning to do it anyway :)

admorphit:
Would be awesome to have updated docs!

Noticed there are a few other instances like `stackOffset`, `titleOffset` with the same issue- do you think it will introduce confusion if we just did offset for tooltip? Sounds like a balance between breaking changes down the track and being consistent at this time.

It's a bit messy either way so am not sure what's best- your call.

liquidpele:
After thinking about it, I think we should make the option name clear and fix the docs to handle it.

admorphit:
By clear, do you mean to make `offset` property for StackedAreaChart more explicit and keep `xxxOffset` for all others? Or do you mean change every instance to `offset` and update the docs?

liquidpele:
na, don't change other stuff or it coudl break existing installs for people.   I'm saying you can just change the option on tooltip to just be "offset" again... you can just revert the commit to that if you want, and don't worry about putting it in documentation, and I'll fix the docs to work more gooder.

admorphit:
sweet- done

liquidpele:
awesome thank you for contributing and sorry for the back and forth!

PR #897, opened 2015-03-18
Added label types to pie charts

joneschrisan:
Added label types key + value, key + percent, and key + value + percent to pie charts.

### DIFF
--- a/src/models/pie.js
+++ b/src/models/pie.js
@@ -287,7 +287,10 @@ nv.models.pie = function() {
                         var labelTypes = {
                             "key" : getX(d.data),
                             "value": getY(d.data),
-                            "percent": labelFormat(percent)
+                            "percent": labelFormat(percent),
+                            "key:value" : getX(d.data) + ' (' + getY(d.data) + ')',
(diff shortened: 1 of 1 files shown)


liquidpele:
This isn't very extendable, it would be better to add a "labelFormat" option that allows you to override the whole thing yourself with a custom function... that way people can set the label format to be anything they want...

joneschrisan:
True but I only added the 3 labels to the existing 3 labels like I have done for use on my own website. Just working with what was already there.

liquidpele:
Well, as-is I don't think this is a pull request I'd pull in since it would add extra code there that would have to be kept around for backward compatibility even if the better option was added later.   If you want to re-submit the PR with the more generalized option I'd gladly accept it :)

If you don't have time or just don't care, that's fine too, we do appreciate all pull requests since at a minimum it lets us see how people are using nvd3 and where features or options might be lacking.

robinfhu:
Also you are not comparing it against the development branch.  I changed some logic around this part of the code.

vineelal:
Are the label types key + value, key + percent, and key + value + percent are available in the current version ?

PR #898, opened 2015-03-18
Added my labels to development branch

joneschrisan:
Added my labels to development branch


liquidpele:
previous PR:   https://github.com/novus/nvd3/pull/897

I still don't think this is a PR we should take as-is since a more general option would be a much better scenario, what do you think @robinfhu ?

joneschrisan:
Closing for adding custom label

PR #900, opened 2015-03-18
Custom Pie Chart Labels

joneschrisan:
Added working custom pie chart labels

### DIFF
--- a/src/models/pie.js
+++ b/src/models/pie.js
@@ -317,6 +317,9 @@ nv.models.pie = function() {
                             case 'key:value:percent':
                                 label = getX(d.data) + ' (' + getY(d.data) + ')' + ' (' + labelFormat(percent) + ')'
                                 break;
+                            default:
+                                label = labelType();
+                                break;
                         }
                         return label;
                     })
--- a/src/models/pie.js
+++ b/src/models/pie.js
@@ -298,19 +298,29 @@ nv.models.pie = function() {
                         var label = '';
                         if (!d.value || percent < labelThreshold) return '';
 
-                        switch (labelType) {
-                            case 'key':
-                                label = getX(d.data);
-                                break;
-                            case 'value':
-                                label = valueFormat(getY(d.data));
-                                break;
-                            case 'percent':
-                                label = valueFormat(percent);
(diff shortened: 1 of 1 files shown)


src/models/pie.js
```diff
@@ -298,16 +298,20 @@ nv.models.pie = function() {
...
-                                label = getX(d.data);
-                                break;
-                            case 'value':
-                                label = valueFormat(getY(d.data));
-                                break;
-                            case 'percent':
-                                label = valueFormat(percent);
-                                break;
+                        if(typeof labelType === 'function') {
+                            label = labelType({ 'key': getX(d.data), 'value': getY(d.data), 'percent': labelFormat(percent) });
```
liquidpele: Just pass d to it and let the function handle everything in this case.

liquidpele:
This looks good besides that one thing!

joneschrisan:
But keeping the other 3 hard coded labels in there would mean backward compatibility. Otherwise everyone will have to change their labels to anonymous functions.

joneschrisan:
I tried passing d to it but it couldn't run the other functions getX, getY, and labelFormat as they had gone out of scope.

liquidpele:
Shouldn't labelFormat be valueFormat there?   Also you can get to it from the chart options... e.g:

    chart.labelType(function(d) {
        var key = chart.x()(d.data);
        var value = chart.y()(d.data);
        var percent = chart.valueFormat()(percent);
        return key + ':' + value + ':' + percent;
    });

liquidpele:
Or if you want, you can pass the object  you're passing right now, but just add an attribute 'data' that is d... mostly I just want to make sure all available data is passed *somewhere* there... :)

joneschrisan:
It should yes, I just noticed that too. I'd agree with you, just didn't know a way and needed to get it working quick. Passing d and then using that would be better, more usable.

liquidpele:
ping.  can you put d into the object you pass with the other params, just so the raw data object is available to the function?

liquidpele:
@joneschrisan you still around?

robinfhu:
I'll take care of this

PR #901, opened 2015-03-18
ParallelCoordinate support missing values.

mgalloue:
- Support missing values : Missing values cannot be displayed per definition, since they are missing. Nevertheless, they also provide useful information when trying to understand the data. The option proposed is to exploit the available horizontal space to introduce a third dimension outside of the normal range. Whenever a displayed row has a missing value in one dimension, the connecting line falls out of the regular space and falls onto the missing value axis.

Ref:  http://informationandvisualization.de/blog/parallel-coordinates-and-missing-values

clipboard01

- Support only one line displayed (Fix)
- Possibility to define a tickFormat for each axis

### DIFF
--- a/examples/documentation.html
+++ b/examples/documentation.html
@@ -765,6 +765,11 @@ as well as used to print out examples using those example inputs.
             default: "[]",
             examples: [["axis1", "axis2", "axis3"]]
         },
+		dimensionFormats: {
+            desc: "D3 format for each x axis",
+            default: "Use the default formatter scale.tickFormat",
+            examples: [["%", "", "0.2f"]]
+        },
         gravity: {
             desc: "Can be 'n','s','e','w'. Determines how tooltip is positioned.",
             default: "w",
--- a/examples/documentation.html
+++ b/examples/documentation.html
@@ -756,6 +756,15 @@ as well as used to print out examples using those example inputs.
             desc: "Specifies each line tension. Values between 0 and 1.",
             default: 1,
             examples: [0.85]
+        },
+        dimensions: {
+            desc: "Deprecated. Use dimensionsNames instead. "
+        },
+        dimensionNames: {
+            desc: "Name of each dimension, used for each axis.",
+            default: "[]",
+            examples: [["axis1", "axis2", "axis3"]]
+        },
         gravity: {
             desc: "Can be 'n','s','e','w'. Determines how tooltip is positioned.",
             default: "w",
--- a/examples/documentation.html
+++ b/examples/documentation.html
@@ -765,7 +765,7 @@ as well as used to print out examples using those example inputs.
             default: "[]",
             examples: [["axis1", "axis2", "axis3"]]
         },
-		dimensionFormats: {
+        dimensionFormats: {
             desc: "D3 format for each x axis",
             default: "Use the default formatter scale.tickFormat",
             examples: [["%", "", "0.2f"]]
--- a/examples/parallelCoordinates.html
+++ b/examples/parallelCoordinates.html
@@ -35,7 +35,7 @@ nv.addGraph(function() {
 
     chart = nv.models.parallelCoordinates()
         .dimensionNames(["economy (mpg)", "cylinders", "displacement (cc)", "power (hp)", "weight (lb)", "0-60 mph (s)", "year"])
-		.lineTension(0.85);
+        .lineTension(0.85);
 		
 
     d3.select('#chart1 svg')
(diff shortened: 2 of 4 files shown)


examples/parallelCoordinates.html
```diff
@@ -12,6 +12,9 @@
         }
         svg {
             display: block;
+            float: left;
+            height: 450px !important;
+            width: 850px !important;
```
liquidpele: I'd like to keep the example charts so that they expand with the window when you view just the example so we can test the responsiveness of the charts that way.  Is there a specific reason it has to be this size?

mgalloue: Nop, it was just to have one next to the other, I can remove it.

liquidpele: ah, I see... is there a reason you need two graphs or can you fit all the features you want to demo into one?  If you need 2 that's fine.  I really need to make a way to page through multiple examples anyway...

src/models/parallelCoordinates.js
```diff
@@ -38,11 +38,24 @@ nv.models.parallelCoordinates = function() {
             // Setup Scales
             x.rangePoints([0, availableWidth], 1).domain(dimensionNames);
 
+            var onlyNanValues = {};
             // Extract the list of dimensions and create a scale for each.
             dimensionNames.forEach(function(d) {
+                var extent = d3.extent(data, function(p) { return +p[d]; });
```
liquidpele: I know it was already calling the variable d, but can we change it to i or index or something to correctly identify it when glancing at the code? d is typically the series data object when used elsewhere...

liquidpele:
Also, and it's not *required*, but if you can add some more comments about what you added and why it would be nice :)

src/models/parallelCoordinates.js
mgalloue: Yes but here d is not really an index, it's more like a key because it's a string corresponding to the name, it's part of data. Maybe I should rename it k ?

liquidpele: oh so d is the value and not the index there?   different than jquery I guess :p   Okay, nevermind, calling it d is fine :)

liquidpele:
Looks pretty cool to me, I'll wait to see what @robinfhu and others say too :)

examples/documentation.html
```diff
@@ -765,6 +765,11 @@ <h3 class="option"><span class="expand all">[ <a href="javascript:$('#optionlist
             default: "[]",
             examples: [["axis1", "axis2", "axis3"]]
         },
+		dimensionFormats: {
```
robinfhu: fix the indentation here

src/models/parallelCoordinates.js
```diff
@@ -38,11 +39,26 @@ nv.models.parallelCoordinates = function() {
...
+                //Scale axis if there is only one value
+                if (extent[0] === extent[1]) {
+                    extent[0] = extent[0] - 1;
+                    extent[1] = extent[1] + 1;
+                }
                 y[d] = d3.scale.linear()
-                    .domain(d3.extent(data, function(p) { return +p[d]; }))
-                    .range([availableHeight, 0]);
+                    .domain(extent)
+                    .range([(availableHeight - 12) * 0.9, 0]);
```
robinfhu: document the numbers 12 and 0.9 here

src/models/parallelCoordinates.js
```diff
@@ -67,6 +84,28 @@ nv.models.parallelCoordinates = function() {
...
+                    .attr("x2", function(d) { return d[2]; })
+                    .attr("y2", function(d) { return d[3]; });
+
+            //Add the text "undefined values" under the missing value line 
+            missingValueslineText = wrap.select('.missingValuesline').selectAll('text').data(["undefined values"]);
+            missingValueslineText.append('text').data(["undefined values"]);
+            missingValueslineText.enter().append('text');
+            missingValueslineText.exit().remove();
+            missingValueslineText.attr("y", availableHeight)
+                    .attr("x", availableWidth - 92 - step / 2)
```
robinfhu: Too many magic numbers. What does 92 represent?

src/models/parallelCoordinates.js
mgalloue: 12 is the height of the text "undefined values", 0.9 represents 90% of the availableHeight minus the text size.

src/models/parallelCoordinates.js
mgalloue: 92 is the width of the text "undefined values", this is necessary to have the text and the missingValues line left align.

robinfhu:
Thanks!

Issue #893, opened 2015-03-17
More problems with Shadow DOM and tooltips


RenatoUtsch:
@liquidpele 

So, today I went and started to test the tooltips again, but a change yesterday must have broken something, because now if I create multiple charts only the tooltip of the first chart gets positioned right. All the other tooltips appear at wrong positions or at the bottom of the screen.

Also, I noticed another problem:
The tooltips stopped showing the information of some charts that are built correctly and used to have correct values on the tooltips. Now only NaN appears.

For example, in my repo in nvd3-scatter-chart/demo.json, the 2nd chart that does NOT set the size of the circles shows only NAN in the tooltips.
The first pie chart example also shows NAN but I don't know why, but it used to work correctly.
The 2nd discrete bar chart example also.

Maybe the tooltips are getting the wrong data?

RenatoUtsch:
I'll revert some nvd3 commits and see why the charts are being positioned incorrectly.

liquidpele:
whoops

liquidpele:
Okay, I won't be able to look at it for a while, so if you can track down the specifics in the meantime that woudl be helpful.

RenatoUtsch:
Actually, I reverted a lot of nvd3 commits and my own too in my repo to test, but the problem is always happening. So I assume it was a bug I didn't notice.

Later today I'll take a deeper look at this.

RenatoUtsch:
The Scatter Chart is also a bit strange, the tooltips are not moving correctly and they disappear if I stop on one of the points.

liquidpele:
this issue is closed now you will need to open something new and make sure to include specific reproduction steps or an online demo that I can look at

liquidpele:
oh wait this is the new one but I still need something more than a vague description

RenatoUtsch:
About the tooltip position problem, the width is being set correctly, but the height is not being changed.

I'll update the online demo for you.

RenatoUtsch:
- Scatter chart disappearance problem on the first chart, incorrect positioning and NaN values in the second chart.
http://renatoutsch.com.br/bower_components/polynvd3/nvd3-scatter-chart/demo.html

- Second pie chart with positioning problems and first chart with NaN values:
http://renatoutsch.com.br/bower_components/polynvd3/nvd3-pie-chart/demo.html

- Other charts:
http://renatoutsch.com.br/bower_components/polynvd3

A note worth mentioning is that the interactive guideline charts are working correctly, even with multiple charts.

[...]

PR #907, opened 2015-03-20
Make tooltips get the right values if people change the getX and getY getter functions.

liquidpele:
:p


liquidpele:
Fixes part of https://github.com/novus/nvd3/issues/893 where the tooltip showed NaN

robinfhu:
Oh right this is a big deal!

liquidpele:
unit tests are passing, the last one failed just because travis CI bombed out completely ;)

robinfhu:
Is this PR a work in progress, or is this everything that needs to be changed?

liquidpele:
I think it is everything but please look and see if I overlooked anything

robinfhu:
The important charts were addressed, it's fine. Thanks!

RenatoUtsch:
Was any chart not addressed?

liquidpele:
I intended to get em all so if I missed one let me know!

RenatoUtsch:
For example, I don't see the files of the bullet and cumulativeLine charts changed... neither ohlc...

liquidpele:
Some charts don't use x/y getters or inherit from other objects that do.  If you can break one in a jsfiddle let me know.

RenatoUtsch:
Okay.

Thanks for the good work!

RenatoUtsch:
I am still noticing problems with the nvd3-discrete-bar-chart when using timestamps as the x values.

Issue #895, opened 2015-03-17
Axis not rescaling with resize (issue with axis.ticks())


RenatoUtsch:
I noticed that something that used to happen with the old nvd3 (the version on the website) that was very important for responsive designs stopped happening somewhere between the creation of nvd3-community and here.

If you open an example at nvd3.org and look at it with no zoom and then zoom it in, you'll notice that the axis adds and removes ticks according to the chart size:

- Before Zoom
![screen shot 2015-03-17 at 5 13 54 pm](https://cloud.githubusercontent.com/assets/836333/6696558/02f35628-ccc9-11e4-9705-b37ffb9d54a7.png)

- After Zoom
![screen shot 2015-03-17 at 5 13 04 pm](https://cloud.githubusercontent.com/assets/836333/6696570/178b46ae-ccc9-11e4-8996-c8364ab36729.png)

In the new nvd3, if I try to do the same thing with the examples, the ticks are not removed and get cramped together in the axis. This is bad.

- Before Zoom
![screen shot 2015-03-17 at 5 20 00 pm](https://cloud.githubusercontent.com/assets/836333/6696680/d4e27e34-ccc9-11e4-9b30-19bed83e1a9f.png)

- After Zoom
![screen shot 2015-03-17 at 5 20 11 pm](https://cloud.githubusercontent.com/assets/836333/6696684/dae23414-ccc9-11e4-8c4e-dfcaa2351557.png)

This is something that needs to be fixed, because it is one of the most important aspects of responsivity.

robinfhu:
Yeah, I see what you mean. Looking at it

robinfhu:
It was one of my changes: 6f57104

robinfhu:
I had tried to fix https://github.com/novus/nvd3/pull/601, but it looks like I broke the original functionality, which had the chart automatically figure out the ticks.

robinfhu:
There needs to be a way we can reconcile the ability to let users set the number of ticks they want, and allow the charts to automatically figure out the number of ticks.

robinfhu:
One idea I have is to add a new property to nv.model.axis, called "autoTicks". If autoTicks is set to true, the axis will ignore the ticks() property and just try to automatically figure out the number of ticks to show, just like before.

liquidpele:
I'm confused why your commit broke that...

RenatoUtsch:
That may be a good idea, but why not implement the original functionality in axis.ticks() and allow it to be overriden?

[...]

PR #915, opened 2015-03-26
Possible fix for axis not rescaling. 

liquidpele:
https://github.com/novus/nvd3/issues/895

THIS IS NOT COMPLETE.   I'm making this pull request for discussion as a possible solution.  

What this does is it actually makes each chart create a duplicate of each option named the same with an underscore prefix (which is thus hidden in the docs).  Calling the underscore version will be IGNORED if the non-underscore version has been called at some point.  This allows nvd3 to change it's own options internally up until someone changes the option manually from code outside of nvd3.  

Thoughts?

### DIFF
--- a/src/models/cumulativeLineChart.js
+++ b/src/models/cumulativeLineChart.js
@@ -362,7 +362,7 @@ nv.models.cumulativeLineChart = function() {
             if (showXAxis) {
                 xAxis
                     .scale(x)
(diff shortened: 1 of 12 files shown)


robinfhu:
Different from my idea, but I like it.  I think this can extend to a lot of options.

liquidpele:
Okay, all the .ticks calls are now ._ticks() so that they can be overridden and the override will keep.

robinfhu:
Thanks. Want to test it out real quick.

robinfhu:
Thanks!!!

PR #921, opened 2015-03-27
nv.graphs is never emptied -- mem leak w/streaming data

tcotav:
I encountered the memory leak when streaming in data for a continuously updating dashboard.  It was only glaring for me because I had eight graphs of time series data on a single page and got to the dead browser screen pretty quick.  Commenting out seems right immediate fix as the array isn't used.  Longer term if nv.graphs is desired, you'd have to make sure it stayed a fixed size -- splicing off as you go.  I can provide test case/code if that'd help.  Thanks!


liquidpele:
Pull requests must be to the development branch not master...

tcotav:
will correct and resubmit shortly

PR #922, opened 2015-03-27
nv.graphs is never emptied -- mem leak w/refreshed data on page

tcotav:
I encountered the memory leak when ajax refreshing in data for a continuously updating dashboard. It was only glaring for me because I had eight graphs of time series data on a single page and got to the dead browser screen pretty quick. Commenting out seems the right immediate fix as the array isn't used. Longer term if nv.graphs is desired, you'd have to make sure it stayed a fixed size -- splicing off as you go.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -8,7 +8,6 @@ nv.tooltip = nv.tooltip || {}; // For the tooltip system
 nv.utils = nv.utils || {}; // Utility subsystem
 nv.models = nv.models || {}; //stores all the possible models/components
 nv.charts = {}; //stores all the ready to use charts
-//nv.graphs = []; //stores all the graphs currently on the page -- NYI fully 
 nv.logs = {}; //stores some statistics and potential error messages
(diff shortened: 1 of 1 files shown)


liquidpele:
This seems fine to me, there's no point storing an array like that imho.  However if we do this it should just delete those lines, not leave the commented out.    @robinfhu care to comment?

tcotav:
For my part, I left it in as I didn't want to presume to know the full intent of having a list of graphs.  I figured there may be some future use for an array of the graphs.

liquidpele:
@tcotav if you go ahead and just delete those lines I'll accept the pull request.  No sense in having them sitting them commented out if they cause problems :)

tcotav:
@liquidpele thanks for the patience on submission.  Hopefully that'll do the trick.  It certain allows my dashboard to run all day after the fix.

liquidpele:
Cool thanks for contributing!

PR #919, opened 2015-03-27
Remove bower install from postinstall hook

statianzo:
Having `bower install` inside the postinstall hook prevents this package from being installed via npm when bower isn't available

PR #923, opened 2015-03-29
Remove bower install from postinstall hook

statianzo:
Followup of #919.
Having `bower install` in the postinstall hook makes this library not possible to install through npm in environments where bower is not available.
To keep compatibility with travis, `bower install` is now invoked after npm install in the travis.yml


robinfhu:
Okay this is fine.

liquidpele:
This went into master, it should also be merged into development...

robinfhu:
Oh okay. Will do

Issue #929, opened 2015-04-01
ScatterChart lost its options method


topicus:
I've tried to use the options method in the scatterChart but apparently it's not available. I'm getting 'Undefined is not a function' when I try to use it.

```javascript
chart.options({showDistX: true});
```

liquidpele:
Looks to be there in the dev branch, closing:    

https://github.com/novus/nvd3/blob/development/src/models/scatterChart.js#L347

PR #930, opened 2015-04-01
novus/nvd3#929 Add options method to the scatterChart model

topicus:
novus/nvd3#929

### Acceptance test
- [ ] Create a new scatterChart
- [ ] Try to configure using options method
- [ ] No error should be threw


liquidpele:
pull requests have to be to the development branch, not master... also this appears to already be fixed in the development branch: 

https://github.com/novus/nvd3/blob/development/src/models/scatterChart.js#L347

PR #934, opened 2015-04-04
Add import.html to support html imports

RenatoUtsch:
This new file will add support to importing nvd3 through html imports.

This will solve compatibility problems between multiple web components using nvd3, as web components import other projects through html imports, and not directly with the <script> tag. This takes advantage of deduping, effectively solving the problem of loading the library multiple times.

### DIFF
--- a/import.html
+++ b/import.html
@@ -1,0 +1,1 @@
+<script src="build/nv.d3.min.js" charset="utf-8"></script>


liquidpele:
I don't understand why this would be a part of nvd3... this seems like something a project using nvd3 should do, especially considering that they might have a different folder structure...

RenatoUtsch:
This should be a part of nvd3 because all projects that use nvd3 have to import it through the same file for import deduping to work. The only other solution is to create a nvd3-import bower package, but I think that this would be a better solution.

I wrote something about it here:
https://github.com/Polymer/polymer/issues/1336

TL;DR; if you want to import a dependency that is shared between multiple projects, all these projects have to import this dependency by html-importing the same file, or the dependency will be imported multiple times (what won't happen if all import the dependency by the same file).

liquidpele:
What I mean is, this seems like what is essentially a middleware or stub, it's something very specific to a particular kind of integration... but as such it really should be part of whatever project is integrating nvd3, not part of nvd3 itself, because it assumes you want to integrate in a particular way which may or may not actually be the case.  Just my $0.02 anyway... @robinfhu  @portante what do ya'll think?

RenatoUtsch:
I added this import.html file to my project.

Issue #945, opened 2015-04-08
Stacked Area "Expanded" view does not handle a point where all y values are 0


lucaslin:
See this plunker for the problem behavior:
http://plnkr.co/edit/cafRVbvPyHbUdKhrD7FS?p=preview

I believe this behavior happens because of a divide-by-zero: 0/0

**What is the desired behavior?**

I can imagine the desired behavior could be showing all the series go to 0%, like in this plunker:
http://plnkr.co/edit/XCfyP5yZSWZzX1LZygiO?p=preview

...or showing each series being at 100/n %.

liquidpele:
In your "fixed" plunkr I get "Uncaught TypeError: undefined is not a function"

lucaslin:
Sorry. Looks like Plunker can't handle the size of the nvd3.js file.
I see this error in the console too, though it still works in Chrome for me despite that.

I just made a website to host it: http://nvd3stackedarea.parseapp.com/

liquidpele:
Ah okay, yep that looks like a bug then.   It also looks like you did something to fix it... if you can send a pull request to the development branch (not master) then we'll take a look and merge it :)

liquidpele:
Fixed by his pull request.

PR #950, opened 2015-04-09
Fixing Bug #945: Stacked Area "Expanded" view does not handle a point where all y values are 0

lucaslin:
Fixing Bug #945: Stacked Area "Expanded" view does not handle a point where all y values are 0

Making stackedArea (in Expanded view) display 0 (no area) at points where the total y value across all series is 0.

Note: style 'expand' and 'stack_percent' now are the same. One should probably be removed. (They seemed to be intended to produce the same thing)

Example: http://nvd3stackedarea.parseapp.com/


liquidpele:
Looks like you make 'expand' and 'stack_percent' the same thing here...  I guess my question is, are they really the same thing?   It seems like they would be, but I'm confused why there were two options there to begin with...   @robinfhu what's your take?

lucaslin:
Yeah, I'm also confused about why there were two options. 'stack_percent' is never called in the current production code. However, when I tried swapping it for 'expand', the result was exactly the same. 

'expand' uses d3's [chart.offset('expand')](https://github.com/mbostock/d3/wiki/Stack-Layout#offset), but clearly produced an undesirable result, so I made changes to 'stack_percent'.
I think we should just use what is in my changes to 'stack_percent' (and maybe rename it to 'expand'?).

liquidpele:
Well, I'm up for merging this in unless @robinfhu has any objections...

liquidpele:
Went ahead and merged it in, thanks for contributing!

PR #951, opened 2015-04-09
Fix #862, legend position 'right' buggy

### LINKED ISSUE #862: The pie/donut option "legendPosition" with value 'right' is buggy
I tried to use this (it was a PR from a community member) and the legend gets cut off easily and when you disable/enable slices it will jump left and right.

rreinhardt9:
This seems to fix the bugginess in issue #862; it adjusts as the browser window in sized up and down and doesn't translate the legend way off to the right on first load.

I couldn't get a good example working in jsfiddle because of all the resizing, but i created [this gist](https://gist.github.com/rreinhardt9/6b055863fcf41c76a5ce) that can be used to try it out locally.

I wanted to open this PR for discussion, like I mentioned in the issue I'm pretty new to nv and d3.

Thanks!

### DIFF
--- a/src/models/pieChart.js
+++ b/src/models/pieChart.js
@@ -123,15 +123,13 @@ nv.models.pieChart = function() {
                     wrap.select('.nv-legendWrap')
                         .attr('transform', 'translate(0,' + (-margin.top) +')');
                 } else if (legendPosition === "right") {
-                    legend.height( availableHeight ).key(pie.x());
-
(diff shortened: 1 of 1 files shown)


liquidpele:
The logic seems fine, and this change only affects things if they use the legendPosition option so it's pretty low risk.  I'll go ahead and merge it in, thanks for the contribution :)

PR #957, opened 2015-04-15
Fix interactiveGuideline bugs when data is missing.

LeeMendelowitz:
This commit fixes a bug with the interactive guideline for line charts
when data is missing (x, y, or both set to null or undefined),
Missing data caused incorrect points to be highlighted, likely
due to undefined behavior with d3.bisect when nulls or undefined's
are encountered.

This fix creates a filtered array of values for each series by
removing points that are undefined in x or y, and a map from the index
of a point in the filtered array to the index of the original point.
These structures are used to correctly highlight points and populate
tooltip when data is missing.

### DIFF
--- a/examples/lineChartWithInteractiveAndMissingData.html
+++ b/examples/lineChartWithInteractiveAndMissingData.html
@@ -1,0 +1,124 @@
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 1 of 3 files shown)


LeeMendelowitz:
Committed prematurely, will resubmit PR with screenshots

PR #956, opened 2015-04-14
Fix typo (presumably) in nvtooltip h3 border-radius.

marcocova:
For some reason, the value of the border-top-left-radius property was different than
the value of the webkit and mozilla versions of the property.

PR #958, opened 2015-04-15
Fix the border-top-left-radius property of .nvtooltip h3.

marcocova:
For some reason, the value of the border-top-left-radius property was
different than the value of the webkit and mozilla versions of the
property (1px vs 5px).

This is just like the earlier #956, but now against the development 
branch (sorry for the noise!).


liquidpele:
Cool, thanks!

Issue #952, opened 2015-04-10
ohlc chart with tooltip problems.


RenatoUtsch:
The ohlc chart tooltip isn't fading after I scroll out of the chart area.

Screenshot of the problem:
![screen shot 2015-04-10 at 1 45 45 pm](https://cloud.githubusercontent.com/assets/836333/7092403/066819de-df88-11e4-8d00-2d1a0c0f7cdc.png)

RenatoUtsch:
Actually...

The problem happens on the examples page also:
http://nvd3-community.github.io/nvd3/examples/site.html

RenatoUtsch:
The Time Series chart has the same problem.

RenatoUtsch:
@liquidpele, this is related to your new tooltip code, isn't it?

liquidpele:
I'll take a look when I get some time...

liquidpele:
should be fixed now, closing.

RenatoUtsch:
Fixed it, thanks.

PR #960, opened 2015-04-16
fix historicalBarChart tooltip not hidding when moused out, #952

jstrength:
Just needed to add 'tooltipHide' to list of dispatch functions in the historicalBarChart.js file. #952


robinfhu:
Looks fine.

PR #962, opened 2015-04-16
Fix interactiveGuideline bugs when data is missing.

LeeMendelowitz:
This commit fixes a bug with the interactive guideline
when data is missing (x, y, or both set to null or undefined),
Missing data caused incorrect points to be highlighted, likely
due to undefined behavior with d3.bisect when nulls or undefined's
are encountered.

This fix creates a filtered array of values for each series by
removing points that are undefined in x or y, and a map from the index
of a point in the filtered array to the index of the original point.
These structures are used to correctly highlight points and populate
tooltip when data is missing.

### DIFF
--- a/examples/lineChartWithInteractiveAndMissingData.html
+++ b/examples/lineChartWithInteractiveAndMissingData.html
@@ -1,0 +1,124 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <link href="../build/nv.d3.css" rel="stylesheet" type="text/css">
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.2/d3.min.js" charset="utf-8"></script>
+    <script src="../build/nv.d3.js"></script>
+
+    <style>
+        text {
+            font: 12px sans-serif;
+        }
+        svg {
+            display: block;
+        }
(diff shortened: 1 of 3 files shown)


LeeMendelowitz:
This is my proposed fix to  https://github.com/novus/nvd3/issues/963

LeeMendelowitz:
Here are screenshots of the fixed behavior.
 - If the guideline is positioned at an X location where there is no data in a series, no point is highlighted from the series and it is not shown in tooltip
 - The guideline will only position itself at an X location where data is defined in at least one series.
 - The guideline will position itself at the closest such X location to the cursor.


fixed_1

fixed_3

robinfhu:
This solution is too complicated to solve this problem.  I actually know of a better way.

robinfhu:
The better solution is to simply not invoke lines.highlightPoint(i, pointIndex, true); when the pointValue is null.

LeeMendelowitz:
@robinfhu d3.bisect may still pick the x value with a null/undefined y

robinfhu:
Oh i see.  You also want it removed from the tooltip.  

I'm going to PR a solution where I show an "N/A" in the tooltip instead.

robinfhu:
@LeeMendelowitz  see https://github.com/novus/nvd3/pull/964

Issue #963, opened 2015-04-16
LineChart Interactive Guideline broken with missing data


LeeMendelowitz:
Adding missing data with null/undefined x or y values is a way to break up the line into discontinuous segments in a line chart, but the missing values mess up the interactive guidlelines. Sometimes the wrong point is highlighted, as shown in screenshots below.

![error_missing_data2_wrong_point](https://cloud.githubusercontent.com/assets/2424933/7182386/f3d49dba-e41e-11e4-9ddc-b68cfffe7687.png)

![error_missing_data1](https://cloud.githubusercontent.com/assets/2424933/7182390/f5bfdd6a-e41e-11e4-8e7c-d19c0cb223e2.png)

RenatoUtsch:
Should this issue be closed?

liquidpele:
yep

PR #964, opened 2015-04-16
Issue #963 Prevent a null point from being highlighted in lineChart

robinfhu:
In example, show N/A in the tooltip if value undefined

### DIFF
--- a/examples/lineChart.html
+++ b/examples/lineChart.html
@@ -49,7 +49,7 @@
         chart = nv.models.lineChart()
             .options({
                 transitionDuration: 300,
-                //useInteractiveGuideline: true
+                useInteractiveGuideline: true
             })
         ;
 
@@ -62,7 +62,12 @@
 
         chart.yAxis
             .axisLabel('Voltage (v)')
-            .tickFormat(d3.format(',.2f'))
+            .tickFormat(function(d) {
+                if (d == null) {
+                    return 'N/A';
+                }
+                return d3.format(',.2f')(d);
+            })
         ;
 
         data = sinAndCos();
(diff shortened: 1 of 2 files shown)


LeeMendelowitz:
This isn't working for my test case - I think this picks points with different X values from different series.

LeeMendelowitz:
@robinfhu it picks the closest defined X for each series, and highlights the point, and shows in the tooltip.  This isn't right because the tooltip also displays an X value which doesn't match the highlighted points.

LeeMendelowitz:
robin_fix

robinfhu:
@LeeMendelowitz , ok.  I am starting to understand why this is a problem for your use case.

First, you are creating data points where X=null, which you should not be doing.  This will break the chart in many ways.  Can you not preprocess your data to prevent nulls in your X values?

LeeMendelowitz:
@robinfhu Okay, I'll fix the test case so x is never null and check it out.

LeeMendelowitz:
964_not_null

@robinfhu here's with not null. Now the point isn't highlighted but still appears in tooltip. Issue is that `chart.y()` converts the null to a 0.0 before your fix checks for null.

robinfhu:
@LeeMendelowitz  is there a way you can augment my solution?

LeeMendelowitz:
@robinfhu Sure, I'll check it out and make a PR to your repo?

robinfhu:
@LeeMendelowitz  sure

LeeMendelowitz:
@robinfhu see https://github.com/novus/nvd3/pull/965

PR #965, opened 2015-04-16
Fix #Issue 963

LeeMendelowitz:
This fix for https://github.com/novus/nvd3/issues/963 adds to @robinfhu's https://github.com/novus/nvd3/pull/964 PR by displaying "N/A" in interactive guideline tooltip when data is missing.

Together with these commits:
 - Series that have a y value of null/undefined for the interactive guideline's x will not have a highlighted point and
 - will appear in interactive tooltip as "N/A"

### Screenshot 1
fix1

### Screenshot 2
fix2


src/models/lineChart.js
```diff
@@ -225,14 +225,17 @@ nv.models.lineChart = function() {
                     })
                     .forEach(function(series,i) {
                         pointIndex = nv.interactiveBisect(series.values, e.pointXValue, chart.x());
-                        lines.highlightPoint(i, pointIndex, true);
                         var point = series.values[pointIndex];
+                        var pointYValue = chart.y()(point, pointIndex);
+                        if (pointYValue != null) {
+                            lines.highlightPoint(i, pointIndex, true);
+                        }
```
liquidpele: This same logic is likely in other models too... linePlusBarChart and multiChart for instance...

LeeMendelowitz: @liquidpele Thanks, I'll take a look at the other models.

robinfhu:
Thanks!

robinfhu:
Yes this looks fine.

src/models/lineChart.js
liquidpele: Looks like stackedAreaChart.js, lineWithFocusChart.js, and cumulativeLineChart.js are the only other ones with multiple lines that also have the interactive guideline option.  Not sure if this specific problem happens with those though.  Making null be zero is probably valid for stackedAreaChart and cumulativeLineChart, so probably only lineWithFocusChart needs to be fixed in the same way...

LeeMendelowitz:
@liquidpele Thanks for the guidance, I updated the lineWithFocusChart and stackedAreaChart.

Fix has a smaller visual impact on the stackedAreaChart but logic is the same - tooltip now reads N/A instead of 0.00 for null y values and the point associated with the null y value isn't highlighted.

I looked at cumulativeLineChart, but that model has more complex logic that currently doesn't properly handle null y-values it looks like, so I'm not confident fiddling with it.

src/models/lineChart.js
```diff
@@ -251,7 +254,7 @@ nv.models.lineChart = function() {
                     .position({left: e.mouseX + margin.left, top: e.mouseY + margin.top})
                     .chartContainer(that.parentNode)
                     .valueFormatter(function(d,i) {
-                        return yAxis.tickFormat()(d);
+                        return d == null ? "N/A" : yAxis.tickFormat()(d);
```
liquidpele: Wait, so here you return "N/A" but in the example html you override tickFormat to do the same, but that wouldn't ever happen since it doesn't call tickFormat if it's null... 

I wonder if we should do this instead so they can format nulls how they want?  Just a thought... 

    return d === null ? d : yAxis.tickFormat()(d);

LeeMendelowitz: The change to the line chart example wasn't me, I can back that out.

In my opinion modifying the `yAxis.tickFormat` is a hack - there should be no y-axis tick if y is null, so why would user expect to handle the null value there? It just so happens that the `yAxis.tickFormat` is used by the tooltip - so yes, it is a way for the user to handle custom formatting of null y-values in tooltip. But it means they have to apply this custom formatting to every chart line chart with null y values. And null y-values are advertised as a way to make a line chart discontinuous, so I think its better to just handle it in the tooltip.

robinfhu: Well by default, the formatter will convert a 'null' to a 0, so you could argue that's the default behavior. Users can override that to their liking.

liquidpele: I agree the tooltip should handle nulls by default, looking back at this it's fine I think, they can override the valueFormatter option on the guideline tooltip if they want to change it, I was thinking the overridden function would get passed the "N/A" but that's not the case so I think it's fine.

liquidpele:
I'm ready to merge unless @robinfhu objects.

robinfhu:
I'm cool with it!

liquidpele:
I have made it so.  Thanks for contributing @LeeMendelowitz!

LeeMendelowitz:
Thanks guys!

PR #974, opened 2015-04-20
Added option to SunburstChart to scale to size, not only count

mrijke:
Right now, the size of the 'pieces' of the sunburst chart is not respective to their data, but to the number of children of a given parent.
This commit adds an option "mode", that allows the user to choose the two modes as illustrated 
[here](http://bl.ocks.org/kerryrodden/477c1bfb081b783f80ad).

As an added bonus, one can now display the sum of the children's values in the tooltip of a given parent node as follows:

```
chart.sunburst.dispatch.on('elementMouseover.tooltip', function(evt) {
    evt['series'] = {
        key: evt.data.name,
        value: evt.data.value,
        color: evt.color
    };
    chart.tooltip.data(evt).hidden(false);
});

```

### DIFF
--- a/src/models/sunburst.js
+++ b/src/models/sunburst.js
@@ -9,6 +9,8 @@ nv.models.sunburst = function() {
     var margin = {top: 0, right: 0, bottom: 0, left: 0}
         , width = null
         , height = null
+        , mode = "count"
+        , modes = {count: function(d) { return 1; }, size: function(d) { return d.size }}
         , id = Math.floor(Math.random() * 10000) //Create semi-unique ID in case user doesn't select one
         , color = nv.utils.defaultColor()
         , duration = 500
@@ -68,8 +70,8 @@ nv.models.sunburst = function() {
             y.range([0, radius]);
 
             node = node || data;
-            path =
-                g.data(partition.nodes).enter()
+            partition.value(modes[mode] || modes["count"]);
+            path = g.data(partition.nodes).enter()
                 .append("path")
                 .attr("d", arc)
                 .style("fill", function (d) {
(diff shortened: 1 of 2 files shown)


src/models/sunburst.js
```diff
@@ -68,8 +70,8 @@ nv.models.sunburst = function() {
             y.range([0, radius]);
 
             node = node || data;
-            path =
-                g.data(partition.nodes).enter()
+            partition.value(modes[mode]);
```
liquidpele: You might want to add a default in there if they try to set mode to something crazy... so like:

    partition.vaoue(modes[mode] || modes['count']);

src/models/sunburst.js
```diff
@@ -170,6 +172,7 @@ nv.models.sunburst = function() {
         // simple options, just get/set the necessary values
         width:      {get: function(){return width;}, set: function(_){width=_;}},
         height:     {get: function(){return height;}, set: function(_){height=_;}},
+        mode:       {get: function(){return mode;}, set: function(_){mode=_;}},
```
liquidpele: Please add this option to the documentation by adding it to the large json object at the bottom of examples/documentation.html

liquidpele:
Awesome, I was meaning to do this at some point, thanks!

liquidpele:
Also, and this is just a suggestion i.e. not necessary to merge this, but you might want to add a control menu like what stackedAreaChart and multiBarChart have so you can easily switch between the modes...

mrijke:
Thanks for the comments @liquidpele!
I've added the documentation in the JSON object, and also added 'sunburst' and 'sunburstChart' to the 'charts' and 'parts' array, because otherwise the documentation wouldn't show.

Another thing, the piece of code I wrote in my PR comment overriding the tooltip dispatch doesn't actually work without exposing the tooltip as a public variable (adding it after line 169 in sunburst.js), I changed that before when working on the concatenated nv.d3.js so I didn't notice it before.
Shall I amend that into this PR as well, or do you want a separate PR for that?

liquidpele:
Cool.  Lets merge this and then you can make a new pull request for further stuff.

PR #977, opened 2015-04-20
SunburstChart: Expose 'tooltip' public variable

mrijke:
This PR exposes the 'tooltip' public variable of the SunburstChart, in order to be consistent with the other charts which expose the 'tooltip' as well.


liquidpele:
awesome, thanks!

Issue #971, opened 2015-04-17
Scatter chart tooltip, hover and distribution broken


RenatoUtsch:
If I open the test/ScatterChartTest.html test, you can notice readily that the X and Y distributions are being extended incorrectly.

Also, If I change the width or the height of the page, the tooltip (and the distribution and the hover effect) stops appearing if I mouse over a point. Maybe they are being calculated incorrectly? Dunno.

@liquidpele remember when you noticed this under my poly-nvd3 scatter chart? It seems it is happening with the normal nvd3 outside shadow DOM too.

RenatoUtsch:
Researched a little bit.

The problem with the distributions being extended incorrectly was introduced in commit 5e121c2.

The  resize bug was introduced in an earler commit, will research more.

RenatoUtsch:
Researched a little more.

The resize bug was introduced in commit 60b4ded.

RenatoUtsch:
Removing the line added to models/scatter.js in 60b4ded seems to solve the resize issue.

@robinfhu, why did you add that line? Is it safe to revert it? (it seems so)

robinfhu:
Oh no you can't remove that otherwise this issue will manifest itself: https://github.com/novus/nvd3/issues/783

liquidpele:
The commit you say the distribution bug was in looks like just a merge it doesn't look like the original commit

RenatoUtsch:
@liquidpele
Yes, it was just a merge, I didn't have the time to look which commit it was. But it was in that merge.

@robinfhu 
Then this fix is wrong, because it brought up another issue hehe

robinfhu:
@RenatoUtsch the fix isn't wrong.  If the other issue is brought up that issue should be fixed, but the original behavior was grossly incorrect.  Having a bunch of nv-clip-paths on the DOM is wrong.

RenatoUtsch:
@robinfhu true, sry. Then we need to figure out why this is happening.

liquidpele:
I fixed the resize issue by just redoing some of the logic, see the PR above.

liquidpele:
@RenatoUtsch when you say "The problem with the distributions being extended incorrectly", do you mean how the colored lines that appear on hover that are supposed to go from the point to each axis?  I notice on the testing page that they're not appearing where they should be, or are too long or short.  Is that what you're referring to?

RenatoUtsch:
@liquidpele yeah that's it.

[...]

PR #979, opened 2015-04-21
Fix scatter window resize issue where it stops doing hover on the scatte...

liquidpele:
...rChart properly - https://github.com/novus/nvd3/issues/971

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -129,6 +129,7 @@ nv.models.scatter = function() {
             wrap.classed('nv-single-point', singlePoint);
             gEnter.append('g').attr('class', 'nv-groups');
             gEnter.append('g').attr('class', 'nv-point-paths');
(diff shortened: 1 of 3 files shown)


liquidpele:
@robinfhu pls review :)

robinfhu:
@liquidpele okay looking.  Sorry been a little busy lately.

robinfhu:
Great thanks!

robinfhu:
Out of curiosity, why did removing the nv-point-clips ID solve this issue?

liquidpele:
it didn't I just made it a class name to be more technically correct in case multiple charts exist on the same page... what actually fixed it was removing the clippaths themselves but I'm honestly not sure why that worked...I suspect something with in d3 itself was losing track of the clippings

robinfhu:
Cool thanks!

PR #980, opened 2015-04-21
StackedAreaChart: Fix wrong yAxis tickFormat on chart.update() in Expanded view

mrijke:
Bug: when calling chart.update() in expanded view and then switching to another view the yTickFormat remains the percentage format.

Fix: only set oldYTickFormat when changing from another view to expanded (and not when remaining in expanded view).


liquidpele:

PR #981, opened 2015-04-22
Sunburst: Reset to actual root node on double click

mrijke:
This commit adds a new event handler that resets to the actual root node (from the data perspective) on doubleclick.

Originally I wanted to make it so that it would only happen if you doubleclick on the current root node with a check in the doubleclick handler for `if (node === d)`, but that doesn't work since a double click also fires a single "click" event, which does `node = d` already. This causes the `if` to always be true. 

So now it resets if you double click on any node, which works for my purposes, but I'm open for suggestions on how to make it behave like I described earlier.

### DIFF
--- a/src/models/sunburst.js
+++ b/src/models/sunburst.js
@@ -30,8 +30,10 @@ nv.models.sunburst = function() {
         .innerRadius(function(d) { return Math.max(0, y(d.y)); })
         .outerRadius(function(d) { return Math.max(0, y(d.y + d.dy)); });
 
-    // Keep track of the node that is currently being displayed as the root.
-    var node;
+    // Keep track of the current and previous node being displayed as the root.
+    var node, prevNode;
+    // Keep track of the root node
+    var rootNode;
 
     //============================================================
     // chart function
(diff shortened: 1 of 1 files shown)


liquidpele:
Hmm... well you could probably track what the last clicked arc was from the click event function, and then calculate what the current root was from that element's parent in the data structure?

mrijke:
As far as I understood the code, the `node` variable already tracks which arc was clicked last. If I check for `node.parent == d` in `onDblClick` it will never work since the click event handler already set `node = d`.

I believe that the root cause of the problem is that a double click action also fires single 'click' events. If only the doubleclick event would be fired I could check for `node.parent == d`, since `node` didn't change. 

Some people fixed this by extending d3's dispatcher to allow for a delay period between clicks to determine if it is 1 double click or multiple single clicks (for example [here](https://gist.github.com/tmcw/4067674)), but that is not in the scope of this PR, that should be done in the core.

Another fix that is in the scope of this PR would be to change the single click event handler to only do `node = d` *after* the transition completes. This could work because if you doubleclick the transition from the single click event will not complete. I'll try to do implement this behavior.

liquidpele:
Maybe you could keep it simple and in the single click handler save the current root node as another variable before it's overwritten by d, then use that second variable for the double click check.

mrijke:
You're right @liquidpele, I was thinking too complicated :stuck_out_tongue: 
The new commit should do the trick!

liquidpele:

PR #982, opened 2015-04-22
fix multichart tooltip position on pages with scroll

marcmrf:
the position of the tooltip was wrong depending of the scroll of the page. Tooltip should always appear near to the point.


liquidpele:
This is to master...   in the development branch we completley redid tooltips and fixed all the position issues.   You'll need to check the development branch (or the v1.8.1-alpha release) and verify that fixes all your issues.

PR #983, opened 2015-04-23
Do not use d3.select() to make all charts work with Shadow DOM

RenatoUtsch:
While working in my poly-nvd3 components that use Shadow DOM I used to have some chart features (like the circle that appears where you hover on a line on the lineChart) that didn't work. I discovered that this is because these features use d3.select(), which doesn't pierce Shadow DOM.

The thing is, if we select from the container of the chart instead of the entire document we start searching from the inside of the Shadow DOM. This works. Also, selecting from the container is faster when compared to selecting from the entire document, so it is a performance improvement.

What do you guys think about this? It works, but maybe it can be improved. I added a new member to all nvd3 charts called 'container' that stores the chart container, so that we don't need to select it anymore (what doesn't work in Shadow DOM and is slower). Everything seems to be working and all tests passed.


liquidpele:
There are a ton of just whitespace changes, it's kind of hard to read through the diff, please do not commit lots of whitespace changes together with logic changes please.

RenatoUtsch:
Sry, my vim automatically removes trailing whitespace when I save. I didn't notice.

RenatoUtsch:
So, what about this change? It is very important for my project that uses NVD3.

liquidpele:
Was going to look at it last night but got too drunk, will try again tonight ;)

RenatoUtsch:
Okay, thanks!

RenatoUtsch:
Sooo... not drunk tonight? hahah

src/models/sunburst.js
```diff
@@ -185,6 +186,7 @@ nv.models.sunburst = function() {
         height:     {get: function(){return height;}, set: function(_){height=_;}},
         mode:       {get: function(){return mode;}, set: function(_){mode=_;}},
         id:         {get: function(){return id;}, set: function(_){id=_;}},
+        container:  {get: function(){return container;}, set: function(_){container=_;}},
```
liquidpele: I'm not sure container should ever be set.... shouldn't this be read-only in all cases?

RenatoUtsch: Yeah, it should be read-only, my bad. I just have to set the get function then and remove the set part?

RenatoUtsch:
Actually, the container property should be public to the chart, but not accessible from outside. This would already solve the problems.

RenatoUtsch:
What do you think now, @liquidpele ?

liquidpele:
Looks good to me, I like the idea of saving the d3 object instead of doing selects anyway, seems like much better practice.    @robinfhu any comments?

RenatoUtsch:
Much better and much faster. We should move on to doing this to everything. But it can be done gradually.

liquidpele:
Went ahead and merged this before it got any more stale, since it touched a lot of stuff due to his whitespace changes.

RenatoUtsch:
Thanks. I don't think there'll be any problems.

PR #985, opened 2015-04-23
Allow multiple markers and labels on the bulletchart.

drapergeek:
Currently, the API shows that we can provide an array of markers and labels to the bulletChart but unfortunately, it only displays the first marker.

This adds the ability to pass an array of markers and an array of labels and have them all show up.

I'm happy to make any changes that are recommended, this was just my first pass when I needed this feature.

I'm also happy to squash all the commit messages, I just assume there will be some feedback first so I thought I'd wait!

Thanks!


liquidpele:
This pull request is to master, it should be to the development branch :) 

https://github.com/novus/nvd3#a-few-rules-for-pull-requests

PR #986, opened 2015-04-23
fixing stackedArea tooltip formatting

lucaslin:
Allowing user to specify how to format tooltip header and values like in
other charts (was previously not possible to format values)

Examples summarizing changes: http://nvd3-lucaslin.parseapp.com/stackedarea-tooltip-formatting/

Currently, using `chart.tooltip.valueFormatter()` and  `chart.tooltip.headerFormatter()` works to format tooltips in [other chart models](http://nvd3-lucaslin.parseapp.com/stackedarea-tooltip-formatting/before-change/multibar-tooltip.html) but not those in [stackedAreaChart](http://nvd3-lucaslin.parseapp.com/stackedarea-tooltip-formatting/before-change/just-tooltip.html).
image

There is a [work-around hack](http://nvd3-lucaslin.parseapp.com/stackedarea-tooltip-formatting/before-change/interactivelayer-tooltip.html) that works for header formatting but not value formatting by using `chart.interactiveLayer.tooltip` instead of `chart.tooltip`. 
image
The `valueFormatter()` doesn't work because it is explicitly overwritten (in line 378) by `yAxis.tickFormat()`.
image

Regardless, users should be able to use `chart.tooltip` like in other charts.

The bug comes about because the user-provided tooltip formatting functions are never passed to `interactiveLayer.tooltip`.

By providing  `tooltip.valueFormatter()` and `tooltip.headerFormatter()` to interactiveLayer, the user's function will be used if provided, and the correct default (like `yAxis.tickFormat()` for `valueFormatter()`) will still be used  if not (see line 44).

## Result:
 `tooltip.valueFormatter()` and `tooltip.headerFormatter()` format the values and headers respectively.
[Example of proposed fix](http://nvd3-lucaslin.parseapp.com/stackedarea-tooltip-formatting/after-change/just-tooltip.html)
image

### DIFF
--- a/src/models/stackedAreaChart.js
+++ b/src/models/stackedAreaChart.js
@@ -12,7 +12,6 @@ nv.models.stackedAreaChart = function() {
         , legend = nv.models.legend()
         , controls = nv.models.legend()
         , interactiveLayer = nv.interactiveGuideline()
-        , tooltip = nv.models.tooltip()
         ;
 
     var margin = {top: 30, right: 25, bottom: 50, left: 60}
@@ -41,7 +40,7 @@ nv.models.stackedAreaChart = function() {
     xAxis.orient('bottom').tickPadding(7);
     yAxis.orient((rightAlignYAxis) ? 'right' : 'left');
 
-    tooltip
+    interactiveLayer.tooltip
         .headerFormatter(function(d, i) {
             return xAxis.tickFormat()(d, i);
         })
@@ -375,15 +374,12 @@ nv.models.stackedAreaChart = function() {
                 //If we are in 'expand' mode, force the format to be a percentage.
                 var valueFormatter = (stacked.style() == 'expand') ?
                     function(d,i) {return d3.format(".1%")(d);} :
-                    tooltip.valueFormatter(); 
-                
-                var headerFormatter = tooltip.headerFormatter();
+                    interactiveLayer.tooltip.valueFormatter(); 
 
                 interactiveLayer.tooltip
                     .position({left: pointXLocation + margin.left, top: e.mouseY + margin.top})
                     .chartContainer(that.parentNode)
                     .valueFormatter(valueFormatter)
-                    .headerFormatter(headerFormatter)
                     .data(
                     {
                         value: xValue,
@@ -450,7 +446,6 @@ nv.models.stackedAreaChart = function() {
     chart.xAxis = xAxis;
     chart.yAxis = yAxis;
     chart.interactiveLayer = interactiveLayer;
-    chart.tooltip = tooltip;
 
     chart.dispatch = dispatch;
     chart.options = nv.utils.optionsFunc.bind(chart);
--- a/src/models/stackedAreaChart.js
+++ b/src/models/stackedAreaChart.js
@@ -375,11 +375,15 @@ nv.models.stackedAreaChart = function() {
                 //If we are in 'expand' mode, force the format to be a percentage.
                 var valueFormatter = (stacked.style() == 'expand') ?
                     function(d,i) {return d3.format(".1%")(d);} :
-                    function(d,i) {return yAxis.tickFormat()(d); };
+                    tooltip.valueFormatter(); 
+                
+                var headerFormatter = tooltip.headerFormatter();
+
                 interactiveLayer.tooltip
                     .position({left: pointXLocation + margin.left, top: e.mouseY + margin.top})
                     .chartContainer(that.parentNode)
                     .valueFormatter(valueFormatter)
+                    .headerFormatter(headerFormatter)
                     .data(
                     {
                         value: xValue,


src/models/stackedAreaChart.js
```diff
@@ -375,11 +375,15 @@ nv.models.stackedAreaChart = function() {
                 //If we are in 'expand' mode, force the format to be a percentage.
                 var valueFormatter = (stacked.style() == 'expand') ?
                     function(d,i) {return d3.format(".1%")(d);} :
-                    function(d,i) {return yAxis.tickFormat()(d); };
+                    tooltip.valueFormatter();
```
liquidpele: The problem with this is that if using the interactive guideline, people should be doing chart.interactiveLayer.tooltip.valueFormatter() instead of chart.tooltip.valueFormatter(), so this is not really correct...   instead we should probably just have a variable to track the original value so we can switch back to it...

src/models/stackedAreaChart.js
```diff
@@ -375,11 +375,15 @@ nv.models.stackedAreaChart = function() {
...
-                    function(d,i) {return yAxis.tickFormat()(d); };
+                    tooltip.valueFormatter(); 
+                
+                var headerFormatter = tooltip.headerFormatter();
+
                 interactiveLayer.tooltip
                     .position({left: pointXLocation + margin.left, top: e.mouseY + margin.top})
                     .chartContainer(that.parentNode)
                     .valueFormatter(valueFormatter)
+                    .headerFormatter(headerFormatter)
```
liquidpele: Again, it shoudn't be using chart.tooltip for the value...  we should just give a value as a default here by itself.   

One thing is that if you do chart.update() it'll probably overwrite these values... if it does, then use the _headerFormater() and _valueFormatter() versions of the options... adding an underscore to any option makes it only save the value if the normal option (without the underscore) has never been called.

src/models/stackedAreaChart.js
lucaslin: Okay. I was under the impression that interactiveLayer was something we wanted to abstract away for the user. Why does stackedAreaChart have tooltip by itself at all then? Is it deprecated?

src/models/stackedAreaChart.js
```diff
@@ -12,7 +12,6 @@ nv.models.stackedAreaChart = function() {
         , legend = nv.models.legend()
         , controls = nv.models.legend()
         , interactiveLayer = nv.interactiveGuideline()
-        , tooltip = nv.models.tooltip()
```
lucaslin: Can we remove tooltip altogether? Or is it being kept for legacy or other reasons?

src/models/stackedAreaChart.js
```diff
@@ -41,7 +40,7 @@ nv.models.stackedAreaChart = function() {
     xAxis.orient('bottom').tickPadding(7);
     yAxis.orient((rightAlignYAxis) ? 'right' : 'left');
 
-    tooltip
+    interactiveLayer.tooltip
```
lucaslin: This seems like a logical next step based on the previous feedback (that users should call `chart.interactiveLayer.tooltip` instead of `chart.tooltip` when formatting tooltips).

This edit is using the previous defaults for `tooltip` as the new defaults for `interactiveLayer.tooltip`, though if necessary, the defaults for `tooltip` can be left in as well.

src/models/stackedAreaChart.js
```diff
@@ -375,7 +374,8 @@ nv.models.stackedAreaChart = function() {
                 //If we are in 'expand' mode, force the format to be a percentage.
                 var valueFormatter = (stacked.style() == 'expand') ?
                     function(d,i) {return d3.format(".1%")(d);} :
-                    function(d,i) {return yAxis.tickFormat()(d); };
+                    interactiveLayer.tooltip.valueFormatter();
```
lucaslin: Same idea as previous iteration, but using `interactiveLayer.tooltip` instead.

Users will be able to use `headerFormatter()` and `valueFormatter()` like in the following example:

        chart.interactiveLayer.tooltip
            .headerFormatter(function (d) {
                return 'Header: ' + d;
            })
            .valueFormatter(function (d, i) {
                return d + ' monkeys';
            });

If not specified, the formatters will default to the axis `tickFormat()` functions (specified at line 43) as before.

lucaslin:
[Live example for Iteration 2](http://nvd3-lucaslin.parseapp.com/stackedarea-tooltip-formatting/iteration-2/)

liquidpele:
Hey there lucus.   So basically there are 2 kinds of tooltips, both fully supported.   chart.tooltip is used if you do NOT have the interactiveGuideline option turned on, otherwise chart.interactiveLayer.tooltip is used because interactiveLayer is another layer of logic that uses the tooltip object.  There are similarities and some code could certainly be consolidated but hasn't yet...   it might actually be nice to make the interactiveLayer use the chart's tooltip object instead of it's own, but that's a whole other can of refactor worms.   For this pull request, you should think of them as entirely different tooltip systems. 

Also, your live example doesn't work properly.   

reproduce steps:
1. click the 'expanded' option
2. resize the window so that chart.update() is called
3. click the 'stacked' option.  You'll notice everything is in crazy percents now instead of the values.  

Also-also github says it can't auto-merge anymore...  try merging the main nvd3/development branch into your branch and resolve any conflicts to fix that.

lucaslin:
Thanks! I did try using stackedArea without interactiveLayer, but no tooltips show up. Am I missing something?
http://plnkr.co/edit/IIrrwfn6wFDc0WkzGhfx?p=preview

liquidpele:
Yea, use the development branch for the new tooltip code.   

http://plnkr.co/edit/4vkyC6KEdp3FZVgQSkjG?p=preview

If you want to use v1.7.1 you just need to call chart.interactive(true) but it uses older tooltip code that was buggy so it was refactored recently.

lucaslin:
I believe I found a fix using the same method `yAxis.tickFormat` uses.

See [this plunk](http://plnkr.co/edit/RGxg04vhFvGl3uLTJ3Cy?p=preview).

Change summary can be found here: https://github.com/lucaslin/nvd3/commit/662b46ce6737839009f9477d5a2aabef7102a48b

I will probably need to submit a new pull request for them, since the repos no longer match, but I just wanted to run it by here.

liquidpele:
Yup, probalby best to submit a new PR now... close this one when you do :)

lucaslin:
Continued: #1001

PR #990, opened 2015-04-23
Adding multiBarWithSeriesStackOption model.

ywplee:
This model is similar to the multiBar model, but it has a capability to add stacked option for each series. I ran into a situation where I needed to implement a way to represent aggregated data of the current day and all the previous day's data. 

This model will be useful for a such case where you need to compare how much profit you generated from the 5 stocks (stacked bar chart) and how much profit you generated from all of your portfolios yesterday (which wouldn't be stacked to the stacked bar chart). 

I have provided examples to both examples and test folder.


liquidpele:
It's a neat concept but wouldn't it be better as a feature within the multiBarChart model that already exists than it being a whole separate model?

liquidpele:
For those that didn't understand the description, it basically has the ability to stack only series 0 and 1, so that they are stacked *next* to series 2 (for instance)... so it's like a combination of grouped and stacked.

ywplee:
That's true. I'll try to incorporate the changes into the existing
multiBarChart. Why didn't I think about that :)

PR #991, opened 2015-04-24
Adding individual stack option each series to multiChart model.

ywplee:
I ran into a situation where I needed to implement a way to represent aggregated data of the current day and previous day's summed data.

This will be useful in such a case where you need to compare how much profit you generated from the 5 stocks (stacked bar chart) today and how much profit you generated from all of your portfolios yesterday (which wouldn't be stacked to the stacked bar chart).


examples/multiBarChart.html
```diff
@@ -33,6 +33,7 @@
     var test_data = stream_layers(3,10+Math.random()*100,.1).map(function(data, i) {
         return {
             key: 'Stream' + i,
+            stacked: true,
```
liquidpele: Your use of this attribute isn't backward compatible... by default it should stack like it has always done, and the new attribute should make the behavior change - i.e. that it should be un-stackable.

ywplee: Got it. I'll make that change

liquidpele:
Awesome, looks good at a glance, I'll wait a bit to see if @robinfhu wants to take a gander too

src/models/multiBar.js
```diff
@@ -245,19 +253,38 @@ nv.models.multiBar = function() {
...
-                        return y(d.y1);
+                    .attr('y', function(d,i,j) {
+                        if (stacked && !data[j].nonStackable) {
+                            return y(d.y1);
+                        } else {
+                            return getY(d,i) < 0 ?
+                                y(0) :
+                                    y(0) - y(getY(d,i)) < 1 ?
+                                y(0) - 1 :
+                                y(getY(d,i)) || 0;
```
liquidpele: using a ternary operator inside another ternary operator is hard to read...

liquidpele:
I just went ahead an merged it, it's a pretty cool feature.

liquidpele:
Well, of course right after I merge it I find a problem... 

When you disable "stream1" which is light blue it turns orange and then starts getting hidden...  

Also, I just noticed that the non-stackable line overlaps the others, so they're not next to one another... that means you can only ever have a single non-stackable line, wouldn't it be better to have them next to each other so that you can have more than one?

liquidpele:
Paging @ywplee :)

src/models/multiBar.js
ywplee: Personally, I don't like using ternary operator. I used it because it was used in the existing code to keep the consistency. I don't mind taking out the ternary operator :)

ywplee:
For turning to a different color when hiding, can you suggest me some direction to mitigate the problem? I'm not completely sure how to fix this problem since it seems to be doing it on  the stacked series already. It's less noticeable since everything is stacked. 

For the second issue, yes I was aware of that. I thought the overlapping the non-stackable series would be a good idea since "grouped" provides the feature of having the series next to each other, but I could see why this could cause potential problems. If you don't have any other suggestion, I'll make non-stackable series be drawn on next to each other.

ywplee:
@liquidpele hey, I guess you didn't get a chance to read my comment. Do you have a suggestion for the color changing issue?

liquidpele:
Yea and further changes has to be a new pull request.   

Yea, I can see where the color changing happened previously too, so no worries there... unless you jsut feel like trying to fix it :)

src/models/multiBar.js
```diff
@@ -66,14 +66,22 @@ nv.models.multiBar = function() {
                         };}
                 )}];
 
-            if (stacked)
-                data = d3.layout.stack()
+            if (stacked) {
+                var parsed = [];
+                var parsed = d3.layout.stack()
```
robinfhu: You have 'parsed' declared twice.

ywplee: Oh yeah, let me fix that up. I have another bug fix as well.

PR #992, opened 2015-04-24
LineWithFocusChart: Fix setting tickFormat, duration

mrijke:
* Fix x- and yTickFormat not calling correct functions on Axis
  * Fix duration not being set on x2Axis and y2Axis


liquidpele:
Cool, thanks!

PR #995, opened 2015-04-24
Fix percent bug in pie chart

jstrength:
Hey guys I noticed a bug in the pie chart when you specify the labelType as percent.
Here is a before and after of the problem:
screen shot 2015-04-24 at 4 09 08 pm
screen shot 2015-04-24 at 4 10 11 pm


liquidpele:
cool thank you for contributing!

Issue #861, opened 2015-03-10
pie/donut chart labels are in wrong positions when a start and end angle are given


liquidpele:
On the examples page, the pie and donut that are a half circle have labels show up as if the chart was a full circle.  This is probably the same situation as this PR:  https://github.com/novus/nvd3/pull/818

jstrength:
I think this can be closed now

liquidpele:
Yup

PR #996, opened 2015-04-25
fixed pie/donut chart wrong label positions bug

jstrength:
Fixed bug #861 
Had to change the logic to not create an entire new arc object and instead use the one already created before settings its innerRadius to zero.
Here is a before and after:
croppercapture 4
croppercapture 5

### DIFF
--- a/src/models/pie.js
+++ b/src/models/pie.js
@@ -213,10 +213,8 @@ nv.models.pie = function() {
                             if (startAngle !== false) labelsArc[i].startAngle(startAngle);
                             if (endAngle !== false) labelsArc[i].endAngle(endAngle);
                         }
-                    } else {
-                        if (!donut) {
+                    } else if (!donut) {
                             labelsArc[i].innerRadius(0);
-                        }
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


src/models/pie.js
```diff
@@ -205,15 +205,17 @@ nv.models.pie = function() {
...
                         if (donut) {
                             labelsArc[i] = d3.svg.arc().outerRadius(arcs[i].outerRadius());
                             if (startAngle !== false) labelsArc[i].startAngle(startAngle);
                             if (endAngle !== false) labelsArc[i].endAngle(endAngle);
-                        } else {
-                            labelsArc[i] = arcs[i];
+                        }
+                    } else {
+                        if (!donut) {
+                            labelsArc[i].innerRadius(0);
```
liquidpele: Are you sure this works for labelsOutside() being both true and false?    

Also it should just do an "else if", not a whole if block inside the else block.

jstrength:
Yes it does, I just check again, with donut false/true and labelsOutside false/true.

liquidpele:
awesome, thanks!

PR #1001, opened 2015-04-29
fixing stackedArea tooltip formatting

lucaslin:
allowing users to set valueFormatter for interactiveLayer.tooltip (it
was previously being overwritten)

This change should also be implemented on other charts that use
interactiveLayer.tooltip

Newer iteration of #986

I believe I found a fix using the same method `yAxis.tickFormat` uses.

See [this plunk](http://plnkr.co/edit/RGxg04vhFvGl3uLTJ3Cy?p=preview).

### DIFF
--- a/src/models/stackedAreaChart.js
+++ b/src/models/stackedAreaChart.js
@@ -389,25 +389,22 @@ nv.models.stackedAreaChart = function() {
 
                 var xValue = xAxis.tickFormat()(chart.x()(singlePoint,pointIndex));
 
+                var valueFormatter = interactiveLayer.tooltip.valueFormatter();
                 // Keeps track of the tooltip valueFormatter if the chart changes to expanded view
                 if (stacked.style() === 'expand' || stacked.style() === 'stack_percent') {
                     if ( !oldValueFormatter ) {
-                        oldValueFormatter = interactiveLayer.tooltip.valueFormatter();
+                        oldValueFormatter = valueFormatter;
                     }
                     //Forces the tooltip to use percentage in 'expand' mode.
-                    interactiveLayer.tooltip.valueFormatter(yAxis.tickFormat(d3.format('%')));
+                    valueFormatter = d3.format(".1%");
                 }
                 else {
                     if (oldValueFormatter) {
-                        interactiveLayer.tooltip.valueFormatter(oldValueFormatter);
+                        valueFormatter = oldValueFormatter;
                         oldValueFormatter = null;
                     }
                 }
 
-                //If we are in 'expand' mode, force the format to be a percentage.
-                var valueFormatter = (stacked.style() == 'expand') ?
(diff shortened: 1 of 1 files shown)


src/models/stackedAreaChart.js
```diff
@@ -379,10 +389,25 @@ nv.models.stackedAreaChart = function() {
...
+                        interactiveLayer.tooltip.valueFormatter(oldValueFormatter);
+                        oldValueFormatter = null;
+                    }
+                }
+
                 //If we are in 'expand' mode, force the format to be a percentage.
                 var valueFormatter = (stacked.style() == 'expand') ?
                     function(d,i) {return d == null ? "N/A" : d3.format(".1%")(d);} :
-                    function(d,i) {return d == null ? "N/A" : yAxis.tickFormat()(d); };
+                    interactiveLayer.tooltip.valueFormatter();
```
liquidpele: Doing this means once it goes to expand it saves that as the valueFormatter so then this returns that so it would always stay that way...

liquidpele: Oh wait, I see that it gets overwritten above too...  this logic is looking confusing and should probably be combined with what's above, and then only save the valueFormatter below where it's already doing it...

lucaslin: you mean combining it with this part?

             if (showYAxis) {
                 var ticks;
                 if (stacked.offset() === 'wiggle') {
                     ticks = 0;
                 }
                 else {
                     ticks = nv.utils.calcTicksY(availableHeight/36, data);
                 }
                 yAxis.scale(y)
                     ._ticks(ticks)
                     .tickSize(-availableWidth, 0);
 
                     if (stacked.style() === 'expand' || stacked.style() === 'stack_percent') {
                         if ( !oldYTickFormat )
                             oldYTickFormat = yAxis.tickFormat();
                         //Forces the yAxis to use percentage in 'expand' mode.
                         yAxis.tickFormat(d3.format('%'));
                     }
                     else {
                         if (oldYTickFormat) {
                             yAxis.tickFormat(oldYTickFormat);
                             oldYTickFormat = null;
                         }
                     }
 
                 g.select('.nv-y.nv-axis')
                     .transition().duration(0)
                     .call(yAxis);
             }

liquidpele: No, I mean lines 392 through 410 should be simplified and just set the valueFormatter variable, which then gets saved later on line 414.

lucaslin:
Sorry, I'm not quite sure I understand. My reason for doing this extra logic is to prevent the undesirable behavior when switching between views (stacked --> expanded, resize --> stacked) we saw on the previous iteration, which was caused by the chart using the previous view's formatter. 
I'm probably just missing something.

liquidpele:
yea, that's basically what I was thinking :)

lucaslin:
Thanks! (I submitted a new iteration)

liquidpele:

PR #1007, opened 2015-04-30
exposing get/set stackedAreaChart controls

lucaslin:
stackedAreaChart defines `cData = ['Stacked','Stream','Expanded']`, but
does not provide accessors. Users can't select which to show and need to
choose to display all or none of the controls. Also, it seems the
'Stack_Percent' option can currently never be set.

[Example Plunk](http://plnkr.co/edit/fj83zH2vdOYOfniG5UVK?p=preview) without "Stream" shown


liquidpele:
The cData body has values that are used in the logic later, that should never be changed.   If people want to change the words themselves, you can use the controlLabels option.

liquidpele:
Oh, I see, you wanted to remove just a single option from the control... I guess I understand the use case, but the option shoudl be named better.   cData isn't very descriptive for an option call... probably need to call it something like "controlOptions" ?    I dunno, does that sound right or can you think of a better name?

lucaslin:
That sounds perfect

PR #1008, opened 2015-04-30
Adding a safer version of getComputedTextLength

pinny13:
Calling native getComputedTextLength on invisible elements will cause an exception ("Unexpected call to method or property access"). It can happen for several reasons. In my single page AngularJS app I got it because I was doing some DOM manipulations on promise.resolve, but user navigated to another page and then got back quickly, however the resolve code was still executing and IE threw an exception because while user went on another page the DOM has changed and text element was not there at some point. I also heard people getting this error in some scenarios using iframe, and older version of IE.
This method makes it safer to get a text length by attempting to check whether we can make this call and if not falls back to existing NVD3 method of calculating approximate text width.

### DIFF
--- a/src/models/furiousLegend.js
+++ b/src/models/furiousLegend.js
@@ -178,16 +178,7 @@ nv.models.furiousLegend = function() {
                 var seriesWidths = [];
                 series.each(function(d,i) {
                     var legendText = d3.select(this).select('text');
-                    var nodeTextLength;
-                    try {
-                        nodeTextLength = legendText.node().getComputedTextLength();
-                        // If the legendText is display:none'd (nodeTextLength == 0), simulate an error so we approximate, instead
-                        if(nodeTextLength <= 0) throw Error();
-                    }
-                    catch(e) {
-                        nodeTextLength = nv.utils.calcApproxTextWidth(legendText);
-                    }
-
+                    var nodeTextLength = nv.utils.getComputedTextLength(legendText.node());
                     seriesWidths.push(nodeTextLength + padding);
                 });
 
@@ -244,7 +235,7 @@ nv.models.furiousLegend = function() {
                     xpos;
                 series
                     .attr('transform', function(d, i) {
-                        var length = d3.select(this).select('text').node().getComputedTextLength() + padding;
+                        var length = nv.utils.getComputedTextLength(d3.select(this).select('text')) + padding;
                         xpos = newxpos;
 
                         if (width < margin.left + margin.right + xpos + length) {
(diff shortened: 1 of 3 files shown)


liquidpele:
What in the world is furious legend?   When did that get added??  Also this broke unit tests....

pinny13:
I just forked the development branch couple of hours ago and the furious legend was there I guess...
The broken test is weird... I ran the changed code locally and it seemed working fine for me.

liquidpele:
@twolfnovus looks like you added furious legend... What is it for?

pinny13:
A thought about broken tests: Is it possible that the test is not taking into consideration exactly the scenario I am trying to fix? Meaning maybe it runs ?in a headless browser? and encounters a situation where a an element is not visible and thus expecting a wrong results? My new code still tries to use the same native getComputedTextLength() first as before in all six cases and ONLY if this method pukes the code falls back to nv.utils.calcApproxTextWidth,which was already used as a fallback in two out of six of those cases that I modified. That is why I wouldn't rule a possibility of looking at the test too.

lucaslin:
Running the code locally, I see nothing wrong visually in the UI. I can't seem to run the tests though... upon running `grunt` I consistently get "Warning: Task "karma:unit" failed...". 
image


Is there maybe something obvious I'm missing?

liquidpele:
Looks like karma can't find d3, you probably need to run "npm install"

https://github.com/novus/nvd3#tips-for-testing

lucaslin:
"This may not work on Windows machines."
that might be the issue

thetylerwolf:
Furious legend is an alternative design for the chart legend. It's a UX improvement providing better affordance illustrating the user's ability to turn different series' on and off (set 'vers' to 'furious'). It also enables the ability to toggle series' from showing in the legend (when 'expanded' is set to true). The code was written such that a developer does not need to change their current code in order to keep the 'classic' flavor of legend.

This was motivated by user research that showed many users are unaware that the legend is interactive in its default configuration. Also, when a very large number of series' are present on a chart, the legend becomes massive and takes up a lot of space, sometimes interfering with the chart. We've found this functionality incredibly useful at Novus.

liquidpele:
@thetylerwolf is it used in one of the examples, or can an example be made?  Sounds neat.

thetylerwolf:
Yes, you can find an example at /examples. It's not a perfect example, but I think it gets the point across

liquidpele:
Very cool, thanks!   @pinny13 sorry for the side conversation ;)

liquidpele:
Sorry, I don't have time to debug the unit tests, but just glancing at the travis ci log it's failing on expecting the legend to be a particular translate, so you probably just need to validate that the new values are correct and then change the unit tests to be the new values.

liquidpele:
Any progress on fixing the unit tests @pinny13 ?

pinny13:
I couldn't get the test running locally. Project priorities shifted and I didn't yet have a chance to make it work.

Issue #1014, opened 2015-05-04
Stacked area chart uses percent formatter for other types


jgrund:
The stacked area chart will use the percent formatter for numbers permanently if the chart is updated while a user is looking at the expanded style. Relevant code:

```
  if (stacked.style() === 'expand' || stacked.style() === 'stack_percent') {
    oldYTickFormat = yAxis.tickFormat();
    //Forces the yAxis to use percentage in 'expand' 
    yAxis.tickFormat(d3.format('%'));
  }
  else {
    if (oldYTickFormat) yAxis.tickFormat(oldYTickFormat);
  }

  g.select('.nv-y.nv-axis')
    .transition().duration(0)
    .call(yAxis);
```

To solve this, we could:

a) Cache the `d3.format('%')` call. That way we can compare if the current `tickFormat` is the same as our reference. If it is, we don't need to do an assignment, and we don't lose the original reference.

b) hold an object literal of styles to formatters, so we don't have the possibility to overwrite the "other" formatter when we update twice with expand or stack_percent.

I prefer the former personally. 

In a larger sense, I wonder if overwriting the tick format for percentage cases is what we want to do. 
It seems we take away control from the user in that they cannot override for that layout.

Perhaps exposing a method on this chart to override the formatting. That way, we could detect the user requested an override and not insert our custom code in that case.

liquidpele:
I think this was *just* fixed this here:

https://github.com/novus/nvd3/pull/1001

jgrund:
@liquidpele Nope, actually I wonder if that code will have the same issue I'm describing here. The issue is an update that forces the oldYTickFormat / oldValueFormatter to hold our percent formatter.

PR #1015, opened 2015-05-04
Fixes: #1014.

jgrund:
This patch modifies how we assign the yAxis valueFormatter so we don't
overwrite the original yAxis with out percent formatter.

### DIFF
--- a/src/models/stackedAreaChart.js
+++ b/src/models/stackedAreaChart.js
@@ -89,6 +89,8 @@ nv.models.stackedAreaChart = function() {
         }
     };
 
+    var percentFormatter = d3.format('%');
+
     function chart(selection) {
         renderWatch.reset();
         renderWatch.models(stacked);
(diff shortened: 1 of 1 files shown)


liquidpele:
Can you describe the exact situation that causes an issue that this fixes?

jgrund:
Sure, the situation is as follows:

1) User is on the expanded style of a stacked area chart.
2) The chart updates *twice* on the expanded style.
3) The `oldYTickFormat` takes on the function `d3.format('%')`.
4) No matter what style the user selects, the y axis shows it in percentages. This is because `oldYTickFormat` is using our percent formatter by accident and there is no way to reset it.

This patch fixes it because we never get into the situation where we assign our formatter to `oldYTickFormat`. We save a reference to it outside the chart and simply compare by reference to ensure we don't cache it in `oldYTickFormat`. This means, the original formatter will be preserved / restored when needed.

liquidpele:
Cool, looks good then, thanks for contributing!

Issue #976, opened 2015-04-20
yAxis min becomes -1 when all points' y values are 0 (stackedAreaChart)


lucaslin:
Not sure if this is intentional, but the behavior seems odd, since the default is to always show the yAxis starting at 0, even when some (but not all) points' y values are 0.

See Plunk:
http://plnkr.co/edit/umasgIoekdoWcVjvigHM?p=preview

Seems like it's a result of stackedArea's dependence on scatter:

            // If scale's domain don't have a range, slightly adjust to make one... so a chart can show a single data point
            if (x.domain()[0] === x.domain()[1] || y.domain()[0] === y.domain()[1]) singlePoint = true;
            if (x.domain()[0] === x.domain()[1])
                x.domain()[0] ?
                    x.domain([x.domain()[0] - x.domain()[0] * 0.01, x.domain()[1] + x.domain()[1] * 0.01])
                    : x.domain([-1,1]);

            if (y.domain()[0] === y.domain()[1])
                y.domain()[0] ?
                    y.domain([y.domain()[0] - y.domain()[0] * 0.01, y.domain()[1] + y.domain()[1] * 0.01])
                    : y.domain([-1,1]);

            if ( isNaN(x.domain()[0])) {
                x.domain([-1,1]);
            }

            if ( isNaN(y.domain()[0])) {
                y.domain([-1,1]);
            }

This line in particular `y.domain()[0] ?` seems to be checking for null, NaN, etc., but the value 0 also fails this condition.

lucaslin:
Maybe we can add a `defaultDomain` option that the user can specify?

lucaslin:
Also, part of this domain checking code is [copied across many charts](https://github.com/novus/nvd3/search?l=javascript&q=%5B-1+%2C1%5D&utf8=%E2%9C%93). Maybe we can create a `util` function to perform it as well? Something like:

    updateDisplayDomain(scale, default)

(probably with a better name)

liquidpele:
It looks like stackedAreaChart builds the y axis scale using stackedArea.yScale() which inherits it from scatter.js, so there is probably a calculation within scatter that causes that...

lucaslin:
Yes. That's where the code above is from. Is it possible that we could allow the user to set the default domain?

For my use-case, it's weird that all my stackedAreaCharts' y-axes start at 0, unless all the data points are 0, in which case it starts at -1.

[...]

Issue #968, opened 2015-04-17
Question: how to hide negative range in bar chart if all values are zero?


zocoi:
Currently it looks like this for me. It's ok to have zero values series, I just don't want to display the negative range.

![localhost 8000 examples historicalbarchart html](https://cloud.githubusercontent.com/assets/212200/7194567/e1140728-e465-11e4-8651-85e9c25b74c9.png)

liquidpele:
Well, that's just d3 trying to display the data you gave it.   You can override things manually by doing stuff like chart.yAxis.range([0,1]), but then it's set to that range going forward so you'd have to try and set it back to null or something if you then changed the data and wanted it to scale things automatically again.

zocoi:
@liquidpele manually set the range on yAxis doesn't work. Tracing the code, I can see the range is set properly, I just couldn't figure out why. Here is a jsFiddle:

http://jsfiddle.net/vkk1tcey/

Issue #602, opened 2014-06-25
Unable to use forceY in stackedAreaCharts


JamesFunk:
I think this is due to a line in the stackedArea.js model, where a scatter model is set up with forceY([0]). It makes sense that a stackedAreaChart has 0 forced into the y-axis, but this also seems to override any other forceY calls made prior to the chart being drawn or updated.

liquidpele:
Probably still an issue.

mausumi345:
I am facing same issue.Is there a way to fix it?

PR #1016, opened 2015-05-04
allowing use of forceY in stackedArea via scatter

lucaslin:
Currently there is no way to use forceY on stackedArea, since stackedArea overwrites its value to [0]. This change checks if the user has provided values to forceY and only if they haven't, overwrite it with [0].

[Working Example](http://codepen.io/anon/pen/bdVdga) (Plunker seems to be down)

Solution to #602, #968, #976

Though does not allow cutting off some data from being displayed (not same as max/min) 

The underlying domain issue (described in #976) should probably still be addressed.

### DIFF
--- a/src/models/stackedArea.js
+++ b/src/models/stackedArea.js
@@ -97,7 +97,13 @@ nv.models.stackedArea = function() {
             gEnter.append('g').attr('class', 'nv-scatterWrap');
 
             wrap.attr('transform', 'translate(' + margin.left + ',' + margin.top + ')');
-
+            
+            // If the user has not specified forceY, make sure 0 is included in the domain
+            // Otherwise, use user-specified values for forceY
+            if (scatter.forceY().length == 0) {
+                scatter.forceY().push(0);
+            }
+            
             scatter
                 .width(availableWidth)
                 .height(availableHeight)


lucaslin:
Perhaps a next step would be to allow users to call forceY on chart? (have it be an exposed function in stackedAreaChart and/or stacjedArea that uses the method from scatter)

src/models/stackedArea.js
```diff
@@ -97,7 +97,13 @@ nv.models.stackedArea = function() {
             gEnter.append('g').attr('class', 'nv-scatterWrap');
 
             wrap.attr('transform', 'translate(' + margin.left + ',' + margin.top + ')');
-
+            
+            // If the user has not specified forceY, make sure 0 is included in the domain
+            // Otherwise, use user-specified values for forceY
+            if (scatter.forceY().length == 0) {
+                scatter.forceY().push(0);
+            }
```
liquidpele: You should be able to just do scatter._forceY([0])

Adding an underscore before an option means it's an internally-set option and that if a user overrides it themselves with forceY then it'll ignore the value from _forceY after that.

liquidpele:
forceY is already on the main chart, it's inherited. 

http://nvd3-community.github.io/nvd3/examples/documentation.html#stackedAreaChart

lucaslin:
`chart._forceY([1, 0])` causes the error "Uncaught TypeError: chart._forceY is not a function".
`chart.forceY([1, 0])` does not do anything (expected, since it is overwritten), even though this is what the documentation gives as an example.
`chart.stacked.scatter._forceY([1, 0])` does nothing.
`chart.stacked.scatter.forceY([1, 0])` does nothing.

Is that what you were suggesting?

liquidpele:
Hmm maybe a bug in how I implemented the underscore stuff...  Oh well

mausumi345:
Is there a way in stacked area chart to start the y axis values from 0 instead of -1...I tried setting by force and domain option.but it is not working

meenu1:
In case of zero values, when using forceY: [0, 1].  y axis will not come -1. it will start from 0  to 1 though tick values will be in decimal like 0.2,0.4 etc

PR #1022, opened 2015-05-06
Bug fixes for multiBar model

ywplee:
- Removing duplicate declaration of 'parsed' 
- In stacked mode, clicking the lastly positioned serie's legend throws an error. It was trying to access the deleted reference. I added a null check.

### DIFF
--- a/src/models/multiBar.js
+++ b/src/models/multiBar.js
@@ -190,7 +190,7 @@ nv.models.multiBar = function() {
                 .attr('y', function(d, i, j) {
                     var yVal = y0(0) || 0;
                     if (stacked) {
-                        if (!data[d.series].nonStackable) {
+                        if (data[d.series] && !data[d.series].nonStackable) {
                             yVal = y0(d.y0);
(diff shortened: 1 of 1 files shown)


liquidpele:
Awesome, thanks!

ywplee:
Np. By the way, do you guys have some type of lint task that I could run on
the source code? I noticed that there is jshint task registered in Grunt,
but wasn't sure if jshintrc is up to date or not.

liquidpele:
It would be nice to add jshint to the build eventually, but it hasn't been done since it would be a lot of work to fix all the errors first...

ywplee:
Got it. Moving forward, can I rely on the jshint rule you guys have specified for the repository to lint my code?

liquidpele:
Never used whatever is configured there yet, so not sure... maybe @robinfhu has?

robinfhu:
jshint was just never really used, I think because there were so many errors it would take a while to just fix them all.

PR #1034, opened 2015-05-13
Add bower install to build instructions

ehjay:
I think that `bower install` is missing from the build instructions


liquidpele:
No, building doesn't require any prefrequisites.  Running the unit tests does, and is handled by npm install.

PR #1039, opened 2015-05-14
create feature to export nvd3 chart as pdf/png

dm89:
It would be very helpful to be able to export the image created by nvd3 as a pdf/png. Some options available are not compatible with all browsers and others do not export the background/css properties of div elements.


liquidpele:
Holy crap I think your branch wasn't up to date!

Issue #1025, opened 2015-05-07
Scatter chart changing point colors when it shouldn't


RenatoUtsch:
This is (part) of a chart that I am using internally on my work:
http://jsfiddle.net/mhv069e6/5/

Never mind the tooltip not showing the actual values (is this a bug? I am actually using a custom tooltip with this so I don't know).

If you play with the chart by removing some points (for example, removing the "Consumo Cíclico" bubbles) and adding them again, sometimes the colors will change in the chart. Running the chart multiple times and testing may be necessary to reproduce this behaviour, as this doesn't happen everytime.

For example, this is the original chart that is generated:
![screen shot 2015-05-07 at 5 12 56 pm](https://cloud.githubusercontent.com/assets/836333/7525012/576b2c54-f4dc-11e4-96ac-a1e966758b4f.png)

After some fiddling, the "Financeiro"'s color changes to red. Note that the legend is still with the original color:
![screen shot 2015-05-07 at 5 11 05 pm](https://cloud.githubusercontent.com/assets/836333/7525042/a793459a-f4dc-11e4-80ee-1cbc80fb830e.png)

RenatoUtsch:
Could reproduce the problem with just a basic set of configs, so it seems to be general:
http://jsfiddle.net/mhv069e6/6/

RenatoUtsch:
I've noticed a pattern: the colors change if I open and close the development console. So maybe it is related to the window resize?

PR #1043, opened 2015-05-17
fix for issue 1025

ehjay:
I was able to reproduce #1025 reliably with these steps:

1. Use a data object where each series does not have a `color` attribute set (ex. series A, B, C)
2. Hover over a point from series A to display the tooltip
3. Cause a chart update (clicking a legend item, resizing the browser window)
4. All colors will shift to the right (B has A's color and C has B's color)

The problem:

The tooltip function needs a series color to display, so it sets the series A color using the color(d, i) function.

Then, when the chart is updated, series A has a color. In the current scatterChart model, if a series has a color then the color function is not called. So, color(d, i) is called for the first time to get the color for series B. Thus, B gets the first color (the color that A had on the first draw).

I think the best solution is just to set the series color if one does not exist. Then, we don't have to worry about series colors getting out of sync if a color is set later on.

A unit test is included that makes sure a color is set if none is specified. It fails without the pull request and passes after.

Thanks for reviewing, nvd3 is a great library!

### DIFF
--- a/src/models/scatterChart.js
+++ b/src/models/scatterChart.js
@@ -178,7 +178,8 @@ nv.models.scatterChart = function() {
                 .width(availableWidth)
                 .height(availableHeight)
                 .color(data.map(function(d,i) {
-                    return d.color || color(d, i);
+                    d.color = d.color || color(d, i);
+                    return d.color;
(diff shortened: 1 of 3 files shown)


robinfhu:
Cool, I think this is fine. Small change.

Issue #1052, opened 2015-05-22
Bullet char - number of ticks is "hardcoded"


tothand:
Hi,

I've just started to use nvd3.js (which is really cool and very easy to use) and I stumbled upon a problem with the number of ticks in bullet charts.

When I have big numbers like thousands, millions as ticks it gets unreadable, too dense (like in the next image)
(sometimes it's ok to scale down, divide it by 1000 or a million and add a subtitle that US$ in millions, but there are cases when the actual number should be there)

![nvd3_bulletchar_ticks_orig](https://cloud.githubusercontent.com/assets/3285907/7775681/8376bf68-00b5-11e5-81f6-cd3fd61dad4d.png)

I've found the relevant code here:
```javascript
 2308       // Compute the tick format.
 2309       var format = tickFormat || x1.tickFormat( availableWidth / 100 );
 2310 
 2311       // Update the tick groups.
 2312       var tick = g.selectAll('g.nv-tick')
 2313           .data(x1.ticks( availableWidth / 100 ), function(d) {
 2314             return this.textContent || format(d);
 2315           });
 2316 
```

The number of ticks are calculated by dividing the availableWidth with 50 so I've changed it to 100
(as shown in the code above, line 2313) and I got a better result (see screenshot 2 below)

![nvd3_bulletchar_ticks_mod](https://cloud.githubusercontent.com/assets/3285907/7775761/35b509e6-00b6-11e5-8fc3-5e4d4a4eeafa.png)

I'm quite new to d3 and nvd3 so I can't really do a proper fix to add a new parameter for the number of ticks or for the number to divide and create a pull request so I'm kindly asking if someone can do that.

In the meantime I'm ok with my 'patch' for the current project but it's far from a fix.

Thanks,
Andrew

liquidpele:
Cool, if you could submit a pull request that makes the input to x1.ticks() be an option with availableWidth/50 as the default that would be great.   Please submit the pull request to the development branch, not master :)

Basically, you can just make a variable defined at the top named (for instance) "numTicks = null;" that gets set by the option if they call chart.ticks(5) for instance, and then in the above code you referenced you can do x1.ticks(numTicks ? numTicks : availableWidth / 50).   That way it tries a default, but anyone can override the number of ticks to whatever they need so it will meet any use case :)

PR #1054, opened 2015-05-22
created an option to configure the number of ticks for bulletChart

tothand:
This is an enhancement resolving #1052

I have to idea how to write a test for this, sorry about that.

One thing I've noticed that the number of ticks won't be exact in some cases.
I've tried the bulletChart example (the last one with custom labels) with
ticks(2) -> result is 2 ticks
ticks(3) -> result is 4 ticks
ticks(5) -> result is 7 ticks
ticks(10) -> result is 7 ticks
ticks(20) -> result is 16 ticks
ticks(30) -> result is 30 ticks

I assume it is related to the d3.js ticks function.

### DIFF
--- a/src/models/bulletChart.js
+++ b/src/models/bulletChart.js
@@ -21,6 +21,7 @@ nv.models.bulletChart = function() {
         , width = null
         , height = 55
         , tickFormat = null
+	, ticks = null
         , noData = null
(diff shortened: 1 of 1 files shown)


liquidpele:
Looks good, thanks!

PR #1055, opened 2015-05-24
Feature: add scatters to multiChart model

ehjay:
I've been using multiChart for lines and bars, but I'd also like to add scatters to my charts. This request extends the multiChart.js model and multiChart.html example to support scatters as well.

Since we're just adding a scatter model to the multiChart model, scatter.coffee provides unit test coverage. I've updated the multiChart.html example and inspected visually. It works as expected; I don't see any glaring errors.

Thanks for reviewing! Let me know if you have any feedback

### DIFF
--- a/examples/multiChart.html
+++ b/examples/multiChart.html
@@ -30,7 +30,7 @@
 
 <script>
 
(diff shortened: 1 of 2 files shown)


liquidpele:
sweet, thanks for contributing!

PR #1056, opened 2015-05-25
Support for error bars in MultiBarChart

lomereiter:
Also added the 'errorBarColor' option to both MultiBarChart and MultiBarHorizontalChart

### DIFF
--- a/examples/multiBarChart.html
+++ b/examples/multiBarChart.html
@@ -45,7 +45,8 @@
             values: new d3.range(0,11).map( function(f,j) {
                 return {
                     y: 10 + Math.random()*100 * (Math.floor(Math.random()*100)%2 ? 1 : -1),
-                    x: j
+                    x: j,
+                    yErr: [-Math.random() * 30, Math.random() * 30]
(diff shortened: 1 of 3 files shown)


robinfhu:
You have an update to the example file so we can see what this looks like?

lomereiter:
Added an example. The usage and appearance are basically the same as in MultiBarHorizontal.

robinfhu:
Looks okay. There's a bar that seems to have one of the error markers cut off.  It's possible this is due to the yScale not taking into account the maximum value it can have in the domain.
multibar with errors

lomereiter:
OK, thanks for the input. I also see that transitions are not applied to the chart after my changes. Don't merge until I'll fix these.

liquidpele:
@lomereiter I went ahead and merged this, but I did notice that the initial transition makes the bars fly over from the left instead of just growing up from the position they should already be at....   can you take a look at fix that?   I made an issue for it here:   https://github.com/novus/nvd3/issues/1076

PR #1060, opened 2015-06-05
Added maxKeyLength variable to truncate legend key values.

gurbirkalsi:
This fix provides an additional variable to set a fixed max key length
value to truncate any legend key values with lengths over the specified
threshold. This provides the option to organize legend keys to be seen
in a more readable manner.


gurbirkalsi:
Saw the failure with PieChart, working on a fix.

liquidpele:
Going okay?

gurbirkalsi:
Yes! PieChart was seeing d.key as undefined so used the getKey variable to retrieve the legend key.

gurbirkalsi:
Support for tooltips on truncated text coming soon

src/models/legend.js
```diff
@@ -176,8 +177,22 @@ nv.models.legend = function() {
...
+                series.each(function(d,i) {
+                  if (getKey(d).length > largestKey) {
+                    largestKey = getKey(d).length;
+                  }
+                });
+
                 series.each(function(d,i) {
-                    var legendText = d3.select(this).select('text');
+                    var legendText;
+                    if (largestKey > maxKeyLength && getKey(d).length > maxKeyLength) {
```
liquidpele: Why do you calculate largestKey, shouldn't you just check if each one is larger than the maxKeyLength?  largestKey will always be larger if any of them are larger and you're doing both calculations so your'e basically doing the same thing twice...

gurbirkalsi: Correct, initially largest text was only taking into account one key; however, getKey(d).length will account for all the keys in the legend. I agree, I can remove largestKey as its redundant. Will resubmit a patch.

liquidpele:
@gurbirkalsi you have other commits in your pull request and you have stuff from the build directory in there as a result too...  you'll need to fix this PR or re-submit it with only your changes for this specific thing.   

Every time you update your branch, the PR will be updated, so it's recommended that you make a separate branch for any pull request your'e going to send so that it remains clean.  

Awaiting your changes, thanks!

gurbirkalsi:
Closing in favor of #1130

PR #1067, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "nvd3",
-  "version": "1.7.1",
   "homepage": "http://www.nvd3.org",
   "authors": [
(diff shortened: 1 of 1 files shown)


liquidpele:
pull requests should be to the development branch as that's the latest code, sorry.  We're going to make master our dev branch soon to avoid confusion like this :)

kkirsche:
This is just going to everyone who does this, so not reading any contrib rules. I'll send off dev

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

PR #1068, opened 2015-06-11
Add ability to change lwfc focus graph margins

portante:
If one would want to place an x-axis label below both the graph and its
focus graph, prior to this patch it is not possible in a fixed height
setting, as there is no way to change the bottom margin to account for
a label.

This change adds the ability to change the bottom margin.  Note that in
order for that label to work while maintaining the focus graph height,
one also has to up the "focusHeight" value.

### DIFF
--- a/src/models/lineWithFocusChart.js
+++ b/src/models/lineWithFocusChart.js
@@ -496,6 +496,12 @@ nv.models.lineWithFocusChart = function() {
             margin.bottom = _.bottom !== undefined ? _.bottom : margin.bottom;
             margin.left   = _.left   !== undefined ? _.left   : margin.left;
         }},
+        focusMargin: {get: function(){return margin2;}, set: function(_){
+            margin2.top    = _.top    !== undefined ? _.top    : margin2.top;
+            margin2.right  = _.right  !== undefined ? _.right  : margin2.right;
+            margin2.bottom = _.bottom !== undefined ? _.bottom : margin2.bottom;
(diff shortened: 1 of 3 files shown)


rreinhardt9:
@portante thanks for this, I just ran into this very situation!

src/models/lineWithFocusChart.js
```diff
@@ -494,6 +494,12 @@ nv.models.lineWithFocusChart = function() {
             margin.bottom = _.bottom !== undefined ? _.bottom : margin.bottom;
             margin.left   = _.left   !== undefined ? _.left   : margin.left;
         }},
+        margin2: {get: function(){return margin2;}, set: function(_){
```
liquidpele: I think this should probably be "focusMargin" to match the "focusHeight" option naming convention...

portante: That makes sense.  I'll make that change.

portante:
Closing submitted a new pull request, #1080, with the update.

portante:
Closed the wrong one.

liquidpele:
Looks like you picked up a bunch of other commits... need to "git rebase origin/master" :)

portante:
Okay rebased, sorry, missed that step.

liquidpele:
oh, nm, I updated dev branch to fix that.

liquidpele:
or maybe you fixed it then, lol.

portante:
I just did a quick rebase and that seems to have removed that last merge commit from this pr.

Let me know if there are any other problems!

PR #1069, opened 2015-06-11
Turn off scatter pts in focus chart for lineWFC

portante:
A line-with-focus-chart will essentially duplicate the number of lines
and points drawn on the screen because there are two copies of an
nv.models.line maintained.  We only need the scatter points for the main
chart, the focus chart will never use them.  This can save a lot of
memory and CPU when rendering graphs with large data sets.


liquidpele:
Good call... please do this for the line charts in the linePlusBarChart.js file too, since it has a focus chart as well.

portante:
Will do.

portante:
Closing, submitted a new pull request, #1080, with the update.

PR #1073, opened 2015-06-15
Fix  to  x-axis  title  to  be in the  middle  when there  is  only o…

kety33:
…ne/two bars

### DIFF
--- a/build/nv.d3.js
+++ b/build/nv.d3.js
@@ -1676,13 +1676,14 @@ nv.models.axis = function() {
                             .style('text-anchor', rotateLabels%360 > 0 ? 'start' : 'end');
                     }
                     axisLabel.enter().append('text').attr('class', 'nv-axislabel');
-                    if (scale.range().length < 2) {
-                        w = 0;
+                    w = 0;
+                    if (scale.range().length === 1) {
+                       w = scale.range()[0] * 2 + scale.rangeBand();
                     } else if (scale.range().length === 2) {
-                        w = scale.range()[1];
-                    } else {
-                        w = scale.range()[scale.range().length-1]+(scale.range()[1]-scale.range()[0]);
-                    }
+                       w = scale.range()[0] + scale.range()[1] + scale.rangeBand();;
+                    } else if ( scale.range().length > 2){
+                      w = scale.range()[scale.range().length-1]+(scale.range()[1]-scale.range()[0]);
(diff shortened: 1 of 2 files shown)


liquidpele:
Unit tests failed.... Also don't check in anything in the build folder.

robinfhu:
@kety33 can you see why the line `scale.rangeBand();` is erroring out? Thanks!

rreinhardt9:
@kety33 Looks like it might have been failing while testing a line chart; I don't think `.rangeBand()` will work except with an ordinal scale. So you may need to add logic that is conditional on whether the scale is ordinal or not. Looks like there is an `isOrdinal` attribute that is checked against around line 157 in axis.js, for example.

kety33:
Hi All,
Sorry  for the “stupid  question“ , but  how  can I run the  tests? This is he  first  time I do check in to github.

I  run ‘grunt production’ from nvd3 folder 
Than go  to  nvd3/test/node folder  and  run:
Npm install 
Grunt
And  didn’t  get  any  errors:
 gruntnvd3

I change  the  code  as Ross  recommendation , but  I  don't  know  if this  fix  the  tests:
w = 0;
                    if (scale.range().length === 1) {
                        w =  isOrdinal ? scale.range()[0] * 2 + scale.rangeBand() : 0;
                    } else if (scale.range().length === 2) {
                        w =  isOrdinal ? scale.range()[0] + scale.range()[1] + scale.rangeBand() : scale.range()[1];
                    } else if ( scale.range().length > 2){
                      w = scale.range()[scale.range().length-1]+(scale.range()[1]-scale.range()[0]);
                    };

rreinhardt9:
To run all the tests, you should be able to go to the projects root directory (whatever directory you cloned the project into, not in the test directory) and run `grunt` there. For example, I run 
```bash
~/nvd3 $ grunt
```
That runs all the tests and build the non minified versions so you can check out the examples for bugs.
This is assuming you have node and bower all set up (if not, see [here](https://github.com/novus/nvd3#tips-for-testing)).

liquidpele:
Hey there @kety33, I can't merge until you remove the build/nv.d3.js from the commit... merging in changes to those files and then overwriting them makes really annoying conflict resolutions for us.   

Other than that it looks good, if you can remove that from the pull request then I'll go ahead and merge this.

Issue #1041, opened 2015-05-15
Tooltip doesn't work properly when using multiple charts


kartikb:
When I try to create multiple charts on a single web page and use the option useInteractiveGuideline(true) the tooltips don't show up correctly. They are all placed with respect to the first chart.

liquidpele:
Use the development branch and see if that resolves your issue.  It has refactored tooltips.

siliconalchemy:
Hi, I have the same problem - use multiple charts on a web page to display metrics with useInteractiveGuideline, and the tooltips all show relative not to the first chart but they only go down as far as the bottom of the page as it's rendered.  In other words if the page scrolls down with more charts past the initial scroll point, those tooltips all display at the point of the bottom of the screen before it's scrolled.

liquidpele:
Please try the development branch or the v1.8.1-alpha tag, thanks.

fnoopdogg:
Hi, this is from the development branch as of this morning.

siliconalchemy:
Hi, this is from the development branch as of this morning.

liquidpele:
Oh that's strange, I tested that at one point... Can you make a jsfiddle that demonstrates the issue?

siliconalchemy:
Sure, I'll give it a go.  It's currently part of a large complex html template, so it might take me a while to isolate it.

liquidpele:
Thanks!

bwarren2:
I have hit this error as well, and replicated it in a [fiddle](https://jsfiddle.net/ye8Lbqqm/8/).

kartikb:
@bwarren2 It seems fixed in your fiddle though..

bwarren2:
When I mouseover the top 3 points in the second chart, I get no tooltip.  [This](http://imgur.com/a/I0eS2) is me mousing over the first chart, then a top point in the second chart.  Can you confirm what behavior you see?

I am using Firefox 38.0 on Ubuntu if it matters.

liquidpele:
Seems to work for me... what browser/OS are you using?

liquidpele:
Oh wait, you said that, doh

kartikb:
Verified again it works fine.. I am on Chrome Version 43.0.2357.81 (64-bit)

bwarren2:
[...]

PR #1079, opened 2015-06-17
Add chart id to clip path id attribute to avoid collision between mul…

liquidpele:
…tiple charts on a page, fixes https://github.com/novus/nvd3/issues/1041


liquidpele:
Small change, so I just merged it myself :)

PR #1082, opened 2015-06-17
Improvements to the time series example

heliodor:
Fixed overlapping bars when there are gaps. Fixed the first and last bar to not get cut off. Improved gap generator. Two styles of bars: x ticks centered on the bars vs x ticks at the bar edges. They are meant to differentiate between the meaning of the bars: bars that represent a point in time versus bars that represent a span of time.


liquidpele:

PR #1083, opened 2015-06-18
 Fix to x-axis title to be in the middle when there is only one/two bars  

kety33:


liquidpele:

PR #1094, opened 2015-06-24
Added 'TOTAL' to stackedAreaChart tooltip.

jhorowitz-coursera:
Stacked area plots are used when you want to understand how various parts add up to a whole, and how this breakdown varies with respect to some *x* dimension. So you often want to know what the whole is. This change adds a **TOTAL** row at the bottom of the `stackedAreaChart` tooltip.

`examples/stackedAreaChart.html` works well and looks like:

screen shot 2015-06-24 at 2 25 31 pm

### DIFF
--- a/src/models/stackedAreaChart.js
+++ b/src/models/stackedAreaChart.js
@@ -342,7 +342,7 @@ nv.models.stackedAreaChart = function() {
 
             interactiveLayer.dispatch.on('elementMousemove', function(e) {
                 stacked.clearHighlights();
-                var singlePoint, pointIndex, pointXLocation, allData = [];
+                var singlePoint, pointIndex, pointXLocation, allData = [], valueSum = 0;
                 data
                     .filter(function(series, i) {
                         series.seriesIndex = i;
@@ -367,6 +367,8 @@ nv.models.stackedAreaChart = function() {
                             color: color(series,series.seriesIndex),
                             stackedValue: point.display
                         });
+
+                        valueSum += tooltipValue;
                     });
 
                 allData.reverse();
@@ -390,6 +392,15 @@ nv.models.stackedAreaChart = function() {
                     });
                     if (indexToHighlight != null)
                         allData[indexToHighlight].highlight = true;
+                }
+
+                //If we are not in 'expand' mode, add a 'Total' row to the tooltip.
+                if (stacked.style() != 'expand' && allData.length >= 2) {
+                    allData.push({
+                        key: "TOTAL",
+                        value: valueSum,
+                        total: true
+                    });
                 }
 
                 var xValue = xAxis.tickFormat()(chart.x()(singlePoint,pointIndex));
--- a/src/models/stackedAreaChart.js
+++ b/src/models/stackedAreaChart.js
@@ -25,6 +25,8 @@ nv.models.stackedAreaChart = function() {
         , showYAxis = true
         , rightAlignYAxis = false
         , useInteractiveGuideline = false
+        , showTotalInTooltip = true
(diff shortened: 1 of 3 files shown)


portante:
@jhorowitz-coursera, nice, looks good.  If one toggles between "expand" and a non-expanded mode, does the TOTAL appear and disappear properly?

joshuahhh:
@portante, that's right -- the total row appears for "Stacked" and "Stream" modes, but not "Expanded".

jhorowitz-coursera:
@portante, that's right -- the total row appears for "Stacked" and "Stream" modes, but not "Expanded".

src/tooltip.js
```diff
@@ -109,7 +109,8 @@
 
             trowEnter.append("td")
                 .classed("key",true)
-                .html(function(p, i) {return keyFormatter(p.key, i)});
+                .html(function(p, i) {return keyFormatter(p.key, i)})
+                .style("font-weight", function(p) {return p.total && "bold"});
```
liquidpele: For something like this, it's better to have a class name and specify the syle elsewhere I think.

src/models/stackedAreaChart.js
```diff
@@ -392,6 +394,15 @@ nv.models.stackedAreaChart = function() {
                         allData[indexToHighlight].highlight = true;
                 }
 
+                //If we are not in 'expand' mode, add a 'Total' row to the tooltip.
+                if (stacked.style() != 'expand' && allData.length >= 2) {
+                    allData.push({
+                        key: "TOTAL",
```
liquidpele: There is no way to change the "TOTAL" string here?  What about for other languages?

src/models/stackedAreaChart.js
```diff
@@ -367,6 +367,8 @@ nv.models.stackedAreaChart = function() {
                             color: color(series,series.seriesIndex),
                             stackedValue: point.display
                         });
+
+                        valueSum += tooltipValue;
```
liquidpele: Is tooltipValue guaranteed to be a number here?

liquidpele:
This also needs an option to turn off the total thing for people who don't want it I think...

src/tooltip.js
joshuahhh: Done!

src/models/stackedAreaChart.js
joshuahhh: Added!

src/models/stackedAreaChart.js
joshuahhh: Hmm well it's computed with
```
var tooltipValue = (stacked.style() == 'expand') ? point.display.y : chart.y()(point,pointIndex);
```
So if we're not in expanded mode, I'm pretty confident we're getting a number. Not sure about expanded mode, but it doesn't matter, since we don't actually add a TOTAL row in that case. To make sure I'm not doing anything silly like concatenating a bunch of strings, I just added an `if` to protect this.

joshuahhh:
@liquidpele I believe I've addressed your comments. Let me know if you have any more thoughts!

liquidpele:

I went ahead and merged this, but if you have time, can you add documentation for the new options in the big json object in examples/documentation.html   ?     Thanks!

DmitryEfimenko:
I know this is closed, but I'd say it should be a requirement for Merges to also have [documentation](http://nvd3-community.github.io/nvd3/examples/documentation.html) updated when necessary (like in this case, when a new property `showTotalInTooltip` was introduced)

liquidpele:
yep, good point

PR #1104, opened 2015-06-29
Fix linePlusBarChart to better handle having multiple lines.

### LINKED ISSUE #463: linePlusBarChart legend behavior incorrectly when there is more than one line series
To reproduce the issue, just go to http://nvd3.org/livecode/index.html, and click line plus bar chart. In Data(JSON) tab, append new key "Price1" with the same values as "Price" key. Then in the chart, click/disable legend for "Price (right axis)", now both "Price" and "Price1" line disappear.

jakemcc:
Prior to these changes when you disabled the first line any other line
that was enabled also disappeared from the chart. Disabling any line
except for the first also resulted in the disabled line still appearing
on the chart but with the first line's color.

Fixes #463.

### DIFF
--- a/src/models/linePlusBarChart.js
+++ b/src/models/linePlusBarChart.js
@@ -114,6 +114,12 @@ nv.models.linePlusBarChart = function() {
         }
     };
 
+    var allDisabled = function(data) {
+      return data.every(function(series) {
+        return series.disabled;
+      });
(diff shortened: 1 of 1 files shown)


liquidpele:
Cool, this looks good, thanks for contributing!

PR #1112, opened 2015-06-30
Added option to switch lines and bars yAxis in linePlusBarChart

klangfarbe:
Hey,

this is a resubmit of https://github.com/nvd3-community/nvd3/pull/54 updated to the new nvd3 version. For details and example screenshots please see the previous pull request.

Best regards,
Chris

### DIFF
--- a/src/models/linePlusBarChart.js
+++ b/src/models/linePlusBarChart.js
@@ -47,6 +47,7 @@ nv.models.linePlusBarChart = function() {
         , defaultState = null
         , legendLeftAxisHint = ' (left axis)'
         , legendRightAxisHint = ' (right axis)'
+        , leftAxisShowBars = true
         ;
 
     lines.clipEdge(true);
@@ -67,6 +68,18 @@ nv.models.linePlusBarChart = function() {
     //============================================================
     // Private Variables
     //------------------------------------------------------------
+
+    var getBarsAxis = function() {
+        return leftAxisShowBars
+            ? { main: y1Axis, focus: y3Axis }
+            : { main: y2Axis, focus: y4Axis }
+    }
+
+    var getLinesAxis = function() {
+        return leftAxisShowBars
+            ? { main: y2Axis, focus: y4Axis }
+            : { main: y1Axis, focus: y3Axis }
+    }
 
     var stateGetter = function(data) {
         return function(){
@@ -137,13 +150,15 @@ nv.models.linePlusBarChart = function() {
 
             x = bars.xScale();
             x2 = x2Axis.scale();
-            y1 = bars.yScale();
-            y2 = lines.yScale();
-            y3 = bars2.yScale();
-            y4 = lines2.yScale();
+
+            // select the scales and series based on the position of the yAxis
+            y1 = leftAxisShowBars ? bars.yScale() : lines.yScale();
+            y2 = leftAxisShowBars ? lines.yScale() : bars.yScale();
+            y3 = leftAxisShowBars ? bars2.yScale() : lines2.yScale();
+            y4 = leftAxisShowBars ? lines2.yScale() : bars2.yScale();
 
             var series1 = data
-                .filter(function(d) { return !d.disabled && d.bar })
+                .filter(function(d) { return !d.disabled && (leftAxisShowBars ? d.bar : !d.bar) })
                 .map(function(d) {
                     return d.values.map(function(d,i) {
                         return { x: getX(d,i), y: getY(d,i) }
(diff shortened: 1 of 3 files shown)


liquidpele:
welp... I just merged another fix for the linePlusBarChart, and was about to look at yours, and apparently it caused some conflicts and github can't auto-merge now, sorry!

Can you please rebase with latest master and fix any conflicts?

klangfarbe:
No problem, I rebased the branch and pushed it.

src/models/linePlusBarChart.js
```diff
@@ -564,6 +590,7 @@ nv.models.linePlusBarChart = function() {
         focusShowAxisY:    {get: function(){return focusShowAxisY;}, set: function(_){focusShowAxisY=_;}},
         legendLeftAxisHint:    {get: function(){return legendLeftAxisHint;}, set: function(_){legendLeftAxisHint=_;}},
         legendRightAxisHint:    {get: function(){return legendRightAxisHint;}, set: function(_){legendRightAxisHint=_;}},
+        leftAxisShowBars:    {get: function(){return leftAxisShowBars;}, set: function(_){leftAxisShowBars=_;}},
```
liquidpele: I'm confused about why this option is necessary... the chart currently looks at if the series has an attribute named "bars" that is true, shouldn't that control whether the data is displayed as a bar or not?

liquidpele:
Also, any new options should be documented in the big json object at the bottom of examples/documentation.html so that they have proper information on the docs page, thanks!

klangfarbe:
Rebased to last version and renamed and documented the new option flag. 

The reason to add this flag is to let the user decide the order of the Y axis labels. Currently the bar labels are always on the left Y axis while the line labels are on the right. This might not be what the user want to display in some cases, especially in conjunction with other graphs where the left axis is always the same scale. Therefore this option is a way to create consistency between different graphs. For example a dashboard from an actual live system contains multiple graphs where the left axis should always be the same scale (percent values) as a user requirement. Enabling this new option flag in the linePlusBarChart allows this.

Default behaviour (_switchYAxisOrder_ flag is set to _false_):
flag is false

Switching the order (_switchYAxisOrder_ flag is set to _true_):
flag is true

src/models/linePlusBarChart.js
```diff
@@ -600,6 +627,13 @@ nv.models.linePlusBarChart = function() {
         }}
     });
 
+    chart.toggleAxis = function() {
+        switchYAxisOrder = !switchYAxisOrder;
+        var tickFormat = y1Axis.tickFormat();
+        y1Axis.tickFormat(y2Axis.tickFormat());
+        y2Axis.tickFormat(tickFormat);
+    }
+
```
liquidpele: why is this here?   The example html page should use the switchYAxisOrder option you added, not require this function on chart that doesn't match how any other options are done within nvd3. 

Other than that it looks good, I'll merge it in if you fix this.

klangfarbe: It is mostly a convenient function to make sure that the specified tickFormat for the bar and line Y-axis are corrected in case the user switches the axis. But I agree that adding this as a function is not the best way to do it. Currently I see a few options how to do it:

1. Do not set/swicht the tickFormat inside the chart at all and let the user do it manually.
2. Introduce two new variables ```barsTickFormat``` and ```linesTickFormat``` which stores the tickFormat and set them accordingly in the _y1_ or _y2_ axis, based on the option. If the user sets these variables we will use these or if he sets _y1axis.tickFormat()_ and/or _y2axis.tickFormat()_ (as you would normally do in nvd3) the new variables will be _null_ and can be initialized with _y1_ or _y2_ depending on the current state of the _switchYAxisOrder_ option.
2. Move the switching part to the setter function. But this would assume that the user sets the initial y1axis tickFormat for the bars and y2axis for the lines. 
```
    switchYAxisOrder:    {get: function(){return switchYAxisOrder;}, set: function(_){
        // Switch the tick format for the yAxis
        if(switchYAxisOrder !== _) {
            var tickFormat = y1Axis.tickFormat();
            y1Axis.tickFormat(y2Axis.tickFormat());
            y2Axis.tickFormat(tickFormat);
        }
        switchYAxisOrder=_;
    }}
```

What do you think about this?

liquidpele: I would do option 3, inside the setter...  that seems like the cleanest thing since that option drives the behavior.

klangfarbe: Ok, I added option 3 and rebased the code to the latest master

liquidpele:
Awesome, thanks!

PR #1116, opened 2015-07-01
Fixed wrong bar color changes in multiBar when a group is disabled

klangfarbe:
When a group is disabled in a multiBarChart (e.g. the first group in the multiBarChart2 example) the bar color for each group is changed but the color in the legend and tooltip is not. 

screenshot

For reference: The change was introduced in 0d7c737 

Previously the removed lines were in the 

    bars.
        .on('mouseover', function(d,i) {

lines but in my tests it does not make any difference if the lines are removed or moved back to the old position. Therefore I decided to remove them.


liquidpele:
I think I'm reverting that whole PR, there are other bugs there too....

liquidpele:
https://github.com/novus/nvd3/pull/1133

Issue #1093, opened 2015-06-24
Cannot Area click toggle on stacked area chart (1.8.1)


wayoutmind:
When updating from 1.7.1 to 1.8.1, I am now encountering the following exception when attempting to click on an area.

```
 Uncaught TypeError: Cannot read property 'y' of undefined
```

This issue exists when running the stacked area example:
https://github.com/novus/nvd3/blob/master/examples/stackedAreaChart.html

wayoutmind:
Any thoughts?

liquidpele:
Sorry, haven't had a chance to look at it yet...  If you want to look at it before I can and send up a fix via pull request then feel free :)

jlwerenskjold:
Hey! I've been working with angular-nvd3 and originally thought that it was the problem with clicking an area of a stacked graph not working but nvd3 is the culprit. To summarize what I posted [here](https://github.com/krispo/angular-nvd3/issues/169), the error that is being thrown is ````TypeError: d.display is undefined```` and is pointed at line 12190 in nvd3: ````.y(function(d) { return d.display.y + d.display.y0 })````

This function expects d to be a point object (I think, not sure what to call it) and is defined on line 12260 of nvd3:
 ````
.on('click', function(d,i) {
                    d3.select(this).classed('hover', false);
                    dispatch.areaClick({
                        point: d,
                        series: d.key,
                        pos: [d3.event.pageX, d3.event.pageY],
                        seriesIndex: d.seriesIndex
                    });
````
Now, for some reason d will sometimes be defined as the bottom graph object of the stacked area graph. For example, in the stacked area graph example [here](https://github.com/novus/nvd3/blob/master/examples/stackedAreaChart.html) d will sometimes be defined as the object:
````Object {key="Consumer Discretionary", values=[77],...```` no matter which area is clicked. In my own graphs the bottom most area on the graph is sometimes defined as d. This is what is causing the error to be thrown. Why this happens I am not sure.

 I tried putting a check in the scatter.y function and that fixed the problem. My scatter.y function is simply:
````
.y(function(d) {
            if(typeof d.display === 'undefined') {}
            else {
              return d.display.y + d.display.y0
            }
          })
````
For the moment this fixes the issue, although getting down to the root of the problem and fixing that should be done as this is just a rudimentary fix.

wayoutmind:
[...]

PR #1118, opened 2015-07-02
Fixed the typeError when selecting an area of a stacked area graph.

jlwerenskjold:
I just added a simple check in the scatter.y function to ignore all d objects that are not point objects. This is the fix for issue #1193.

### DIFF
--- a/build/nv.d3.js
+++ b/build/nv.d3.js
@@ -12324,7 +12324,10 @@ nv.models.stackedArea = function() {
                 .width(availableWidth)
                 .height(availableHeight)
                 .x(getX)
-                .y(function(d) { return d.display.y + d.display.y0 })
+                .y(function(d) {
+                  if( typeof d.display === 'undefined' ) {}
+                  else { return d.display.y + d.display.y0 }
+                })
                 .forceY([0])
                 .color(data.map(function(d,i) {
                     return d.color || color(d, d.seriesIndex);


build/nv.d3.js
```diff
@@ -12324,7 +12324,10 @@ nv.models.stackedArea = function() {
                 .width(availableWidth)
                 .height(availableHeight)
                 .x(getX)
-                .y(function(d) { return d.display.y + d.display.y0 })
+                .y(function(d) {
+                  if( typeof d.display === 'undefined' ) {}
+                  else { return d.display.y + d.display.y0 }
```
liquidpele: That's an odd way to do it... should probably be this: 

    if (d.display !== undefined) { 
        return d.display.y + d.display.y0;
    }

liquidpele:
Also can you link to the exact issue this is for?    I don't see an issue 1193, you can copy/paste the whole issue url here and github will convert it to just a shorthand link.

build/nv.d3.js
jlwerenskjold: Ah yes, that's much better, you might need the "typeof" in front of d.display but I'm not sure. The issue was https://github.com/novus/nvd3/issues/1093 not #1193, as you can see I was quite tired when I did this...

liquidpele: > you might need the "typeof" in front of d.display but I'm not sure. 

na, you can just compare the the literal undefined, instead of checking if typeof returns a string of 'undefined'...   a bit cleaner that way imho :)

liquidpele:
This is a fix for https://github.com/novus/nvd3/issues/1093

liquidpele:
Oh....   and you changed the build file instead of the source file...   can't do that, it'll just get overwritten later.   You need to change the source file.   Nothing in the build directory should ever be touched in a pull request.   Closing this, please re-submit a new PR that changes the source file.   Seems to be line 113 in stackedArea.js

PR #1122, opened 2015-07-04
Fix proposal for #1121

### LINKED ISSUE #1121: MultiBarChart stacked shows 1px bar height if value == 0
nvd3

devilicecream:
Attempting to fix #1121. This fixes the issue in my case, not sure about side effects of this fix.

### DIFF
--- a/src/models/multiBarChart.js
+++ b/src/models/multiBarChart.js
@@ -334,12 +334,14 @@ nv.models.multiBarChart = function() {
 
     multibar.dispatch.on('elementMouseover.tooltip', function(evt) {
         evt.value = chart.x()(evt.data);
-        evt['series'] = {
-            key: evt.data.key,
-            value: chart.y()(evt.data),
-            color: evt.color
-        };
-        tooltip.data(evt).hidden(false);
+        if (evt.data.size > 0 || evt.data.size == undefined) {
+            evt['series'] = {
+                key: evt.data.key,
+                value: chart.y()(evt.data),
+                color: evt.color
+            };
+            tooltip.data(evt).hidden(false);
+        }
     });
 
     multibar.dispatch.on('elementMouseout.tooltip', function(evt) {
--- a/src/models/multiBarChart.js
+++ b/src/models/multiBarChart.js
@@ -334,7 +334,7 @@ nv.models.multiBarChart = function() {
 
     multibar.dispatch.on('elementMouseover.tooltip', function(evt) {
         evt.value = chart.x()(evt.data);
-        if (!chart.stacked() || (evt.data.y > 0 || evt.data.y == undefined)) {
+        if (!chart.stacked() || chart.y()(evt.data) > 0) {
             evt['series'] = {
                 key: evt.data.key,
                 value: chart.y()(evt.data),
--- a/src/models/multiBarChart.js
+++ b/src/models/multiBarChart.js
@@ -334,7 +334,7 @@ nv.models.multiBarChart = function() {
 
     multibar.dispatch.on('elementMouseover.tooltip', function(evt) {
         evt.value = chart.x()(evt.data);
-        if (!chart.stacked() || (evt.data.size > 0 || evt.data.size == undefined)) {
+        if (!chart.stacked() || (evt.data.y > 0 || evt.data.y == undefined)) {
             evt['series'] = {
                 key: evt.data.key,
                 value: chart.y()(evt.data),
--- a/src/models/multiBarChart.js
+++ b/src/models/multiBarChart.js
@@ -334,7 +334,7 @@ nv.models.multiBarChart = function() {
 
     multibar.dispatch.on('elementMouseover.tooltip', function(evt) {
         evt.value = chart.x()(evt.data);
-        if (evt.data.size > 0 || evt.data.size == undefined) {
+        if (!chart.stacked() || (evt.data.size > 0 || evt.data.size == undefined)) {
             evt['series'] = {
                 key: evt.data.key,
                 value: chart.y()(evt.data),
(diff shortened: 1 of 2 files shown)


liquidpele:
I think that might be by design...   @robinhfu ?

robinfhu:
Pretty sure that's intentional, in a way to indicate an empty value.

robinfhu:
One alternative idea, is keep the one pixel rect, but give it the right series color

devilicecream:
@robinfhu @liquidpele if it is by design, why every other column doesn't show two 1px rect? The data in the example is like: {serie_1: {day1: 1, day2: 0, day3: 0 ...}, serie_2: {day1:0, day2: 0, day3: 0}}
To me it doesn't seem like a coherent behavior.
@robinfhu I have no way to predict if the serie with the green color is gonna be 0 or the serie with red color is gonna be 0. Plus, there could be situations in which there is a day with 0 in the red series and 1 in the green series and also a day with 1 in the red series and 1 in the green series. 

The problem is only with stacked, in the grouped layout I understand that the 1 pixel height is a way to indicate an empty value, and the behavior is coherent in every column.

liquidpele:
Thinking about this more, I think it may be more correct to have it actually be a height of zero...   they wouldn't be able to use the tooltip on it, but it would be annoying anyway and the interactive guideline should still work I think...

liquidpele:
@devilicecream can you provide a jsfiddle showing your changes?  If the tooltips all behave correctly then I'd be willing to merge this.

devilicecream:
@liquidpele here the jsfiddle with the latest fix to avoid showing the tooltip if the y value is 0.
http://jsfiddle.net/ptce0078/5/

src/models/multiBarChart.js
```diff
@@ -334,12 +334,14 @@ nv.models.multiBarChart = function() {
 
     multibar.dispatch.on('elementMouseover.tooltip', function(evt) {
         evt.value = chart.x()(evt.data);
-        evt['series'] = {
-            key: evt.data.key,
-            value: chart.y()(evt.data),
-            color: evt.color
-        };
-        tooltip.data(evt).hidden(false);
+        if (evt.data.size > 0 || evt.data.size == undefined) {
```
liquidpele: Can you explain what this is doing here?    What is the size attribute supposed to represent?   (I haven't looked at that logic before sorry)

src/models/multiBarChart.js
```diff
@@ -334,12 +334,14 @@ nv.models.multiBarChart = function() {
 
     multibar.dispatch.on('elementMouseover.tooltip', function(evt) {
         evt.value = chart.x()(evt.data);
-        evt['series'] = {
-            key: evt.data.key,
-            value: chart.y()(evt.data),
-            color: evt.color
-        };
-        tooltip.data(evt).hidden(false);
+        if (!chart.stacked() || (evt.data.size > 0 || evt.data.size == undefined)) {
```
liquidpele: Can you explain why this is necessary?  Also, what does the size attribute represent there?

liquidpele:
I updated the jsfiddle to use a recent version of d3 and the nvd3 from master:  

https://jsfiddle.net/ptce0078/6/

src/models/multiBarChart.js
devilicecream: There are some cases where the tooltip is shown with y = 0 (`evt.data.size` is the y value of the column), even if the column is stacked onto another column with y > 0. The effect is not the best.
With this code I solve this problem hiding the tooltip when the graph is stacked and y < 0.
I'll try to reproduce the issue in the updated jsfiddle.

devilicecream:
Updated the jsfiddle to show the behavior I wrote in the previous comment. If point your cursor in the first column, on y = 3, you will see a tooltip saying `Good : 0`. Furthermore, you can see `Excellent : 0` by positioning your cursor on every empty column at y = 0. My last commit resolves this issue.

src/models/multiBarChart.js
liquidpele: I don't see where size is ever set in multiBar.js ... I did a console.log(evt.data) in there on my branch and it never has a size attribute.   I think you might be mistaken...

devilicecream: I'm pretty sure I have it. Look I'm now using the same d3.js version you put in the jsfiddle and nvd3 version 1.8.1, and I have the `size` in every multiBarChart (stacked).

liquidpele: I don't see how though... I don't see where it's set anywhere in multiBar.js, and that's what passes the data on the "mouseover" d3 event...   and I can't make it have a size attribute using the multiBar.html or multiBar2.html examples in the examples folder...

devilicecream: I can actually get it in the `multiBarChart.html` and the `multiBarChart2.html` examples. I guess the problem is only with some browsers. The `series` field actually disappears when I use Safari, but it's there when I'm using Chrome or Firefox. Anyway I always get the `evt.data.y` and we can use that one for the same purpose. I'll do another commit.

liquidpele: Well, you need to really do multibar.getY(evt.data) I think, you can't assume that everyone will use the y attribute name since they can override that using the getY function :)

devilicecream: Makes sense :+1: updating again :)

devilicecream:
I'm actually using the `chart.y(evt.data)` function to get the y value, as that one is used to get the value to show in the tooltip. Probably if I should use the overridable `getY()`, the same one should also be used to render the value to show in the tooltip. What do you think @liquidpele ?

liquidpele:
Na, your'e right, it should have been chart.y, merged and thanks for contributing!

devilicecream:
You're very welcome! Thanks for helping me with this issue :smile:

PR #1130, opened 2015-07-10
Added maxKeyLength variable to truncate legend key values.

gurbirkalsi:
This fix provides an additional variable to set a fixed max key length
value to truncate any legend key values with lengths over the specified
threshold. This provides the option to organize legend keys to be seen
in a more readable manner.

Added tooltip for truncated legend text.
When mousing over truncated text within the legend,
a tooltip showing the full key value is revealed.

### DIFF
--- a/src/models/legend.js
+++ b/src/models/legend.js
@@ -10,6 +10,7 @@ nv.models.legend = function() {
         , height = 20
         , getKey = function(d) { return d.key }
         , color = nv.utils.getColor()
+        , maxKeyLength = 20 //default value for key lengths
         , align = true
         , padding = 32 //define how much space between legend items. - recommend 32 for furious version
         , rightAlign = true
@@ -67,7 +68,6 @@ nv.models.legend = function() {
                     .attr('class','nv-legend-symbol')
                     .attr('rx', 3)
                     .attr('ry', 3);
-
                 seriesShape = series.select('.nv-legend-symbol');
 
                 seriesEnter.append('g')
@@ -177,7 +177,14 @@ nv.models.legend = function() {
 
                 var seriesWidths = [];
                 series.each(function(d,i) {
-                    var legendText = d3.select(this).select('text');
+                    var legendText;
+                    if (getKey(d).length > maxKeyLength) { 
+                        var trimmedKey = getKey(d).substring(0, maxKeyLength);
+                        legendText = d3.select(this).select('text').text(trimmedKey + "...");
+                        d3.select(this).append("svg:title").text(getKey(d));
+                    } else {
+                        legendText = d3.select(this).select('text');
+                    } 
                     var nodeTextLength;
                     try {
                         nodeTextLength = legendText.node().getComputedTextLength();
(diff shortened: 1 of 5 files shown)


src/tooltip.js
```diff
@@ -44,6 +44,9 @@
             ,   enabled = true  //True -> tooltips are rendered. False -> don't render tooltips.
             ,   duration = 100 // duration for tooltip movement
             ,   headerEnabled = true
+            ,   colorDivEnabled = true //enable color div in tooltip for legend elements
```
liquidpele: Not very fond of this name, what if we wanted to change it to a span or anchor tag in the future?   Also, it's not really enabling functionality, it's just showing them... so maybe name them like "showColors", "showKey", and "showValues" ?

src/tooltip.js
```diff
@@ -388,10 +396,14 @@
             enabled: {get: function(){return enabled;}, set: function(_){enabled=_;}},
             hideDelay: {get: function(){return hideDelay;}, set: function(_){hideDelay=_;}},
             contentGenerator: {get: function(){return contentGenerator;}, set: function(_){contentGenerator=_;}},
+            tooltipContent: {get: function(){return tooltipContent;}, set: function(_){tooltipContent=_;}},
```
liquidpele: What is this for?   It doesn't appear to reference an existing variable?

liquidpele:
Your stuff around the max key length looks fine... in the future please try to separate pull requests into features... this one is really 2 different things in it.

src/models/legend.js
```diff
@@ -90,13 +91,25 @@ nv.models.legend = function() {
...
                 .on('mouseover', function(d,i) {
                     dispatch.legendMouseover(d,i);  //TODO: Make consistent with other event objects
+                    if (getKey(d).length > maxKeyLength) {
+                        d['series'] = {
+                            key: getKey(d)
+                        };
+                        tooltip.duration(0).headerEnabled(false);
+                        tooltip.data(d).hidden(false);
+                        tooltip.data(d).colorDivEnabled(false);
+                        tooltip.data(d).valueEnabled(false);
```
liquidpele: Why not just override the contentGenerator if you're hiding everything but the key?

portante: @liquidpele, it seemed to me that the modifications to tooltip.js where smaller than adding our own content generator just for legend.  But if it is better to have a content generator that duplicates what tooltip is doing, and that would be more maintainable, we could certainly do this.

liquidpele: I think having it's own content generator might be easier, since it would be just a single function instead of a convoluted set of options...   @robinfhu do you have any thoughts on that?

robinfhu: Just realized, if the goal is to just show a tooltip with the full legend text, why not use a title element? http://stackoverflow.com/questions/10643426/how-to-add-a-tooltip-to-an-svg-graphic

portante: Good suggestion, wish I had thought of that!  We'll check this out and see how it goes.

portante:
@liquidpele, @robinfhu, this work looks pretty solid now.  Does anybody know of a reason why this should not be merged?

liquidpele:
Looks good, much cleaner that way I think!

dinatih:
Hello, nice feature ! 

How can I change the default value, I've tried : `chart.options({ maxKeyLength: 40 });` but it doesn't seem to work, please ?

liquidpele:
That option is on the legend object that exists on chart... so you'd do something like chart.legend.maxKeyLength(40) or chart.legend.options({maxKeyLength: 40})

PR #1131, opened 2015-07-13
Fix noData container placement

mrijke:
Always use 'null' to call available{Width,Height} when creating
the noData container so that the the proper width and height
is calculated from the container.
Otherwise the value from the chart options would be used, 
which might not have been overridden.


robinfhu:
Cool, looks good.

PR #1135, opened 2015-07-14
Fixed TypeError bug when a user clicks on an area of a stacked area chart, Issue #1093

jlwerenskjold:
Fix for Issue #1093, still get warnings when a user clicks on an area in the chart but no errors occur and all functionality is there.


liquidpele:
perfect, thanks!

Issue #1124, opened 2015-07-06
tooltip with 'w' gravity jumps down when it hits top of window


jmensch1:
I'm using the tooltip with the default 'w' gravity, and when the tooltip hits the top of the window, it jumps down below the mouse instead of hugging the top of the window.  I found the problem in the code that calculates the tooltip position (starting at line 664 in version 1.8.1 dated 2015-06-17) -- 

```
switch (gravity) {
    case 'e':
        left = pos[0] - width - distance;
        top = pos[1] - (height / 2);
        tLeft = tooltipLeft(tooltipElem);
        tTop = tooltipTop(tooltipElem);
        if (tLeft < scrollLeft) left = pos[0] + distance > scrollLeft ? pos[0] + distance : scrollLeft - tLeft + left;
        if (tTop < scrollTop) top = scrollTop - tTop + top;     // LINE A
        if (tTop + height > scrollTop + windowHeight) top = scrollTop + windowHeight - tTop + top - height;
        break;
    case 'w':
        left = pos[0] + distance;
        top = pos[1] - (height / 2);
        tLeft = tooltipLeft(tooltipElem);
        tTop = tooltipTop(tooltipElem);
        if (tLeft + width > windowWidth) left = pos[0] - width - distance;
        if (tTop < scrollTop) top = scrollTop + 5;       // LINE B
        if (tTop + height > scrollTop + windowHeight) top = scrollTop + windowHeight - tTop + top - height;
        break;
```

Notice that the formula for calculating top is different for 'e' and 'w' (compare Line A with Line B).  To fix the problem, I simply made Line B the same as Line A, so Line B is now:

``` 
if (tTop < scrollTop) top = scrollTop - tTop + top;  
```

This fixed the issue in all browsers.

Thanks for your great work on this library, my client is very happy with it.

liquidpele:
Hey there!   Can you send us a pull request with the fix?

jmensch1:
Done!  Although it's saying something about a travis build error

liquidpele:
You can ignore the travis build error... I broke the unit tests earlier today apparently :/

RenatoUtsch:
Should this be closed?

PR #1139, opened 2015-07-14
tooltip-fix

jmensch1:
This is a one-line fix to Issue #1124.  As I said there, the tooltip with 'w' gravity was jumping down when it hit the top of the window, instead of staying at the top.  (The problem did not exist when the tooltip hit the bottom of the window).  

If you look at the switch statement that calculates the 'top' for each of the four gravity types, it's pretty clear that the 'top' formula for 'w' was accidentally taken from the 'left' formula for 'n' and 's'.   My fix simply makes the 'top' formula for 'w' the same as the 'top' formula for 'e'.  

Before and after pics are attached.  The important points in the pics are the location of the mouse vs. the location of the tooltip.

Thanks much,
Jake

after
before

### DIFF
--- a/build/nv.d3.js
+++ b/build/nv.d3.js
@@ -678,7 +678,7 @@ nv.nearestValueIndex = function (values, searchVal, threshold) {
                         tLeft = tooltipLeft(tooltipElem);
                         tTop = tooltipTop(tooltipElem);
                         if (tLeft + width > windowWidth) left = pos[0] - width - distance;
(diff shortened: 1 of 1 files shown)


jmensch1:
Oops, the top pic is actually the after pic.

liquidpele:
You modified the build files... you need to modify the source files.   The build files should never be touched for pull requests... please re-submit a pull request that modifies only the source files (under the src directory), thanks!

PR #1140, opened 2015-07-14
initial commit (to src instead of build)

jmensch1:
Okay, I'm trying this again after fixing that build issue.

This is a one-line fix for Issue #1124. As I said there, the tooltip with 'w' gravity was jumping down when it hit the top of the window, instead of staying at the top. (The problem did not exist when the tooltip hit the bottom of the window).

If you look at the switch statement that calculates the 'top' for each of the four gravity types, it's pretty clear that the 'top' formula for 'w' was accidentally taken from the 'left' formula for 'n' and 's'. My fix simply makes the 'top' formula for 'w' the same as the 'top' formula for 'e'.

Before and after pics are attached. The important points in the pics are the location of the mouse vs. the location of the tooltip.

Thanks much,
Jake

Before: tooltip jumping down when it hits top of window
before

After: tooltip hugs top of window
after

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -206,7 +206,7 @@
                         tLeft = tooltipLeft(tooltipElem);
                         tTop = tooltipTop(tooltipElem);
                         if (tLeft + width > windowWidth) left = pos[0] - width - distance;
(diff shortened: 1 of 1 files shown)


liquidpele:
cool, thanks!

PR #1149, opened 2015-07-17
Update bower license to match proper license string identifiers e.g. …

liquidpele:
…from https://spdx.org/licenses/


robinfhu:
oh yeah, npm 2 is checking for that now.

PR #1151, opened 2015-07-17
scatter.highlightPoint is broken

### LINKED ISSUE #1147: scatter.highlightPoint is broken
Highlighting a point with the scatter model currently runs this code:

```
this.highlightPoint = function (seriesIndex, pointIndex, isHoverOver) {
            nv.dom.write(function() {
                container.select(" .nv-series-" + seriesIndex + " .nv-point-" + pointIndex)
                    .classed("hover", isHoverOver);
            });
        };
```

The problem here is container.select updates the data for the point with the data it holds. Once that happens, future operations will throw errors.

For an example click on a series in the stacked area chart example.

We need to find a way to get the point without altering the data on the way.

jgrund:
Fixes #1147

This patch fixes data rebinding when highlighting points in the scatter
model.

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -443,8 +443,10 @@ nv.models.scatter = function() {
         };
         this.highlightPoint = function (seriesIndex, pointIndex, isHoverOver) {
             nv.dom.write(function() {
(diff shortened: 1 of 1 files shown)


liquidpele:
Cool, tested and seems good, thanks!

PR #1158, opened 2015-07-21
adding legendEnabled to tooltip default contentGenerator

bkjelden:
pretty simple change - adding an option to toggle the legend on/off on a tooltip, similar to the already existing option for tooltip headers.


liquidpele:
This is already easily done with some CSS though... 

    .legend-color-guide { width: 1px; }
    .legend-color-guide > * {display: none}

bkjelden:
Just curious - what's the reason for the headerEnabled option, then? It seems like that could also be easily hidden through css.

liquidpele:
The pie charts don't utilize header since they have one less data point.

bkjelden:
Makes sense. Thanks for clarifying.

PR #1163, opened 2015-07-23
Documentation Fix for Default Margins

bkjelden:
The documentation says the default margin for all charts is 0 in all dimensions - which isn't correct.

This change instantiates the chart and pulls the default margin off of it, and then serializes that to a string to show as the documentation.

This pattern could probably be used for some other pieces of documentation, but I don't have an exhaustive enough knowledge of the library to know where it would be useful.

### DIFF
--- a/examples/documentation.html
+++ b/examples/documentation.html
@@ -177,10 +177,10 @@
 
                         var desc = option_info[name]['desc'];
                         var def;
-                        if (typeof option_info[name]['default'] == 'function') {
-                            // use function in option_info to pull default value of the option off an instance of the chart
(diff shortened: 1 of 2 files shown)


liquidpele:
huh... that's a clever idea...

liquidpele:
However, some options actually have functions for their defaults, so this would mess that up.   But this is a neat idea, it probably just needs to be a different attribute than "default", like call it "default_getter" or something and set that to the function and then use that if it's set instead of only relying on "default" for everything.

liquidpele:
Cool, thanks!

Issue #1164, opened 2015-07-23
the distribution position is wrong in scatterChart due to the scatter point position calculation changed


Alex--wu:
Since the commit in Mar 6, 2015 'scatter point position is calculated based on the overall page position of the container now...', the position of scatter calculation is changed. Before it is base on the container, now it is base on the overall page. But the calculation in scatterChart has not been updated. The problem is in line 317 of scatterChart.js

Alex--wu:
I can fix it by recalculate the POS in scatterChart. But there is noway now to know the position of scatter point based on the container. My opinion is keep both.

liquidpele:
Can you provide a jsfiddle showing the issue using the master branch code?   You can use rawgit.com to link to the build files in master.

Alex--wu:
https://jsfiddle.net/wushengyuan/qmxtvy1z/

liquidpele:
Ah, do you mean how the colored lines that appear on hover that are supposed to go from the point to each axis are off?   Yea I haven't had a chance to fix but I had seen it before...   if you already looked into it and know how to fix please send a pull request!

Alex--wu:
Yes that is what I mean! I will fix it soon today. Do you know what other chart is using the scatter position? Can you tell me why you change the position to base on the overall page?

liquidpele:
Okay, so this was also an issue brought up in https://github.com/novus/nvd3/issues/971 so I closed that old one and we'll track this particular issue here.   

The position is sent up the hover events to the tooltip position code, so the tooltips expect position to be based on the page.  That was done because the browsers make it a lot easier to position things that way... they account for scroll and all kinds of stuff for us, where calculating it ourselves based on relative position was a nightmare and we had like 10 different tooltip position bugs previously.  

Thus if you change the position for those, don't change what position is passed up to the tooltips :)

Alex--wu:
okay so I will keep both position(base on page and base on container). I
keep the current key 'pos' for the position base on the page. And I will
create the key 'posObj' for the position base on the container. Does it
make sense for you? Do you have better key name for it?

2015-07-24 14:40 GMT+02:00 liquidpele <notifications@github.com>:

liquidpele:
relativePos maybe?

Alex--wu:
y

Alex--wu:
hi liquidpele,

When I am checking why my build test failed, I have a question for you about the position about the mouse even in scatter.js.
[...]

PR #1170, opened 2015-07-27
Bugfix/scatter

Alex--wu:
1. Fix the bug #1164 
2. Add new feature for the lineChart: add distribution lines like how the scatterChart showing.

### DIFF
--- a/src/models/lineChart.js
+++ b/src/models/lineChart.js
@@ -355,23 +355,23 @@ nv.models.lineChart = function() {
                 tooltip.data(evt).position(evt.pos).hidden(false);
 
                 if(showDistX){
-                    container.select('.nv-series-' + evt.seriesIndex + ' .nv-distx-' + evt.pointIndex)
-                        .attr('y1', evt.pos.top - availableHeight - margin.top);
+                    container.select('.nv-distributionX .nv-series-' + evt.seriesIndex + ' .nv-distx-' + evt.pointIndex)
+                        .attr('y1', evt.relativePos[1] - availableHeight);
                 }
                 if(showDistY){
-                    container.select('.nv-series-' + evt.seriesIndex + ' .nv-disty-' + evt.pointIndex)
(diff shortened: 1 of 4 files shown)


src/models/lineChart.js
```diff
@@ -306,6 +351,30 @@ nv.models.lineChart = function() {
...
+                tooltip.hidden(true);
+                if(showDistX){
+                    container.select('.nv-distributionX .nv-series-' + evt.seriesIndex + ' .nv-distx-' + evt.pointIndex)
+                        .attr('y1', 0);
+                }
+                if(showDistY){
+                    container.select('.nv-distributionY .nv-series-' + evt.seriesIndex + ' .nv-disty-' + evt.pointIndex)
+                        .attr('x2', distY.size());
+                }
+            });
```
liquidpele: Why are you moving these functions here?   They'll get overwritten every chart.update() in this case.

liquidpele:
Hmm, well the unit tests failed, and it looks like you're trying to add in a feature with the bug fix... please submit 2 different pull requests, it's hard to judge code with stuff all put into a single pull request like that.  For instance, the bug fix looks good but I'm not sure I even like the idea of the distribution lines on the line chart... I'm not sure why I left them on the scatter chart honestly :p

src/models/lineChart.js
Alex--wu: I just make it like the scatterChart. You are right , I will move it
outside the chart scope. Also I will move the one in scatterChart outside
the scope too. And I will do some optimization on the scatterChart.

2015-07-27 19:18 GMT+02:00 liquidpele <notifications@github.com>:

Alex--wu: I will do that tomorrow. Too busy these days.

2015-07-27 19:30 GMT+02:00 Shengyuan Wu <shengyuan.w@gmail.com>:

Issue #171, opened 2013-06-27
StackedAreaChart Problems: about date values, hide controls, and size too small


hustcer:
There are a few problems in stackedAreaChart:
1. The x axis dates values are unix timestamp, which is quite inconvenient and not intuitive for us. Currently, If we don't use unix time stamp the tooltips won't show up.
2. When we hide the controls by ".showControls(false)", the legends won't flow to the left of the chart, especially when the svg element has a smaller width and height, the legends just aligned vertically, which will make the chart much lower.
3. Check out the https://github.com/novus/nvd3/blob/master/examples/stackedAreaChart.html example, if we add ".width(300).height(250)" to the chart settings, and open it in the browser the chart will be messed up. Sometimes we use it as a small widget but if the display area is too small, then the chart was broken.

liquidpele:
Really old and the major issue was fixed, closing it.  Please open any still existing issues based on the latest version, thanks!

PR #1171, opened 2015-07-28
Fix #1164: scatter mouseover wrong position.

### LINKED ISSUE #1164: the distribution position is wrong in scatterChart due to the scatter point position calculation changed

Alex--wu:
Also for #171.

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -263,6 +263,7 @@ nv.models.scatter = function() {
                             point: point,
                             series: series,
                             pos: pos,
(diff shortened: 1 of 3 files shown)


Alex--wu:
@liquidpele Any problem with this pull request?

liquidpele:
ah, sorry.  Yup, looks fine.

PR #1172, opened 2015-07-29
Fixed tooltip transform style interpolation issue under Safari

lhethert:
Added extra tween for the -webkit-transform style attribute, so that in Safari (and Qt5.3+ applications using Qt webkit), the tooltips do not appear to fly in from the origin. This does not appear to impact other browsers in any way. 

Could not find an existing test for this, so please let me know if one needs to be added and where.

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -273,7 +273,7 @@
                         // Safari has its own `-webkit-transform` and does not support `transform`
                         .styleTween('-webkit-transform', function (d) {
                             return translateInterpolator;
-                        }, 'important')
+                        })
                         .style('-ms-transform', new_translate)
                         .style('opacity', 1);
                 }
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -271,8 +271,9 @@
                             return translateInterpolator;
                         }, 'important')
                         // Safari has its own `-webkit-transform` and does not support `transform`
-                        // transform tooltip without transition only in Safari
-                        .style('-webkit-transform', new_translate)
+                        .styleTween('-webkit-transform', function (d) {
+                            return translateInterpolator;
+                        }, 'important')
                         .style('-ms-transform', new_translate)
                         .style('opacity', 1);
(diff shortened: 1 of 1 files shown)


liquidpele:
I'm confused, why doesn't the normal "transform" work in safari?   

http://caniuse.com/#feat=transforms2d

Also, even if this was necessary, it shouldn't use the 'important' since that's reserved for the normal transform to keep the two attributes from conflicting if a browser supports both.

lhethert:
I agree that 'transform' should work, but I have no idea why it doesn't. Having said that I have also tested this using a Qt5.3-based browser application (which uses QtWebkit), and it suffers from the same issue. Agree with you re: the 'important' tag, that was an oversight on my part - I will update this.

lhethert:
Just noting that work based on previous feedback is complete. Now waiting on further review / comment.

cymen:
I'm seeing this too. What is holding up getting this in?

cymen:
Or did this change get in another way? Source looks different but still broken in Safari.

lhethert:
Seems to work fine for me. Which example are you looking at?

cymen:
@lhethert I'm looking at our local codebase that makes use of `chart.tooltip.contentGenerator`. On Safari, the tooltips transform in from the upper left of the screen for each point which is odd. Trying to figure out if it is something in our local code.

@lhethert by work fine, do you mean with this PR or master?

cymen:
This example exhibits the issue when viewed with Safari:

http://nvd3-community.github.io/nvd3/examples/site.html

cymen:
Crap, can't actually link -- the ones that exhibit the issue on that page are:

* LinePlusBarChart
* MultiChart
* ScatterChart

liquidpele:
Merged, thanks!

lhethert:
Great! @cymen, apologies for confusion, when I was saying it "works for me" I was meaning this PR, not master.

PR #1176, opened 2015-07-30
Fix when tooltip gravity is 'center', the container is undefined.

Alex--wu:


Alex--wu:
@liquidpele and this one

liquidpele:
looks good!

PR #1182, opened 2015-08-04
Fix bug with the candlestick chart.

RenatoUtsch:
If the data changed the candlestick would plot the candles in the wrong place. Now this is fixed.

### DIFF
--- a/src/models/candlestickBar.js
+++ b/src/models/candlestickBar.js
@@ -107,8 +107,10 @@ nv.models.candlestickBar = function() {
                 .data(function(d) { return d });
             ticks.exit().remove();
 
+            var tickGroups = ticks.enter().append('g');
+
             // The colors are currently controlled by CSS.
-            var tickGroups = ticks.enter().append('g')
+            ticks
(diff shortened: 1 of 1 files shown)


src/models/candlestickBar.js
```diff
@@ -135,14 +135,17 @@ nv.models.candlestickBar = function() {
                     return open > close ? y(close) - y(open) : y(open) - y(close);
                 });
 
-            container.selectAll('.nv-candlestick-lines').transition()
+            ticks
+                .attr('class', function(d, i, j) { return (getOpen(d, i) > getClose(d, i) ? 'nv-tick negative' : 'nv-tick positive') + ' nv-tick-' + j + '-' + i});
+
```
liquidpele: Instead of overwritting the whole 'class' attribute you should use classed:

https://github.com/mbostock/d3/wiki/Selections#classed

Otherwise it's easy to accidentally remove stuff that it shouldn't.

RenatoUtsch:
I actually want to replace the class, because it is the class that decides if the candle is positive or negative. I just added it to the right place now, when adding or updating the candles.

liquidpele:
That works too, looks good now.

PR #1190, opened 2015-08-07
Refactor tooltip code

### LINKED ISSUE #1138: Critical bug with the guidelined tooltip when scrolling
If the element that does the scrolling is not the body, the tooltip gets all buggy:
http://jsfiddle.net/RenatoUtsch/h4rvoehd/2/

Also, this only happens with the interactive guideline. Without it the tooltip is positioned correctly (as in the example).

RenatoUtsch:
This commit is an entire refactor of the tooltip code, with emphasis on the tooltip positioning.

The tooltip positioning has been wrong for a long time, I even submitted two .html files that show bugs in the tooltip positioning in the older versions (test/scrollTest.html and test/scrollTest2.html). I rewrote all the code to fix these positioning bugs and remove that ugly broken code that is used for the interactive guideline tooltip. I fixed a TODO from the tooltip code also.

The tooltip behaviour when being displayed didn't change, the position and gravity are still the same. I fixed a bug that happens when the tooltip goes up and gets outside the viewport. Now it is staying correctly in the viewport.

Although the tooltip behaviour is the same, this __is__ a (small) breaking change for the API, so it may be a good idea to bump the minor version when releasing a new tag. Because this already is a breaking change,  I removed the deprecated tooltip functions, as this will need a new minor version (1.9).

Here are the breaking changes:
- Removed the offset attribute: it just doesn't make sense to me. Why always subtract this offset, without minding the gravity?
- Removed the fixedTop attribute: why the fuck would I fix the tooltip on the top of the page and make it disappear when scrolling?
- Removed the original position attribute: now it is a function that returns where to position the tooltip. This function can be overridden if you want and, by default, returns the current mouse position. Because of this, the charts do not need to specify the position directly anymore, this is done inside the tooltip code. However, if they want to actually change the position to one different from the mouse position, that can be easily done, it just needs to override the position() function and return the object that was commented in the code.

This pull request is quite big, but it fixes a lot of bugs and cleans up the code and API.
This fixes #1138 (also #1124, although it is already fixed).


RenatoUtsch:
It seems like I have to update the tests.

Alex--wu:
Hi RenatoUtsch, I am interested of your code and I have a question I hope you can help me. For the current tooltip, I can't make multi-tooltip on the axis. For the old version I can do that by using nv.tooltip.show. Here is the screenshot:


It seems you are expert of the tooltip, that's why I ask you about it. :)

RenatoUtsch:
It seems that this feature is not in the codebase anymore. It probably was removed when they deprecated nv.tooltip.show in favor of nv.models.tooltip (I wasn't here at the time).

liquidpele:
Yea, I removed the small tooltips that showed up on the axis in the scatter plot, and robin was also in support of it.  It really didn't fit a tooltip model well and was kind of hacked in for... I have no idea why really since the same data shows in the tooltip itself.

Alex--wu:
@liquidpele Yes I agree that was kind of hacked in .... But now I have no idea how to make the tooltips showing on the yAxis like the screenshot I show above. Because the number of yTooltips is dynamic.

A real use case is: Our charts in the product need to show tooptips on the Axis, not showing the box tooltips inside the graphic area because it will cover the bar/line. This is real use case in our financial software.

liquidpele:
Well, I did make the tooltip see-through so that it didn't hide the things behind it, you could decrease that opacity further if you wanted.

Alex--wu:
@liquidpele Thank you for your answer. Hide the tooltip is not the problem, the problem is how to show tooltips on yAxis.

RenatoUtsch:
So... about the pull request? Because this has kind of digressed.

liquidpele:
ping!   sorry, will look at this when I can

src/models/multiChart.js
```diff
@@ -293,12 +292,11 @@ nv.models.multiChart = function() {
                     color: evt.point.color
                 };
                 tooltip
-                    .duration(100)
+                    .duration(0)
```
liquidpele: why do you want duration 0 here?  Then it would jump instead of moving smoothly around between scatter points.

src/tooltip.js
```diff
@@ -1,440 +1,327 @@
...
- window.nv.tooltip.* also has various helper methods.
+/* Model which can be instantiated to handle tooltip rendering.
+ Example usage:
+ var tip = nv.models.tooltip().gravity('w').distance(23)
+ .data(myDataObject);
+
+ tip();    //just invoke the returned function to render tooltip.
  */
-(function() {
+nv.models.tooltip = function() {
```
liquidpele: Wait, so now every time you call tooltip() it's going to calculate a new random ID and thus create a new dom object?

src/models/multiChart.js
RenatoUtsch: Sorry, don't remember why I changed this. I will change it back.

src/tooltip.js
RenatoUtsch: Well, of course. Every time you call nv.models.tooltip() you're asking for a new tooltip. If you want to use the same tooltip you just save the tooltip you created and use it. This is the same behavior of all the other models.

liquidpele: Oh wait, so this returns the tooltip object that then you're doing tooltip() on, I see, I had the two confused.

RenatoUtsch: Yes, this is the same as the old code. It is indeed a bit confusing hehe

src/tooltip.js
```diff
@@ -1,440 +1,327 @@
...
+        return false;
+    };
+
+    // Calculates the gravity offset of the tooltip. Parameter is position of tooltip
+    // relative to the viewport.
+    var calcGravityOffset = function(pos) {
+        var height = tooltip.node().offsetHeight,
+            width = tooltip.node().offsetWidth,
+            clientWidth = document.documentElement.clientWidth, // Don't want scrollbars.
+            clientHeight = document.documentElement.clientHeight, // Don't want scrollbars.
```
liquidpele: can documentElement have a margin?  If so, that messes up the calc...

src/tooltip.js
```diff
@@ -1,440 +1,327 @@
...
-                    _.series.value = _.point.y;
-                    _.series.color = _.point.color || _.series.color;
-                }
-                data = _;
-            }},
-
-            // read only properties
-            tooltipElem: {get: function(){return tooltipElem;}, set: function(_){}},
-            id: {get: function(){return id;}, set: function(_){}}
+            positionTooltip();
```
liquidpele: .... so it looks like you're essentially rewriting the tooltip content every mouse move here?

RenatoUtsch: Not the tooltip content, but the tooltip position. This is the same as the old version, I only cleaned it up:
https://github.com/novus/nvd3/blob/master/src/tooltip.js#L345

src/tooltip.js
RenatoUtsch: http://stackoverflow.com/questions/1248081/get-the-browser-viewport-dimensions-with-javascript
http://andylangton.co.uk/blog/development/get-viewport-size-width-and-height-javascript

I used only clientWidth and clientHeight because innerWidth and innerHeight have wrong values on some mobile browsers.

This doesn't seem to be a problem.

RenatoUtsch:
Up.

Updated the pull request with the latest commits.
Will this be accepted? It's been 1 month already.

liquidpele:
Okay, well it's a huge merge but lets do it and see what breaks (if anything)... I'd rather avoid any further merge conflicts.

liquidpele:
I built latest so that's up now.

RenatoUtsch:
If anything breaks I'll help fixing it. But remember that the API changed a little so anything that used the position(), offset() or fixedTop() values is expected to break.

liquidpele:
Well, I think I added position, so that was pretty new... fixedTop was added pretty recently by someone else IIRC so they might be mad about that...   offset I think was always there?  Not sure.  

If someone is using those, does it just throw errors?   If so, we should probably keep the options and let them do nothing except log a deprecation message.

RenatoUtsch:
The other two properties just throw errors. I can add that deprecated thing for the properties and make them do nothing in another PR.

liquidpele:
Yea, please do.  It's always good to give some warning before something just starts throwing errors :)

PR #1213, opened 2015-08-26
fix tooltip key label in multichart (bar series)

scne:
Show the key label in multichart when a bar serie is enabled


liquidpele:
You have more than your own commits here.

PR #1214, opened 2015-08-26
BugFix: Don't loose classed for main chart in lineWithFocusChart

owaaa:
Classed was only being applied to the focus control chart and lost on the main chart as it was overwritten. This bug fix allows styling of the line series in the main chart.

Prior to Bug Fix:
image

Fixed:
image


liquidpele:

Issue #1216, opened 2015-08-27
Multibarhorizontal show width 1 when the data is undefined, I think that the width should be 0


rubenns:
into
 if (stacked)
                bars.watchTransition(renderWatch, 'multibarhorizontal: bars')
into 
 else
                bars.watchTransition(renderWatch, 'multibarhorizontal: bars')
original :
  .attr('width', function(d,i) {
                        return Math.max(Math.abs(y(getY(d,i)) - y(0)),1)
                    });

Modified
  .attr('width', function(d,i) {
                        return Math.max(Math.abs(y(getY(d,i)) - y(0)),1) || 0
                    });

PR #1217, opened 2015-08-27
return 0 when the y value is undefined

rubenns:
This commit is related with the issue :
Multibarhorizontal show width 1 when the data is undefined, I think that the width should be 0 #1216


liquidpele:
Looks fine, thanks for contributing!

rubenns:
you are welcome for the library :) . 

Do you know when you will do the build ?

PR #1224, opened 2015-08-31
Multi chart tooltip fix

uenz:
Now all keys (curve names) where displayed in multiChart tooltip. See sample file ./examples/multiChart.html


uenz:
This fix makes the key (curve name) available for all curves of the multiChart sample in the tooltip. See ./examples/multiChart.html

liquidpele:
thx!

PR #1225, opened 2015-08-31
Added event dispatch for mouseDown and mouseUp in interactivelayer to…

mehranshakeri:
… allow dragging to select an area


liquidpele:
Sorry, what is the purpose of this?

liquidpele:
eh, whatev, it doesn't hurt anything :)

mehranshakeri:
My specific problem was that I wanted to allow user to draw a rectangle in chart, and I could zoom in for selected area. But with current implementation I didn't have mouseDown and mouseUp events. I tried to add another rect on top of others, but then I missed the interaction with the chart. I thought this way would be cleaner and maybe useful for others. :D
thank you for merge

Issue #1228, opened 2015-09-02
Using dates on X axis with multiChart


syndesis:
Hello,

I had to modify the source (line 93 of multiChart.js) to get dates to work:
 x   .domain(d3.extent(d3.merge(series1.concat(series2)), function(d) { return new Date(d.x) } ))
                .range([0, availableWidth]);

Using
config: {
x: function(d) {return new Date(d.x).getTime();},
...
xAxis: {
tickFormat: function(d) {return d3.time.format('%e')(new Date(d))
}
}

Without the change, tickFormat gets passed NaN.  Obviously that change isn't suitable for graphs with non-date X axes.  What would be the preferable way to implement this?  A config option?

liquidpele:
Feel free to create a pull request with your change for us to review.

liquidpele:
line in question:

https://github.com/novus/nvd3/blob/master/src/models/multiChart.js#L93

syndesis:
Pull request submitted.

leisms:
Hi there, I'm also trying to use dates on the xAxis and am running into NaNs being passed into tickFormat.

I pulled the latest code which includes the pull request which is supposed to fix this problem, but the issue persisted. I also tried passing in the date as an integer Unix time, but that came back NaN as well.

Here's the relevant config:

```coffeescript
        chart:
            type: 'multiChart'
            height: 200
            x: (d) -> d[0]
            y: (d) -> d[1]
            xAxis:
                tickFormat: (d) ->
                    # d is NaN
                    date = moment.tz(d,$scope.timeZoneId)
                    switch $scope.dateRange.range
                        when "1D"
                            formattedDate = date.format("ha")
                        when "1Y"
                            formattedDate = date.format("MMM")
                        else
                            formattedDate = date.format("l")
                    return formattedDate
```

leisms:
Just a heads up that I ended up using the line change @syndesis originally suggested. I don't think the merged pull request solved the issue.

PR #1229, opened 2015-09-02
Allow rendering of dates on X axis in multiChart

syndesis:
This change allows the rendering of dates in multiChart as described in issue #1228.  It adds a config option called xAxisDates.

### DIFF
--- a/src/models/multiChart.js
+++ b/src/models/multiChart.js
@@ -90,7 +90,7 @@ nv.models.multiChart = function() {
                     })
                 });
 
-            x   .domain(d3.extent(d3.merge(series1.concat(series2)), function(d) { return d.x } ))
+            x   .domain(d3.extent(d3.merge(series1.concat(series2)), function(d) { return getX(d) }))
                 .range([0, availableWidth]);
 
             var wrap = container.selectAll('g.wrap.multiChart').data([data]);
--- a/src/models/multiChart.js
+++ b/src/models/multiChart.js
@@ -16,7 +16,8 @@ nv.models.multiChart = function() {
         getX = function(d) { return d.x },
         getY = function(d) { return d.y},
         interpolate = 'monotone',
-        useVoronoi = true
+        useVoronoi = true,
+        xAxisDates = false
         ;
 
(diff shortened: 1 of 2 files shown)


src/models/multiChart.js
```diff
@@ -90,8 +91,8 @@ nv.models.multiChart = function() {
                     })
                 });
 
-            x   .domain(d3.extent(d3.merge(series1.concat(series2)), function(d) { return d.x } ))
-                .range([0, availableWidth]);
+            x   .domain(d3.extent(d3.merge(series1.concat(series2)), function(d) {
+                return xAxisDates ? new Date(d.x) : d.x } )).range([0, availableWidth]);
```
liquidpele: Hmm... Seems like this should be using chart.x()(d) to get the value... Then you could just override the x getter option to return a date....

syndesis: That does seem like a better idea, yes.

syndesis:
Allow rending of dates (or anything) on X axis in multiChart.  Using new fix suggested by @liquidpele in #1228

src/models/multiChart.js
```diff
@@ -90,7 +90,7 @@ nv.models.multiChart = function() {
                     })
                 });
 
-            x   .domain(d3.extent(d3.merge(series1.concat(series2)), function(d) { return d.x } ))
+            x   .domain(d3.extent(d3.merge(series1.concat(series2)), function(d) { return chart.x()(d) } ))
```
liquidpele: Oh, it looks like right above this it uses the getX function (which is what chart.x() actually returns)... so should probably just use that for consistency...     sorry!   

If you change that (and please verify it actually works for your use case) then I'll merge it in :)

syndesis:
Updated to use getX.. still works in my use case.

syndesis:
Oops, just noticed this included some unrelated changes I was working on.  Will fix that.

syndesis:
Sorry, pushed an unrelated change.  Will open a new issue for that.

liquidpele:

Issue #1231, opened 2015-09-03
Adding option for line icon in legend


syndesis:
I wanted to use a line icon instead of the circle in the legend for a multiChart with a line graph in it.  I added a data option called legendIconType to allow this.  For example:

var data = [{type: 'line', values: [1, 2, 3, 4, 5], yAxis: 1, legendIconType: 'line'}];

It's rendered initially as stroke-width: 4 and shrinks to 2 when it's disabled.

liquidpele:
Send up a pull request... if you use the built in d3 symbols, check in nv.util, there is a custom symbol getter there that allows people to add their own symbols to use (currently used by scatter chart).

PR #1235, opened 2015-09-03
Add option to data to allow line symbol in legend

syndesis:
I don't have time right now to implement this as symbols.. maybe in the future.  Would probably require another legend vers. Re: issue #1231


liquidpele:
Going to close this, as it would be better I think to use a d3 symbol and provide custom symbols which is already an ability for the scatter chart.

PR #1240, opened 2015-09-06
Unit property for Tooltip

uenz:
For scientific plots it would be nice to have a tooltip with name, value and unit and not only name and value. I added the unit property to the graphs. When the property is not set the behaviour is like usual. If the unit property is set, it will be appended to the Tooltip. I modified the lineChart.html sample to show the functionality. 
screen shot 2015-09-06 at 21 32 26
screen shot 2015-09-06 at 21 31 53


liquidpele:
You can already override the value formatter for the tooltip, so you can add whatever you want to those values.

mgalloue:
Hi @liquidpele, the problem is when we override the value formatter, we only have the information on the value, we don't have the key for example to identify which unit should be displayed or maybe I missed something?
Thanks for your help

PR #1242, opened 2015-09-07
Added percent calculation and corresponding column

mehranshakeri:
"showPercent" adds percent column at the right side of tooltip.
"getPercent" function calculates percent value. This function can be replaced with a custom one.


liquidpele:
This seems like you should just make your own tooltip content for special cases like this.

PR #1251, opened 2015-09-11
Add deprecated messages for tooltip options that were removed.

RenatoUtsch:
Follow-up of #1190.

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -26,7 +26,6 @@ nv.models.tooltip = function() {
         ,   gravity = 'w'   // Can be 'n','s','e','w'. Determines how tooltip is positioned.
         ,   distance = 25 // Distance to offset tooltip from the mouse location.
         ,   snapDistance = 0   // Tolerance allowed before tooltip is moved from its current position (creates 'snapping' effect)
-        ,   fixedTop = null // If not null, this fixes the top position of the tooltip.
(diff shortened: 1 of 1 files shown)


liquidpele:
Actually, with offset removed, what would be the correct way to adjust where the tooltip hovered in relation to the mouse?     For instance, on a pie chart the tooltip follows the mouse, but lets say I wanted it to be 20px to the right instead of the default, how would I do that?

RenatoUtsch:
It doesn't make sense to make the tooltip 20px to the right, because if the orientation changes when you're on the border of the page, this 20px could make the tooltip overlap with the mouse.

What does make sense is specify the distance to the mouse (note I'm not specifying a direction), and then coordinate the position of the tooltip with the orientation. This can be done with the distance property.

liquidpele:
Ah, so the distance property is the distance from the mouse given the orientation?

liquidpele:
Neat, I guess I missed that at some point

RenatoUtsch:
I found that the distance property had this behavior when I inspected the
source code, and decided to mantain it and remove the offset property that
as I said before has some problems.

I don't remember if distance was documented, if it wasn't I added a comment
to it.

Em sex, 11 de set de 2015 12:12, liquidpele <notifications@github.com>
escreveu:

RenatoUtsch:
@liquidpele, I think you probably need to rebuild after this, or the errors will continue.

PR #1257, opened 2015-09-17
Fixed bug when yDomain[0] is negative

sezgi:
Fixed bug where, when yDomain[0] is negative, bars are too tall (since the subtraction is based on y(yDomain[0]) instead of y(0)).


sezgi:
Here's a jsfiddle showing the bug:

https://jsfiddle.net/ahrjmrqe/1/

If you comment out line 47 (`.yDomain([-50, 100])`), the bug goes away.

liquidpele:

PR #1260, opened 2015-09-23
Integrate Meteor Tasks Into GruntFile

jdlubrano:
Whenever a new release is created for NVD3, a new package should be published to [Atmosphere](atmospherejs.com).  This pull request integrates the `meteor publish` command into the existing `grunt release` task.

### DIFF
--- a/GruntFile.js
+++ b/GruntFile.js
@@ -1,5 +1,8 @@
 module.exports = function(grunt) {
     var _pkg = grunt.file.readJSON('package.json');
+    // Meteor packages must have a number in order to play nicely
+    // with the packaging system
+    var versionNumber = _pkg.version.replace(/(\d\.\d\.\d).*$/g, "$1");
 
     //Project configuration.
     grunt.initConfig({
@@ -56,7 +59,7 @@ module.exports = function(grunt) {
                 overwrite: true,
                 replacements: [{
                     from: /(version?\s?=?\:?\s\')([\d\.]*)\'/gi,
-                    to: '$1' + _pkg.version + "'"
+                    to: '$1' + versionNumber + "'"
                 }]
             }
         },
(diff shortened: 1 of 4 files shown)


GruntFile.js
```diff
@@ -125,6 +128,18 @@ module.exports = function(grunt) {
                     ]
                 }
             }
+        },
+        exec: {
+            "meteor-init": {
+                // Make sure that Meteor is installed.
+                command: "type meteor > /dev/null 2>&1 || { curl https://install.meteor.com/ | sh; }"
```
liquidpele: whoa...    you should be using https://github.com/MeteorPackaging/grunt-gulp-meteor  I think, not downloading manually like that.

liquidpele: oh, it literally says to do just that on that page.    wow.   that's terrible.

jdlubrano: So I'll admit that I was just somewhat blindly following the guide. If you think that Gulp is a better method, then I say let's go for it. I personally don't think that Meteor is developed enough in terms of workflows and continuous integration to dictate the "proper" way to setup Grunt, Travis, etc. I have never used Gulp. Do you see a superior solution? If so, I am all for it.

-----Original Message-----

liquidpele:
So, I'm looking over this... their documentation is confusing as hell...  but I found this which lists 1.7.1 and 1.8.1... 

https://atmospherejs.com/nvd3/nvd3

And from what I can tell, it's supposed to just work with github once you register the app, so I'm not sure why we would need to integrate doing a meteor publish every time... their page on publishing only shows it doing that once, much like npmjs.org or bower: 

https://atmospherejs.com/i/publishing

So, I'm not really confident that this pull request is even the right approach.   What problem is this supposed to be solving exactly?

jdlubrano:
This is potentially embarrassing on my part.  Originally, I had thought that the changes from PR #1245 did not propagate to Atmosphere JS.  I had attempted to add the `nvd3:nvd3` Meteor package to a Meteor project shortly after (I think hours after) PR #1245 was merged with master, and still found that the `nv` object was undefined at the global scope.  As I added the `nvd3:nvd3` package to a test application this morning everything worked just fine.  Perhaps Atmosphere does pickup changes from Github albeit after a brief period of time.

Just curious, did you happen to run `meteor publish` manually or via the Grunt tasks that I added?

If it is the case that `meteor publish` was never executed through any means, then this pull request is no longer needed.

I do, however, still like the idea of integrating the Meteor specific tests into the Gruntfile, but I would submit a new PR with much less added/changed in the Gruntfile.  Am I correct in assuming that you would desire to see the Meteor tests integrated into your Grunt workflow?

liquidpele:
I've never done anything with meteor besides google it and read some docs the other day. I'll close this out for now.

PR #1264, opened 2015-09-26
Unify lineChart and lineWIthFocusChart

tony-jacobs:
The lineChart and lineWithFocusChart have subtly different options and is not drop-in replaceable when trying to add a focus widget to a line chart.  There's no obvious reason for that.

So, I've taken the lineWithFocusChart model, forward ported the recent renderWatch and tooltip data changes, and then renamed it back to lineChart.


liquidpele:
So, you basically combined lineChart and lineWithFocusChart?    Cool, I had been meaning to do that... however, you'll need to add a way to hide/show the focus area.   Check out the linePlusBarChart, it was combined with a linePlusBarWithFocusChart already, and allows the focus part to be hidden.  

If you do that though, you'll want to delete the lineWithFocusChart file, set the focus area in lineChart to be hidden by default, and add a small lineWithFocusChart object in the same file that just calls lineChart and sets the focus to be shown for you for backward compatibility.

tony-jacobs:
I'd done the thing to do the disable and set the default to false.  Turns out that the nv-background wasn't resizing to follow along, so this fixes that.  I'll start a patch to blow away the lineWithFocusChart shortly

tony-jacobs:
Ok, the compatibility shim is now in place - please review and merge at your leisure.

liquidpele:
Wow, that was fast.   Looks good to me, good work and thanks for contributing :)

PR #1265, opened 2015-09-26
jslint fixes for lineChart

tony-jacobs:
This is a style-only change set - it just fixes a bunch of as missing ; and avoids using non-coercive comparisons (=== !==) to null


liquidpele:
looks good

PR #1268, opened 2015-09-28
Different colors for leaves of the same parent.

Crenshinibon:
I need different colors for the leaves of the same parent in a sunburst diagram, the only way to accomplish this, afaik, is to change the fill-call in the sunburst model.

Unfortunately the current behaviour, coloring all leaves of a parent with the parents color is broken that way. For me it's fine. But I guess others might have a different opinion for that matter.

### DIFF
--- a/src/models/sunburst.js
+++ b/src/models/sunburst.js
@@ -79,7 +79,7 @@ nv.models.sunburst = function() {
                 .append("path")
                 .attr("d", arc)
                 .style("fill", function (d) {
-                    return color((d.children ? d : d.parent).name);
+                    return color(d.name);
                 })
                 .style("stroke", "#FFF")
                 .on("click", function(d) {


liquidpele:
No, we should fix to handle both ways.  Like using a "color" attribute in the data if provided so people can override individual slices.  That's already the case in several graphs besides sunburst.  Or a new option to not group the color by parent slice.  Or both!

Crenshinibon:
Even better.

I haven't looked at the implementation of the other graphs, but that doesn't sound very difficult to achieve. Let me know if I can help out with this.

Regards
Dirk

liquidpele:
lol, well, by "we" I actually meant "you" ;)

If you change your pull request to do that, then I'll check it out and merge it.

Crenshinibon:
I made a new one. Unfortunately I checked in the build results. Just realized that ...

Issue #393, opened 2013-12-22
Add composer.json (for php projects using composer)


kmindi:

liquidpele:
Not enough info, please submit a pull request with the changes so it can be reviewed.

PR #1270, opened 2015-09-28
Added composer file to allow easier deployment

chewett:
Adding a composer.json file brings nvd3 into line with d3 and its use of
composer files. This will allow PHP and any projects using composer to
more easily integrate the nvd3 library like that do with d3.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,32 @@
+{
+  "name": "novus/nvd3",
+  "description": "A reusable charting library written in d3.js",
+  "keywords": [
+    "nvd3",
+    "d3",
+    "chart",
+    "graph"
+  ],
+  "homepage": "https://github.com/novus/nvd3",
+  "license": "Apache-2.0",
+  "authors": [
+    {
+      "name": "Bob Monteverde"
+    },
+    {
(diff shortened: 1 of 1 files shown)


chewett:
Adding a composer.json file allows PHP and any other projects using composer to easily install nvd3 with composer. This file mimics the one that d3 uses (https://github.com/mbostock/d3/blob/master/composer.json) and will ease integration into these type of projects.

This has been requested previously (https://github.com/novus/nvd3/issues/393) but a pull request was never followed through with.

To fully increase the usability you will need to follow step 3 of this guide (https://knpuniversity.com/screencast/question-answer-day/create-composer-package#step-3-registering-with-packagist) which will properly register the nvd3 package and allow one step installs with composer. If this is not done, merely merging the file will increase the usability of nvd3 but I strongly recommend adding it on packagist.

chewett:
@liquidpele will this be added to packagist.org?

chewett:
Also, After testing I have noted that the composer.json file wont work unless it is in a release tag. So this wont work until 1.8.2 is released (or similar)

Thanks for the merge.

PR #1271, opened 2015-09-28
Sunburst acknowledges color attribute in data and groupColorByParent option integrated

Crenshinibon:
the latter is set to true by default, so the coloring behaves as before.


liquidpele:
Please revert the stuff under the build directory, thanks!

src/models/sunburst.js
```diff
@@ -187,6 +196,7 @@ nv.models.sunburst = function() {
         mode:       {get: function(){return mode;}, set: function(_){mode=_;}},
         id:         {get: function(){return id;}, set: function(_){id=_;}},
         duration:   {get: function(){return duration;}, set: function(_){duration=_;}},
+        groupColorByParent: {get: function(){return groupColorByParent;}, set: function(_){groupColorByParent=_;}},
```
liquidpele: please do groupColorByParent=!!_;  so that it guarantees that it saves a boolean.

src/models/sunburstChart.js
```diff
@@ -126,6 +127,10 @@ nv.models.sunburstChart = function() {
             color = _;
             sunburst.color(color);
         }},
+        groupColorByParent: {get: function(){return groupColorByParent;}, set: function(_){
```
liquidpele: You don't need to add this here, it inherits the options from sunburst.js

See this line below:
    nv.utils.inheritOptions(chart, sunburst);

src/models/sunburst.js
```diff
@@ -14,6 +14,7 @@ nv.models.sunburst = function() {
         , id = Math.floor(Math.random() * 10000) //Create semi-unique ID in case user doesn't select one
         , container = null
         , color = nv.utils.defaultColor()
+        , groupColorByParent = true
```
liquidpele: Please add info about this new option to the json object at the bottom of examples/documentation.html so that people know what it is and how to use it :)

Crenshinibon:
Where should in the docs should I mention the availability of the 'color' attribute in the data? And shouldn't there be an option to change the attributes name that gets interpreted as 'color' somewhere?

liquidpele:
We haven't made a way to tie attributes into the docs yet, so you can ignore that for now...

liquidpele:

PR #1276, opened 2015-09-30
Use tooltip formatter for interactiveGuideline instead of axisFormatter

tony-jacobs:
There's two meaningful fixes here:
1. The interactive tooltip's header had been using the xAxis tickFormat.  Instead, use the formatter that's associated with the tooltip.  It's a no-op for folks who have not custom-set the tooltip headerFormat.

2. While auditing to find that fix, I noticed that the data value being used was indexed into series and not currentValues.  currentValues is a filtered copy of series, so there's a subtle bug.

### DIFF
--- a/src/models/lineChart.js
+++ b/src/models/lineChart.js
@@ -379,14 +379,13 @@ nv.models.lineChart = function() {
                         allData[indexToHighlight].highlight = true;
                 }
 
-                var xValue = tooltip.headerFormatter()( chart.x()(singlePoint,pointIndex) );
                 interactiveLayer.tooltip
                     .chartContainer(that.parentNode)
                     .valueFormatter(function(d,i) {
                         return d === null ? "N/A" : yAxis.tickFormat()(d);
                     })
                     .data({
-                        value: xValue,
+                        value: chart.x()( singlePoint,pointIndex ),
                         index: pointIndex,
                         series: allData
                     })();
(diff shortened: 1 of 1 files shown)


src/models/lineChart.js
```diff
@@ -379,7 +379,7 @@ nv.models.lineChart = function() {
                         allData[indexToHighlight].highlight = true;
                 }
 
-                var xValue = xAxis.tickFormat()(chart.x()(singlePoint,pointIndex));
+                var xValue = tooltip.headerFormatter()( chart.x()(singlePoint,pointIndex) );
```
liquidpele: It seems to me that it shoudl just be sending the unformatted value since the tooltip calls headerFormatter already... the bug is really that we were doing tickFormat() at all there...

tony-jacobs:
Ok - updated to delegate to the interactiveLayer.tooltip, and also updated the relevant setters so users have a chance of it working.

src/models/lineChart.js
```diff
@@ -611,6 +610,8 @@ nv.models.lineChart = function() {
         xTickFormat: {get: function(){return xAxis.tickFormat();}, set: function(_){
             xAxis.tickFormat(_);
             x2Axis.tickFormat(_);
+            tooltip.headerFormatter(_);
+            interactiveLayer.tooltip.headerFormatter(_);
```
liquidpele: No, don't do this here.... tooltip.headerFormatter() is already set to a default on line 58, you should just do the same thing with the interactiveLayer's tooltip.   Using the method on line 58 people can update xTickFormat and it'll use that change, but if they manually update headerFormatter then it will keep that so that there aren't issues.

tony-jacobs:
Good catch.  Then, the yAxis stuff will just work as well.

Fixed.

liquidpele:
Looks good!

PR #1282, opened 2015-10-03
Performance improvements for lineChart

tony-jacobs:
When lineChart was unified with lineWithFocusChart, a bug crept in cause the main chart to be called twice (once during the standard update, again during the onBrush() update).  This patch makes those two updates to the main chart mutually exclusive, based on the focusEnable flag.

The majority of the diff is whitespace, now that the focusChart logic is wrapped by an if.  

A few other changes are along for the ride:
* The background rects are now added with the rest of the skeleton code instead of (confusingly) later in the code.

* Avoid creating a local variable called 'that' on general principal.  It was only used in one place.

* The scatter model now offers an option for the interactiveUpdate duration - it had been hard coded to 300ms.  On my machine, that deferred update was taking 8ms to set up while a direct update was completed in under a millisecond. Library users can now choose to accept a degraded animation in favor of faster raw chart updates.  Note: I wasn't able to visually spot a difference from running at the default 300ms, so perhaps it's safe to set the default on that to 0?

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -425,7 +425,9 @@ nv.models.scatter = function() {
                 timeoutID = setTimeout(updateInteractiveLayer, interactiveUpdateDelay );
             }
             else
+            {
                 updateInteractiveLayer();
+            }
 
             //store old scales for use in transitions on update
             x0 = x.copy();
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -40,6 +40,7 @@ nv.models.scatter = function() {
         , dispatch     = d3.dispatch('elementClick', 'elementDblClick', 'elementMouseover', 'elementMouseout', 'renderEnd')
         , useVoronoi   = true
         , duration     = 250
+        , interactiveUpdateDelay = 300
         ;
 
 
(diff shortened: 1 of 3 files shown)


src/models/scatter.js
```diff
@@ -418,9 +419,13 @@ nv.models.scatter = function() {
...
-            clearTimeout(timeoutID); // stop repeat calls to updateInteractiveLayer
-            timeoutID = setTimeout(updateInteractiveLayer, 300);
-            //updateInteractiveLayer();
+            if( interactiveUpdateDelay )
+            {
+                clearTimeout(timeoutID); // stop repeat calls to updateInteractiveLayer
+                timeoutID = setTimeout(updateInteractiveLayer, interactiveUpdateDelay );
+            }
+            else
+                updateInteractiveLayer();
```
liquidpele: Please always use brackets

src/models/scatter.js
```diff
@@ -491,6 +496,7 @@ nv.models.scatter = function() {
         clipRadius:   {get: function(){return clipRadius;}, set: function(_){clipRadius=_;}},
         showVoronoi:   {get: function(){return showVoronoi;}, set: function(_){showVoronoi=_;}},
         id:           {get: function(){return id;}, set: function(_){id=_;}},
+        interactiveUpdateDelay: {get:function(){return interactiveUpdateDelay;}, set: function(_){interactiveUpdateDelay=_;}},
```
liquidpele: New options should be documented in the large json object in examples/documentation.html

tony-jacobs:
Added documentation and braces per comments.

PR #1288, opened 2015-10-06
Add defined to stacked area chart

slestang:
I don't know why *defined()* exist in line chart but not in stacked area chart but I needed it. So here comes this PR where I added it.
I tested it on several stacked area chart and it works fine.


liquidpele:

PR #1290, opened 2015-10-06
Fixes #1278 classed option for series

npezolano:

PR #1294, opened 2015-10-08
Fix Travis CI then Fix Bugs and Tests

jdlubrano:
I made a significant and terrible mistake with #1245.  In essence, I messed up the Travis CI configuration file such that *only* Meteor tests were being run as a part of this project's continuous integration.  This PR fixes my Travis mistake and attempts to clean up the resulting collateral damage.

#### Collateral Damage

Perhaps fortunately, there was only one "bug" (according to the current suite of mocha tests) introduced between #1245 and now.  It looks like #1282 introduced a bug where axes were not rendered on a line chart if `focusEnabled` was set to `false`.  There is a patch in `lineChart.js` in this PR to fix that.

On a lesser note, #1290 changed how series are classed in a scatter model which, in effect, doubles the number of `<g>` elements that receive a class from a data parameter in a line chart.  For instance

```
var data = [
  {
    key: 'series1',
    classed: 'foo',
    values: [...]
  },
  ...
];
```

would result in a line chart with two `<g>` elements classed as `foo`, one group belongs to the `.nv-scatterWrap` and one represents the actual line in the chart.  I modified the existing test case in `line.coffee` to check for two classed elements and to confirm that one was indeed part of the scatter.

Moving forward, it might be desirable to wrap the lines and `.nv-scatterWrap` in separate divs such that some query selector could reasonably distinguish between the two.  Nevertheless, for now, I wanted to get this PR in the pipeline as soon as possible considering that the mistakes that I made in #1245 are so significant to the testing workflow of the project.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,12 @@ before_install:
     - "export DISPLAY=:99.0"
     - "sh -e /etc/init.d/xvfb start"
 #   Meteor Tinytest support
-    - "curl https://install.meteor.com | /bin/sh"
-    - export PATH="$HOME/.meteor:$PATH"
     - "npm install -g spacejam"
 install:
     - "npm install"
     - "bower install"
 
-script: "spacejam test-packages ./"
+script: 
+    - "npm test"
+    - "spacejam test-packages nvd3:nvd3"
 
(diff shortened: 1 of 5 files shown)


robinfhu:
Cool thanks!

liquidpele:
Looks okay to me, thanks for fixing that!

PR #1328, opened 2015-10-29
fix: zero line in discreteBarChart model

### LINKED ISSUE #1327: fix: zero line in discreteBarChart model
When [```rightAlignYAxis```](https://github.com/nvd3-community/nvd3/blob/master/src/models/discreteBarChart.js#L20) is set to ``` TRUE ``` zero line starts from the beginning of YAxis instead of the beggining of chart.

Steps to reproduce: 
 - create discreteBarChart with some negative values in data and right allined YAxis
 - for testing reasons hide tick lines with following CSS

``` CSS
g.tick {
    opacity: 0 !important;
}
```

Expected result:
expected result

Actual result: 
actual result

tgosp:
Close #1327

### DIFF
--- a/src/models/discreteBarChart.js
+++ b/src/models/discreteBarChart.js
@@ -193,7 +193,7 @@ nv.models.discreteBarChart = function() {
             // Zero line
             g.select(".nv-zeroLine line")
                 .attr("x1",0)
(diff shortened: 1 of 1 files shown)


liquidpele:
Looks fine, thanks for contributing!

PR #1329, opened 2015-10-30
Add Interactive guildline to multiChart

balmma:
There is no interactive guidline in multiChart and the existing tooltips are a bit cumbersome to use. So this might be an implementation mostly inspired by lineChart.

### DIFF
--- a/src/models/multiChart.js
+++ b/src/models/multiChart.js
@@ -369,7 +369,58 @@ nv.models.multiChart = function() {
               }
             }
 
-            if(!useInteractiveGuideline){
+            if(useInteractiveGuideline){
+                interactiveLayer.dispatch.on('elementMousemove', function(e) {
+                    clearHighlights();
+                    var singlePoint, pointIndex, pointXLocation, allData = [];
+                    data
+                    .filter(function(series, i) {
+                        series.seriesIndex = i;
+                        return !series.disabled;
+                    })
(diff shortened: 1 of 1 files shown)


liquidpele:
Someone just make a new PR for this same thing...   https://github.com/novus/nvd3/pull/1368

You guys might want to look over each other's pull requests... I notice there are some big differences.

balmma:
I guess the big difference is mainly due to the fact that my pull request is for multiChart, the other one for multiBarChart :-)

liquidpele:
oh, haha, Hadn't had my coffee yet!

src/models/multiChart.js
```diff
@@ -334,47 +349,120 @@ nv.models.multiChart = function() {
...
+                });
+                bars1.dispatch.on('elementMousemove.tooltip', function(evt) {
+                    tooltip();
+                });
+                bars2.dispatch.on('elementMousemove.tooltip', function(evt) {
+                    tooltip();
+                });
+            }
+
+            interactiveLayer.dispatch.on('elementMousemove', function(e) {
```
liquidpele: shouldn't all the interactiveLayer logic be inside an else{} from the above if{} block?  It would make no sense to run all that if it's turned off...

balmma: Yes, makes sense. Will refactor that.

liquidpele:

PR #1330, opened 2015-10-30
Correctly handle empty or no markers in bulletChart

balmma:
When the markers Array for a bulletChart is not set or empty I would expect to have no markers on the chart at all.

But in fact the markers Array gets defaulted to [0] which places one marker at position 0. When I provide an empty Array, the range calculation fails and I see no x axis

This fix should solve this issue.


liquidpele:
Looks good, also thank you for the extra test!

Issue #1332, opened 2015-10-30
Pie Char labels without text should not be hashed.


MarekBleschke:
Very similar issue to #529. Labels that don't have text value (because of percent < labelThreshold condition in pie.js) are affecting labels with text.

jsFiddle: https://jsfiddle.net/fwyunr30/3/

In this example J label is "pushed" by H and I.

PR #1333, opened 2015-10-30
Bug #1332 - don't hash labels without text

MarekBleschke:


liquidpele:
Seems fine, thanks for contributing!

PR #1338, opened 2015-11-03
Allow an arbitrary number of ranges and sort labels accordingly to their values in BulletChart

balmma:
Currently it's only possible to define three ranges (min, mean, max). It might be quite usefull to define an arbitrary number of ranges.

Also, the existing code sorts the values in descending order,  but leaves the label Arrays untouched. This might cause some confusion when the values are not already provided in the right order.

This pull requests allows to define an arbitrary number of ranges and keeps the labels in the right order. If no range labels are provided, the code falls back to the previous defaults (Minimum, Mean, Maximum).

### DIFF
--- a/src/models/bullet.js
+++ b/src/models/bullet.js
@@ -30,10 +30,11 @@ nv.models.bullet = function() {
         ;
 
     function sortLabels(labels, values){
+        var lz = labels.slice();
         labels.sort(function(a, b){
-            var iA = labels.indexOf(a);
-            var iB = labels.indexOf(b);
-            return values[iA] < values[iB];
+            var iA = lz.indexOf(a);
+            var iB = lz.indexOf(b);
+            return d3.descending(values[iA], values[iB]);
         });
     };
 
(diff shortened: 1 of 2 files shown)


balmma:
Well, it's probably better to keep the legacy class names for ranges there. Might be that somebody has already styled this classes and without them their code would break.

src/models/bullet.js
```diff
@@ -67,9 +88,10 @@ nv.models.bullet = function() {
             var gEnter = wrapEnter.append('g');
             var g = wrap.select('g');
 
-            gEnter.append('rect').attr('class', 'nv-range nv-rangeMax');
-            gEnter.append('rect').attr('class', 'nv-range nv-rangeAvg');
-            gEnter.append('rect').attr('class', 'nv-range nv-rangeMin');
+            for(var i=0,il=rangez.length; i<il; i++){
+                gEnter.append('rect').attr('class', 'nv-range nv-range'+i+' nv-range'+legacyRangeClassNames[i]);
+            }
```
liquidpele: If you have more than 3 ranges, you'll throw an index error here on legacyRangeClassNames...

balmma:
Well, it wouldn't throw an exception, but all ranges with i > 2 will be tagged with class name "nv-rangeundefined" which is not really nice. Will fix it in a way that legacy class names won't be added if i > 2.

balmma:
Should be fixed now.

liquidpele:
Okay, merging.

PR #1339, opened 2015-11-04
Reposition bullet markers on resize

balmma:
Currently,  markers in bullet charts are positioned when the chart gets created. Afterwards the marker positions are untouched which can cause displacements when for example resize the browser screen.

This pull request moves the translate attribute handling from gEnter to g.


liquidpele:
Seems fine, thanks for contributing!

Issue #1317, opened 2015-10-22
NVD3 Sunburst chart is not updated with new data/latest data


nounla:
Hi,
I am using NVD3 sunburst chart and we will get the chart dynamically by selecting some parameters and time range.
if we change the time range and parameters, we are getting latest data but the chart is not rendering new data or chart is not updated.
any suggestions to render chart with new data?

liquidpele:
try using the latest master branch I think someone contributed and added the update function to it but it was not in the last version

nounla:
I checked the latest also but whatever is in earlier version i see in latest version


            chart.update = function() {
                if (duration === 0)
                    container.call(chart);
                else
                    container.transition().duration(duration).call(chart)
            };

I dont see any new code change related to update.
can you give me any suggestion, how can we achieve this?

Crenshinibon:
I have seen the pull request #1272 that was supposed to fix the issue, but it's not working for me either. I build the js file from the master branch this afternoon.

Crenshinibon:
I'm trying to pinpoint the problem, but I don't really understand how that stuff is supposed to work (I'm taking pieChart as the reference). What I can tell is that the __data__ in the container is updated, BUT that the objects there aren't 'enriched' with the usual coordinates.

Which part of the 'chart' function is supposed to calculate that layouting stuff?

Crenshinibon:
My current workaround is to 'clear' the graph: updating with `data = undefined`and then update it with the data I actually like to display. It's not really pretty, but that's not that important for me.

Here is the actual Meteor code. Maybe it's helping:

```
if Meteor.isClient
  @chart  = nv.models.sunburstChart()
    .color (e1) -> 'black'
    .groupColorByParent false
    .mode 'size'

  @updateChart = (data) ->
    selected = d3.select '#sunburst'
    selected.datum(data).call chart
    chart.update()

  Template.sunburst.onRendered  () ->
    nv.addGraph () ->
      updateChart()
      chart

    this.autorun () ->
      size = _sizeSelection.get()
      filter = _filter.get().filterDef
      hierarchy = _hierarchySteps.find().fetch()

      Meteor.call 'getSunburstData', size, filter, hierarchy, (e, r) ->
        if e? or not r?
          console.log e, r
        else
          updateChart()
          updateChart [r]
```

Crenshinibon:
[...]

Issue #1284, opened 2015-10-05
showing lables and legends on the Sunburst chart  # liquidpele


nounla:
I am new to NVD3 sunburst chart and tried to show labels and legend for sun burst chart.
can we show labels and legend for sun burst chart using nvd3 and if yes, how?

liquidpele:
Nope, no labels or legend.  It depends on the tooltip as the data/slices are typically a lot more compact than a pie chart and the legend doesn't make much sense with the child/parent relationship of the data.  If you want to implement something though, feel free to send us a pull request.

PR #1343, opened 2015-11-06
Some changes/enhancements to Sunburst chart

Crenshinibon:
Hi,

I was actually trying to fix the updating of the sunburst chart (https://github.com/novus/nvd3/issues/1317), but utterly failed.

But along the way I cleaned up some stuff and added some other functionality.

- customizable labels
- tooltip value formatter as parameter
- sort parameter to specify the order of the arcs 

Labels were requested in this issue: https://github.com/novus/nvd3/issues/1284

### DIFF
--- a/examples/documentation.html
+++ b/examples/documentation.html
@@ -899,6 +899,26 @@ as well as used to print out examples using those example inputs.
             desc: "For sunburst only: specifies wether the leaf elements of a parent get the same color as the parent or if they are colored individually.",
             default: "true",
             examples: [true, false]
+        },
+        showLabels: {
+            desc: "For sunburst only: specifies if labels are displayed. The labels are centered on the arcs. The actual output can be adjusted by specifying the 'labelText' function.",
+            default: "false",
+            examples: [true, false]
+        },
+        labelText: {
+            desc: "For sunburst only: you can alter the label output by providing a callback function. The function gets the data of the current arc as parameter. It expects you to return the plain text to be shown.",
+            default: "function(d){if(mode === 'count'){return d.name + ' #' + d.value}else{return d.name + ' ' + d.value}}",
+            examples: [function(d){ return d.name;}]
+        },
+        sort: {
+            desc: "For sunburst only: let's you specify the sort order of the arcs.",
+            default: "function(d1, d2){return d1.name > d2.name;}",
+            examples: [function(d1, d2){return d1.value > d2.value;}]
+        },
+        tooltipValueFormatter: {
+            desc: "For sunburst only: let's you specify how the values in the tooltip should be formatted. It expects a callback function which is called with the value attribute of data as the only parameter.",
+            default: "function(v) {return v;}",
+            examples: [function(v){return Math.round(v);}]
         }
     };
 </script>
--- a/examples/documentation.html
+++ b/examples/documentation.html
@@ -890,10 +890,10 @@ as well as used to print out examples using those example inputs.
             desc: "For tooltip: returns the dom element of the tooltip.  This is read-only, you cannot set this value."
         },
         mode: {
-            desc: "For sunburst only: specifies the mode of drawing the sunburst segments. Can be 'size' or 'count'. 'size' draws the segments according to the 'size' attribute of the leaf nodes, 'count' draws according to the amount of siblings a node has.",
+            desc: "For sunburst only: specifies the mode of drawing the sunburst segments. Can be 'value' or 'count'. 'value' draws the segments according to the 'value' attribute of the leaf nodes, 'count' draws according to the amount of siblings a node has.",
             default: "count",
             refs: ["http://bl.ocks.org/kerryrodden/477c1bfb081b783f80ad"],
-            examples: ["size"]
+            examples: ["value"]
         },
         groupColorByParent: {
             desc: "For sunburst only: specifies wether the leaf elements of a parent get the same color as the parent or if they are colored individually.",
--- a/examples/documentation.html
+++ b/examples/documentation.html
@@ -414,7 +414,7 @@ as well as used to print out examples using those example inputs.
             examples: [0.05]
         },
         labelThreshold: {
-            desc: "Pie/donut charts: The slice threshold size to not display the label because it woudl be too small of a space",
+            desc: "Pie/donut and Sunburst charts: The slice threshold size to not display the label because it would be too small of a space",
             default: 0.02,
             examples: [0.05]
         },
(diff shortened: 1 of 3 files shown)


Crenshinibon:
selection_005

src/models/sunburst.js
```diff
@@ -60,16 +102,6 @@ nv.models.sunburst = function() {
...
-            chart.update = function() { 
-                if ( duration === 0 ) {
-                    container.call(chart);
-                } else {
-                    container.transition().duration(duration).call(chart);
-                }
-            };
-            chart.container = this;
-
-
```
Crenshinibon: I removed this, because it didn't solve the problem. In fact it wasn't called ...

src/models/sunburst.js
```diff
@@ -144,48 +239,51 @@ nv.models.sunburst = function() {
...
+            //       // at the current zoom level. (We only need to do this once.)
+            //       var xd = d3.interpolate(x.domain(), [node.x, node.x + node.dx]);
+            //       return function (t) {
+            //         x.domain(xd(t));
+            //         return tween(t);
+            //       };
+            //     } else {
+            //       return tween;
+            //     }
+            //   }
```
Crenshinibon: I commented out the function because it wasn't called, but probably should in case the updating is triggered.

examples/documentation.html
```diff
@@ -899,6 +899,26 @@ <h3 class="option"><span class="expand all">[ <a href="javascript:$('#optionlist
...
+            desc: "For sunburst only: you can alter the label output by providing a callback function. The function gets the data of the current arc as parameter. It expects you to return the plain text to be shown.",
+            default: "function(d){if(mode === 'count'){return d.name + ' #' + d.value}else{return d.name + ' ' + d.value}}",
+            examples: [function(d){ return d.name;}]
+        },
+        sort: {
+            desc: "For sunburst only: let's you specify the sort order of the arcs.",
+            default: "function(d1, d2){return d1.name > d2.name;}",
+            examples: [function(d1, d2){return d1.value > d2.value;}]
+        },
+        tooltipValueFormatter: {
```
liquidpele: You could already do this by simply doing chart.tooltip.valueFormatter(..), adding a duplicate way to do it here will cause confusion.

examples/documentation.html
```diff
@@ -899,6 +899,26 @@ <h3 class="option"><span class="expand all">[ <a href="javascript:$('#optionlist
             desc: "For sunburst only: specifies wether the leaf elements of a parent get the same color as the parent or if they are colored individually.",
             default: "true",
             examples: [true, false]
+        },
+        showLabels: {
+            desc: "For sunburst only: specifies if labels are displayed. The labels are centered on the arcs. The actual output can be adjusted by specifying the 'labelText' function.",
+            default: "false",
+            examples: [true, false]
+        },
+        labelText: {
```
liquidpele: DiscreteBarChart has a similar thing, called valueFormat:

https://nvd3-community.github.io/nvd3/examples/documentation.html#discreteBarChart

We should use the same option name if it does the same thing.

examples/documentation.html
```diff
@@ -899,6 +899,26 @@ <h3 class="option"><span class="expand all">[ <a href="javascript:$('#optionlist
             desc: "For sunburst only: specifies wether the leaf elements of a parent get the same color as the parent or if they are colored individually.",
             default: "true",
             examples: [true, false]
+        },
+        showLabels: {
```
liquidpele: discreteBarChart has showValues....   we should make options that do the same thing named the same.

examples/documentation.html
liquidpele: hmm... pieChart has labelFormat 

https://nvd3-community.github.io/nvd3/examples/documentation.html#pieChart

that's probably better actually, since this chart is more related to pieChart.

src/models/sunburst.js
liquidpele: Please don't comment it out

liquidpele:
This is a lot of changes, I'll have to try it out when I get a chance...

Crenshinibon:
I will incorporate your suggestions on monday.

examples/documentation.html
Crenshinibon: I will change it to the pieChart wording than. 'valueFormat' is working differently. You pass it a formatting string, e.g. ',.2f' (which isn't very descriptive, imho). Here, you pass it a function, which is called with the data and it expects that the function returns the formatted string. I will look at the PieChart implementation on monday, as well.

examples/documentation.html
Crenshinibon: I wasn't aware of that options. I will remove it.

src/models/sunburst.js
Crenshinibon: Should I better delete it than? 

I guess, someone with a better grasp of the whole update-matter of d3 should have a look at it ... the way it was implemented could have never worked. 

Or maybe I should have another look at the implementation of the original example where this chart was inspired from.

Crenshinibon:
Just to let you know, I removed some of the boilerplate code that was wrapping the chart in many 'g' elements. I tried to reduce it to the minimum amount necessary.

examples/documentation.html
liquidpele: valueFormat takes in a d3 formatter.... you would pass the result of d3.format(',.2f') to it.

src/models/sunburst.js
liquidpele: Maybe...  I originally tried to copy the example here when I made it, so the updating thing I think was more about changing the way the slices were sized.... 

http://bl.ocks.org/kerryrodden/477c1bfb081b783f80ad

examples/documentation.html
Crenshinibon: Just checked the pie ... labelFormat is marked as deprecated:

    // deprecated after 1.7.1
        labelFormat: {get: function(){ return valueFormat;}, set: function(_) {
            valueFormat=_;
            nv.deprecated('labelFormat','use valueFormat instead');
        }},)

And just maps to valueFormat. For my purpose I need some more fine-tuning for the label than just specifying the way the value should be displayed. Should I remove the whole option than or do you suggest another name to make the difference obvious.

liquidpele: What kind of fine tuning??   The labels are just a displayed text value...    if you're referring to the tooltips, you would use the contentGenerator to control the full tooltip content.

Crenshinibon:
After a good amount of learning I finally fixed the updating issue: https://github.com/novus/nvd3/issues/1317

I had to introduce a key-function-option that determines how to identify the data elements (which are new, updated, deleted). It defaults to 'name'. 

When the data is updated, the chart zooms out. This was the easiest way to circumvent weird behaviour when there were massive changes to the underlying data.

I rewrote a good amount of the chart and changed along the way the mode-option options. It's now count|value. That makes it more aligning to d3, since value is used there, everywhere.

I finally added a labelThreshold option. It works similar to pie/donut-chart.

src/models/sunburstChart.js
```diff
@@ -90,9 +76,9 @@ nv.models.sunburstChart = function() {
     //------------------------------------------------------------
 
     sunburst.dispatch.on('elementMouseover.tooltip', function(evt) {
-        evt['series'] = {
+        evt.series = {
             key: evt.data.name,
-            value: evt.data.size,
+            value: evt.data.value,
```
liquidpele: You changed this, but this will break charts for people already using the chart and expecting size to work.   please do this for backward compatiblity:

    value:  evt.data.value || evt.data.size,

liquidpele: And do the same elsewhere that value is used please :)

liquidpele:
Looks pretty good to me besides the one thing, looking forward to merging this to get update working!

Crenshinibon:
@liquidpele I hope, I will find tomorrow the time to fix everything up.

Crenshinibon:
I'm done so far, but have only checked quickly in my app.

canadian-thackeray:
Hi there, I am trying to use the labelFormat and labelThreshold options for the sunburst without any success. I am only getting "label value" ie "vis 432629", and the threshold doesn't seem to reduce the labels when you get to the smaller percentages. Any suggestions on what might be causing this?

PR #1344, opened 2015-11-06
Support log scale on lineChart (and scatter)

mgalloue:
- Reuse code badly removed (nv.utils.NaNtoZero)
- Support zero values on log scale.


liquidpele:
Neat, thanks for contributing!

PR #1351, opened 2015-11-10
Fix issue on Stroke width

mgalloue:
Stroke width property doesn't work on parallelCoordinate chart because variable is badly named, so I renamed it. Sorry, my bad.


liquidpele:

Also though, could you add info about that data option on the new wiki page for that? 

PR #1366, opened 2015-11-19
fix tooltip duplication bug with proper enter/update/exit pattern

mfurlend:
Updating the chart with new data while the tooltip is displayed would (sometimes) cause a new tooltip element to be created. My experience was specifically with a line chart. This fixes that issue.

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -243,18 +243,23 @@ nv.models.tooltip = function() {
 
     // Creates new tooltip container, or uses existing one on DOM.
     function initTooltip() {
-        if (!tooltip) {
+        if (!tooltip || !tooltip.node()) {
             var container = chartContainer ? chartContainer : document.body;
-
             // Create new tooltip div if it doesn't exist on DOM.
-            tooltip = d3.select(container).append("div")
-                .attr("class", "nvtooltip " + (classes ? classes : "xy-tooltip"))
-                .attr("id", id);
-            tooltip.style("top", 0).style("left", 0);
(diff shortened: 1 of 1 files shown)


adnasa:
You shouldn't check in files under the `build` folder.
Check the [README](https://github.com/novus/nvd3#a-few-rules-for-pull-requests)

liquidpele:
Yup please remove changes under build.

ThomasBrierley:
Hi @mfurlend, removing the build files will also change them. The purpose of not committing build files with source files is to keep the history clean and legible. You can achieve this by re-writing your first commit:

- Reset current branch to your first commit (9b8b3ad):
```
git reset 9b8b3ad --hard
```
- Checkout build directory from 1 commit before this:
```
git checkout HEAD~1 ./build
```
- Add build directory and amend commit:
```
git add ./build
git commit --amend --no-edit
```

This should result in a commit that doesn't touch the build directory but has your other changes. You will have to use the ```--force``` option when pushing because it will result in a different history, use this option carefully as it can be destructive.

If you mess up at any point, repeat the first step and it will get you back to your first commit (orphaned commits aren't actually deleted until git does garbage collection).

mfurlend:
Thanks Thomas - I wasn't sure how to go about doing that.

mfurlend:
Let me know if there is any outstanding issue with my implementation or with my commit and I'll do my best to resolve it.

liquidpele:
Looks okay to me, thanks!

PR #1367, opened 2015-11-20
Expose lineChart's state

flood4life:
I had to modify lineChart's state manually and couldn't access it, though I could access state of other chart models. I looked up the sources and found that instead of exposing lineChart's state there's just blank line.


liquidpele:
Oh, wow, wonder how that got missed.  Thanks for contributing!

PR #1368, opened 2015-11-21
Added useInteractiveGuideline option to multiBarChart

Bajena:
I needed to useInteractiveGuideline with multiBarChart but it was missing. I think it might be useful for some people.


liquidpele:
Seems to be doing the same thing as https://github.com/novus/nvd3/pull/1329

You guys might want to look over each other's pull requests... I notice there are some big differences.

liquidpele:
Thanks for contributing!

Issue #1369, opened 2015-11-23
Assign dispatch-events from parent down to the inheriting model


adnasa:
Allow models such as pieChart to inherit dispatching events from model they inherit from, just the same way as options are inherited through `inheritOptions`.

##### Motivation

Since `nv.models.pieChart` is a model that extends `nv.model.pie`, it would make sense to use the same registered event namespace as well from the slices created by the model.

```
	// nv.models.pieChart() extends nv.models.pie() under the hood.
	var chart = nv.models.pieChart()
		.x(function(d) { return d.key })
		.y(function(d) { return d.y })
        .width(width)
        .height(height);

	// listen to the same event dispatched in nv.models.pie() slices
	// with our extended chart model.
	chart.dispatch.on('elementClick', function(payload) {
		// do something when the user clicked on a slice.
	})

```

The listener callback above yeilds the following problem:

```
Uncaught TypeError: Cannot read property 'on' of undefined
```

##### Addressing this manually

A workaround for this case was the following code (which was in no way an elegant solution) before working out a commit:

```
	// nv.models.pieChart() extends nv.models.pie() under the hood.
	var chart = nv.models.pieChart()
		.x(function(d) { return d.key })
		.y(function(d) { return d.y })
        .width(width)
        .height(height);
        
    // manually re-assign the dispatch property for nv.models.pie.
	chart.dispatch = _.extend(chart.dispatch, chart.pie.dispatch);

```

##### Solution

In my commit (PR coming!), I focused my solution specifically for the __pieChart__ model.  
What I have done here was *assigning* the same callbacks registered in the dispatchers of the parent model straight down to the model that inherits it.

- Extension signature: `nv.utils.inheritDispatchers()`

----

If you have ideas/feedbacks on how this can get be done more elegantly, let me know.
Kudos to the good work!

adnasa:
Closing this issue since it is no longer relevant

PR #1370, opened 2015-11-23
Issue 1369: Assigned dispatching events from nv.models.pie to nv.mode…

adnasa:
Addresses the issue described in #1369

### DIFF
--- a/src/models/pieChart.js
+++ b/src/models/pieChart.js
@@ -229,6 +229,7 @@ nv.models.pieChart = function() {
             margin.left   = _.left   !== undefined ? _.left   : margin.left;
         }}
     });
+    nv.utils.inheritDispatchers(chart, pie);
     nv.utils.inheritOptions(chart, pie);
     nv.utils.initOptions(chart);
     return chart;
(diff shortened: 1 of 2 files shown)


liquidpele:
unit tests failed...  also I'm not sure blindly inheriting all dispatch event is a good idea... I'll need to research that concept some...

adnasa:
@liquidpele 
hmm.. I tested it locally before submitting the patch.

as for the inheritance; the thing here is that slices used in the pieChart model are created in pie.
In some way the *same* instance of the event dispatcher has to be shared between the two.

let me know when you got your "aha!" moment and let's discuss the implementation!

liquidpele:
I mean, as a convenience you could do that, but you can already access them by chart.pie.dispatch too...

adnasa:
I see... oh well.
It took a while to find the dispatching events, so I'm going to go ahead and update the [code example](https://github.com/nvd3-community/nvd3/blob/gh-pages/examples/pieChart.html) and have it there in the mean time for someone else's sake.

PR #1371, opened 2015-11-24
Fix a few issues

mgalloue:
Typo: Fix issue on brush badly resizing in a particular case (with only NAN values on a dimension).
Typo: Fix issue on drag axis (available width doesn't exit, use width instead)
Update availableHeight and availableWidth function to avoid negative values (use zero instead).

### DIFF
--- a/src/models/parallelCoordinates.js
+++ b/src/models/parallelCoordinates.js
@@ -176,10 +176,10 @@ nv.models.parallelCoordinates = function() {
             foreground.exit().remove();
             foreground.attr('d', path)
                 .style("stroke-width", function (d, i) {
-                if (isNaN(d.stroke)) { d.stroke = 1;} return d.stroke;})
+                if (isNaN(d.strokeWidth)) { d.strokeWidth = 1;} return d.strokeWidth;})
                 .attr('stroke', function (d, i) { return d.color || color(d, i); });
             foreground.on("mouseover", function (d, i) {
-                d3.select(this).classed('hover', true).style("stroke-width", d.stroke + 2 + "px").style("stroke-opacity", 1);;
+                d3.select(this).classed('hover', true).style("stroke-width", d.strokeWidth + 2 + "px").style("stroke-opacity", 1);;
                 dispatch.elementMouseover({
                     label: d.name,
                     color: d.color || color(d, i)
@@ -187,7 +187,7 @@ nv.models.parallelCoordinates = function() {
 
             });
             foreground.on("mouseout", function (d, i) {
-                d3.select(this).classed('hover', false).style("stroke-width", d.stroke + "px").style("stroke-opacity", 0.7);
+                d3.select(this).classed('hover', false).style("stroke-width", d.strokeWidth + "px").style("stroke-opacity", 0.7);
                 dispatch.elementMouseout({
                     label: d.name,
                     index: i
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -598,14 +598,14 @@ nv.utils.sanitizeWidth = function(width, container) {
 Calculate the available height for a chart.
 */
 nv.utils.availableHeight = function(height, container, margin) {
-    return nv.utils.sanitizeHeight(height, container) - margin.top - margin.bottom;
+    return Math.max(0,nv.utils.sanitizeHeight(height, container) - margin.top - margin.bottom);
```
liquidpele: why is this necessary?

src/models/parallelCoordinates.js
```diff
@@ -366,7 +377,7 @@ nv.models.parallelCoordinates = function() {
             }
 
             function dragMove(d) {
-                dragging[d.key] = Math.min(availableWidth, Math.max(0, this.parentNode.__origin__ += d3.event.x));
+                dragging[d.key] = Math.min(width, Math.max(0, this.parentNode.__origin__ += d3.event.x));
```
liquidpele: width is only set if they use that option, otherwise it's null, why would you want this change?

src/models/parallelCoordinates.js
```diff
@@ -100,7 +100,7 @@ nv.models.parallelCoordinates = function() {
                         max = y[d].domain()[1];
                     }
                         //If the brush extent is > max (< min), keep the extent value.
-                    else if (!f[0].hasNaN && displayBrush) {
+                    else if (!f[0].hasOnlyNaN && displayBrush) {
```
liquidpele: What is hasOnlyNaN?   Does this replace hasNaN, or a supplement to it?

src/utils.js
mgalloue: Because some time sanitizeWidth return zero and with the margin, availableWidth could return a negative value giving js error. So if sanitizeWidth is zero then available width too.

src/models/parallelCoordinates.js
mgalloue: HasOnlyNan is already existing, in fact when I commit this code in a first place I made a typo on setting hasNaN, here we need to check hasOnlyNaN. This check if there is only nan values on the brush (so the brush is below the dimension axis, and on the undefined values line).

src/models/parallelCoordinates.js
mgalloue: On my case it doesn't work with availableWidth and works well with width but your comment make sense. I will try to understand why so I removed this from the PR for now

src/tooltip.js
```diff
@@ -47,8 +47,8 @@ nv.models.tooltip = function() {
      */
     var position = function() {
         return {
-            left: d3.event.clientX,
-            top: d3.event.clientY
+            left: d3.event !== null ? d3.event.clientX : 0,
+            top: d3.event !== null ? d3.event.clientY : 0
```
mgalloue: This is needed because when data are updated, if the tooltip is currently display, it can stay visible even if the mouse is not on the point anymore, because no mouse move event is detected. Later if I go out of the chart the tooltip is still displaying (and never hide until going on another point). To avoid that I called chart.tooltip.hidden(true) at each update, but d3.event is null in this case.

Issue #80, opened 2013-02-26
[Stagger labels] stagger label flag works when setting it on but doesn't turn off


rmleon:
I am redrawing a chart.  Doing "staggerLabel(true)" works, but when I try reversing it  "staggerLabel(false)", the chart doesn't pick up the change.

I saw this on the discrete bar chart, but it probably happens across all charts.

britvik:
Hello, I encountered the same issue in **multiBarChart**. Why is this closed? Just by negligently looking at the source code I can see that there is no **else** branch when processing staggerLabel. If we set staggerLabel to true, texts get translated on y dimension but if we set it back to false it is never set back to 0,0.

Just adding
```javascript
else {
 xTicks.selectAll("text").attr('transform', "translate(0,0)");
}
```
after
```javascript
if (staggerLabels) {
...
}
```
fixed my issue.

Could someone look into this? Thanks :-)

liquidpele:
This was closed before I started working with nvd3, but glancing at it I agree it looks to be an issue so I'll reopen this.   If you send us a pull request with a fix then we can merge it in.

dontbesorry80:
This still seems to be a problem on the multiBarChart? Any chance of a fix?

PR #1372, opened 2015-11-25
Fixed turning off staggerLabels after they have been previously turned on

britvik:
If you set staggerLabels to true and decide to change it to false later, chart won't update the labels and they remain staggered forever.

This should close issue https://github.com/novus/nvd3/issues/80


britvik:
I moved staggerLabels processing to else branch of rotatingLabels. Because staggering doesn't work anyway when labels are rotated and it also makes sense (and it passes the test now) :).

liquidpele:

PR #1376, opened 2015-11-26
Update pie chart example

adnasa:


liquidpele:
Cool, I like the idea of adding some doc into the examples :)

adnasa:
It was already there. Just added more :+1: 
have you thought about the event inheritance further?

liquidpele:
Yea, I don't think the inheritance will be good, because some charts use multiple graph types (like linePluBarChart, mutliChart, etc) while others use multiple chart types (line chart uses scatter, stacked uses lines with in turn uses scatter), and in many cases a higher level catches an event and modifies it... especially with tooltips.  Basically, I think it would cause a lot of problems and bugs, when you can already control things granularly as it is.

PR #1378, opened 2015-11-27
Optimise scatter model

ThomasBrierley:
We've been using nvd3.js at Learning Science Ltd. for plotting continuously generated data, and found the performance of updating a chart to reduce dramatically as the total number of data points increased.

Through some profiling I attributed this to the scatter model (used by most models). The performance is proportional to the total number of nodes because of some unnecessary operations to the D3 update selection.

I have carefully removed or moved operations that do not need to be called on existing nodes and culled the remaining ones by adding some caching (e.g updating position). I've tested the affect of each where possible.

This passes the unit-tests and I have done a side by side comparison of the examples including interactivity. My performance tests show vastly reduced overhead in maintaining existing nodes on redraw.

Resizing a chart will still be fairly slow because it requires all node positions to be recalculated. The impact of this can be reduced further by setting axis range manually and setting transition durations to zero:

``` JavaScript
var chart = nv.models.lineChart();
chart.duration(0);
for (var key in chart) {
    if (typeof chart[key].duration === 'function') {
        chart[key].duration(0);
    }
}
```

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -374,6 +374,7 @@ nv.models.scatter = function() {
                 .attr('class', function(d,i) {
                     return (d.classed || '') + ' nv-group nv-series-' + i;
                 })
+                .classed('nv-noninteractive', !interactive)
                 .classed('hover', function(d) { return d.hover });
             groups.watchTransition(renderWatch, 'scatter: groups')
                 .style('fill', function(d,i) { return color(d, i) })
@@ -414,7 +415,6 @@ nv.models.scatter = function() {
                 d3.select(this)
                     .classed('nv-point', true)
                     .classed('nv-point-' + d[1], true)
-                    .classed('nv-noninteractive', !interactive)
                 ;
             });
             points
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -415,7 +415,6 @@ nv.models.scatter = function() {
                     .classed('nv-point', true)
                     .classed('nv-point-' + d[1], true)
                     .classed('nv-noninteractive', !interactive)
-                    .classed('hover',false)
                 ;
             });
             points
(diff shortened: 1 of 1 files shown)


ThomasBrierley:
Hi @liquidpele 

Is there anything I can do to help get this accepted?

liquidpele:
Yup, sorry... I just cut the 1.8.2 release and didn't want to merge in any big changes until that... just took way longer for me to have some free time.   This looks okay to me, so lets merge and if there are any bugs then I'll flag you down ;)

ThomasBrierley:
Makes sense, it's especially easy cause regressions with optimisations however much care is taken. I'm fairly confident with these because I only went after the low hanging fruit - but you never know.

Thanks for all your hard work @liquidpele there have been loads of PRs to take care of over the last couple of months :)

mgalloue:
Hi @liquidpele and @ThomasBrierley, using the last release of NVD3 brokes my linechart and it works well without this PR. The issue is (after resizing my chart) when I try to highlight a point on a line, the point is not displayed at the good place (seems that it keeps the initial position).   
Thanks,

ThomasBrierley:
Hi @mgalloue, sorry to hear that. I will attempt to reproduce this but it would be very helpful if you could provide a test case.

Resizing the chart should break through all pruning mechanisms. There is a ```scaleDiff``` condition that should be true if the chart is resized, this is resolved each update and is one of the things used to determine if point positions need updating.

ThomasBrierley:
I've reproduced it in the following example, the scaleDiff bool is not true when the chart height changes: https://jsfiddle.net/99jbqtxL/

However oddly enough when a minified d3.js is used with the same example the problem is not present: https://jsfiddle.net/99jbqtxL/1/

The scaleDiff bool uses the x y z functions from d3.scale.linear ... My understanding was that these should also be different if any of the chart dimensions were to change but I may well be wrong. In my example y appears to be zero whatever I do with the chart.

I will add an additional check based on the actual chart dimensions.

liquidpele:
merged, @mgalloue please verify if that resolves your issues.

mgalloue:
Thanks a lot for your efficiency @liquidpele and @ThomasBrierley, it works well!

PR #1389, opened 2015-12-04
Require D3 for browserify support

m-h-t:


src/core.js
```diff
@@ -11,6 +11,11 @@ nv.charts = {}; //stores all the ready to use charts
 nv.logs = {}; //stores some statistics and potential error messages
 nv.dom = {}; //DOM manipulation functions
 
+// Node/CommonJS - require D3
+if (typeof(module) !== 'undefined' && typeof(exports) !== 'undefined' && typeof(d3) == 'undefined') {
+    d3 = require('d3');
+}
+
```
liquidpele: I'm not sure this is right... it's not ensuring that nvd3 loads after d3 is loaded...

m-h-t: You're talking about the order in which it is loaded? Or whether it's loaded at all? In which case do you think there is going to be a problem?

mqklin: @m-h-t , the problem is that `nvd3` requires `d3` to be in global space. It's hurt, because global namespacing is a bad pattern.

mqklin:
Can anyone review this PR?

mqklin:
Can anyone review this PR?

mqklin:
I think it should be support not only for UMD, but also for AMD. I will try to make it.

readcodelearn:
I can confirm that this PR works for browserify

topicus:
I can confirm it works for browserfy.

liquidpele:
Okay, well if it works for ya'll then I'll merge it in, thanks for confirming it :)

Issue #1263, opened 2015-09-25
Tooltip doesn't hide when changing tab


slestang:
When some tooltip are displayed and you switch to another tab (with Ctrl +T for example), if you switch back to the previous tab, the tooltip are still displayed despite of the cursor not hovering the graph anymore. The *mouseout* event is emitted but in *interactiveLayer.js* *layer()* *mouseHandler()* the position of the mouse returned by *d3.mouse(this)* is the old one so the tooltip isn't hidden.

May be we should modify this *if* statement https://github.com/novus/nvd3/blob/master/src/interactiveLayer.js#L86 by adding a test on the event type, e.g:
```javascript
                if (d3.event.type === 'mouseout'
                    || mouseX < 0 || mouseY < 0
                    || mouseX > availableWidth || mouseY > availableHeight
                    || (d3.event.relatedTarget && d3.event.relatedTarget.ownerSVGElement === undefined)
                    || mouseOutAnyReason
                    ) {
  ```

liquidpele:
Feel free to implement a fix and test it, and then send us a pull request :)

gb96:
Implemented fix proposed by @slestang, see pull request [https://github.com/novus/nvd3/pull/1394]

PR #1394, opened 2015-12-08
Bug fix for #1263 Tooltip doesn't hide

gb96:
Applied and tested fix as proposed by @slestang
Tested under Google Chrome 47.0.2526.73 m (64-bit) / Windows 7 (x64).
Tooltips now hide after switching browser tabs (as soon as mouse focus returns to the tab containing the chart)


slestang:
Thank @gb96, I should have done that a long time ago.

liquidpele:
sweet, thanks!

PR #1395, opened 2015-12-10
Extended object attribute access for Box Plot model.

jjzazuet:
- Extending Box Plot model to support attribute accessor callbacks for custom models.
- Model remains backwards compatible with original data structure implementation (e.g. `label, values, outliers`, etc.). `y` chart option now emits a deprecation warning to the developer.
- Adding support for outlier elements as objects including label, color and value.
- Updated Box plot test case page.
- Cleaned up JSHint linter warnings for Box plot model.
- Updated examples page with existing and new example entry.

### DIFF
--- a/examples/boxPlot.html
+++ b/examples/boxPlot.html
@@ -5,7 +5,7 @@
 
     <script src="https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.2/d3.min.js" charset="utf-8"></script>
     <script src="../build/nv.d3.js"></script>
-    
+
     <style>
         text {
             font: 12px sans-serif;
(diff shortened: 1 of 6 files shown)


billderose:
there are some changes in here I'd like to see as well. @jjzazuet, is it possible to include in this change the ability for user defined ticks. [This line](https://github.com/novus/nvd3/pull/1395/files#diff-1324737ebf62c585c9b05589372c802fR143) always uses the height to determine the number of ticks but it seems like we could add that option here.

jjzazuet:
@billderose ok so let me see if I understood correctly. Instead of determining the tick size with a fixed value:

    .ticks( Math.floor(availableHeight/36) )

We could instead say:

    .ticks( Math.floor(availableHeight/someCustomTickSizeOrDefaultToFixedValue) )

Is this about right? Thanks!

billderose:
@jjzazuet that's along the right lines. I'm wondering if it would be best to do something like `ticks(customTicks)` which would allow any number of ticks and would not require the user to guess and check their `customTicks` value. We can always fallback to the `.ticks( Math.floor(availableHeight/36) )` in the absence of a used defined value.

jjzazuet:
@liquidpele , any chances of merging this one before I try further changes? Thanks for your time and help!

liquidpele:
Oh wow, that's a lot of change...  @Turgar you want to take a look at this since you were the original boxplot contributor?

Turgar:
@liquidpele New code LGTM -- I'll try it out soon. Thanks!

Turgar:
@liquidpele No objections here if you want to merge.

liquidpele:
Cool, merging in then.

jjzazuet:
@liquidpele @Turgar Thanks! ;)

PR #1399, opened 2015-12-14
more data for more flexible custom tooltip

torstenloos:
For the purpose of having more data in the contentGenerator to make even more customizable tooltips in stackedAreaChart.

Would also be great in other charts.

### DIFF
--- a/src/models/stackedAreaChart.js
+++ b/src/models/stackedAreaChart.js
@@ -367,7 +367,7 @@ nv.models.stackedAreaChart = function() {
                             key: series.key,
                             value: tooltipValue,
                             color: color(series,series.seriesIndex),
-                            stackedValue: point
+                            point: point
                         });
 
                         if (showTotalInTooltip && stacked.style() != 'expand') {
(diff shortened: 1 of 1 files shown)


britvik:
Was about to create a pull request for the same thing =). +1 for merge. Maybe we should consider renaming the property though (like 'point' or something...).

liquidpele:
Yes, agreed on renaming it, please do that and I'll merge it.

torstenloos:
About the renaming, do you mean that stackedValue should be renamed to point ? So that it is like:

point: point

liquidpele:
Yes.  It's named that for the tooltips already if you turn off interactive guideline:  

https://github.com/novus/nvd3/blob/master/src/models/scatter.js#L303

torstenloos:
Thanks for the fast response! I would like to do the same for linechart, can I do that in this pull request or should I make a seperate one?

liquidpele:
either way, it's up to you.

torstenloos:
I just saw that it must have been recently added to lineChart:
https://github.com/novus/nvd3/blob/master/src/models/lineChart.js#L402

But there it is named:
data: point

liquidpele:
Awesome, thanks!

nathan-lee:
@liquidpele Looks like this has broken https://github.com/novus/nvd3/blob/master/build/nv.d3.js#L13509

Two lines should be updated to use series.point.display.y0 and series.point.display.y

PR #1408, opened 2015-12-28
pieChart.js: fixed issue with regards to width and height not being used. (Scaling issues)

pieter-vincken-agilians:
The available width and height were determined by using the width and height of the parent element. This is incorrect behaviour if the width and height are explicitly set. Width and height are inherited properties are should be accessed by referencing pie instead of use width and height directly (which would result in undefined).

Pull request fixes the issue by referencing the pie width and height instead of the pieChart width and height.

### DIFF
--- a/src/models/pieChart.js
+++ b/src/models/pieChart.js
@@ -66,8 +66,8 @@ nv.models.pieChart = function() {
             nv.utils.initSVG(container);
 
             var that = this;
-            var availableWidth = nv.utils.availableWidth(width, container, margin),
-                availableHeight = nv.utils.availableHeight(height, container, margin);
+            var availableWidth = nv.utils.availableWidth(pie.width(), container, margin),
+                availableHeight = nv.utils.availableHeight(pie.height(), container, margin);
 
             chart.update = function() { container.transition().call(chart); };
             chart.container = this;


liquidpele:
Oh... I like this better than the other fix.

src/models/pieChart.js
```diff
@@ -66,8 +66,8 @@ nv.models.pieChart = function() {
             nv.utils.initSVG(container);
 
             var that = this;
-            var availableWidth = nv.utils.availableWidth(width, container, margin),
-                availableHeight = nv.utils.availableHeight(height, container, margin);
+            var availableWidth = nv.utils.availableWidth(pie.width(), container, margin),
+                availableHeight = nv.utils.availableHeight(pie.height(), container, margin);
```
mgalloue: @liquidpele I still have the sizing issue with that fix. Why it's a better fix than the other? I saw that every charts have width and height added in there option, pie chart is the only one missing...

liquidpele: oh really?  that's strange...   pieChart should inherit those options from pie so I thought that should have worked...    regardless if you still have the issue then please submit your fix again and I'll merge it.

mgalloue:
Thanks, I have submitted the PR [#1461](https://github.com/novus/nvd3/pull/1461)

liquidpele:
Reverting this due to https://github.com/novus/nvd3/issues/1481

PR #1410, opened 2015-12-29
create option for customizing donut charts inner and outer radius

jacobscarter:
Unless I am mistaken I didn't see an easy way to do this in your library.  Let me know if the tests and examples I made work for you.  Thanks.

### DIFF
--- a/examples/donutChart.html
+++ b/examples/donutChart.html
@@ -76,6 +76,8 @@
             .height(height)
             .padAngle(.08)
             .cornerRadius(5)
+            .arcInnerRadius(100)
+            .arcOuterRadius(120)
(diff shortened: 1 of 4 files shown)


mgalloue:
Please look at monitoring chart example [here](https://github.com/nvd3-community/nvd3/blob/gh-pages/examples/monitoringChart.html) and source code [here](https://github.com/nvd3-community/nvd3/blob/gh-pages/examples/monitoringChart.html). I think it do what you want.

liquidpele:
Yup, this is already possible.

jacobscarter:
Awesome thanks.

PR #1414, opened 2015-12-30
Model for force directed graph

c-martinez:
As discussed in issue https://github.com/nvd3-community/nvd3/issues/40, I'm not an expert in D3, but I think this works.

Not sure whether to send PR here or nvd3-community so apologies if this is not the right place to send this PR.


liquidpele:
Right place, thanks!

liquidpele:
Well... it's neat as a d3 demo, but what can it be used for as far as making data accessible?   Without having a tooltip or something that lets you dig down into the points, it's just a bunch of circles and lines...

c-martinez:
Yup, agree. But it is tricky to say before hand what is the best way to dig down -- tooltips (name of each data point)? colours (class of each data point)? Size (importance of the data point)?

That's why it is 'open ended' -- the example `forceDirected.html`, uses `nodeExtras` to add tool tips. But in principle, you could add other properties to the node. There is also a `linkExtras`, which allows to modify the links themselves.

```
 .nodeExtras(function(node) {
                  node
                    .append("text")
                    .attr("dx", 12)
                    .attr("dy", ".35em")
                    .text(function(d) { return d.name });
                });
```

liquidpele:
Well, the problem with that is that we have a whole tooltip object that is re-used everywhere, so trying to allow the invention of some custom thing that is really going to cause confusion.  You should try to use the tooltip.js that all the other charts are using.

c-martinez:
Good point. I wasn't aware of `tooltip.js`, I'll make changes to use it. Thanks!

c-martinez:
Sorry for the delay -- it took me a bit to get my head around tooltips.

Small issue -- force directed graph's do not work with Series (at least not explicitly) so this code makes a series on the flight, just for the tooltip. Not sure it's the best way, but it works. The tooltip displays all the content in the node (in the les miserables example, name and group).

liquidpele:
Hmm, github seems to think there are conflicts, you may need to merge in the lastest nvd3 master and fix any conflicts it creates in your branch and then push that back up.

c-martinez:
Yes, I was quite far behind from master. No conflicts now.

c-martinez:
Hi @liquidpele -- any chance this may get merged any time soon?

liquidpele:
cool

c-martinez:
Great. Thanks!

PR #1422, opened 2016-01-05
Bugfixes

mgalloue:
- Avoid ticks overlapping (needed especially for log scale).
- Do not display the legend if there isn't enough space (line chart)


src/models/lineChart.js
```diff
@@ -184,7 +184,19 @@ nv.models.lineChart = function() {
...
+                if (availableHeight1 === 0) {
+                    g.select('.nv-legendWrap')
+                        .datum([])
+                        .call(legend);
+
+                    if (margin.top != legend.height()) {
+                        margin.top = legend.height();
+                        availableHeight1 = nv.utils.availableHeight(height, container, margin) - (focusEnable ? focusHeight : 0);
+                    }
+                }
```
liquidpele: This seems like a good idea, but I wonder if there is a way to make this more generic so it can be re-used for other charts too...

mgalloue: @liquidpele ,There is not really an easy way to do that because it need many arguments but I guess we can "duplicate" this code for other charts.

PR #1246, opened 2015-09-09
[discreteBarChart,multibarChart] Add wrapLabels parameter.

ovvn:

PR #1423, opened 2016-01-05
Add wrapLabels parameter description.

ovvn:
Adds description of the parameter that was added by me in the PR https://github.com/novus/nvd3/pull/1246


examples/documentation.html
```diff
@@ -327,7 +327,7 @@
 
         <div style="height: 50px"></div>
 
-		<h3 class="option"><span class="expand all">[ <a href="javascript:$('#optionlist>ul').toggle()">toggle all</a> ]</span></h3>
+    <h3 class="option"><span class="expand all">[ <a href="javascript:$('#optionlist>ul').toggle()">toggle all</a> ]</span></h3>
```
adnasa: Looks to me like an unecessary commit and that this change messed up the indentation.

ovvn: The commit is not about indentation, it rather adds a useful api information to the documentation. But even if we take a closer look to that indentation at https://github.com/novus/nvd3/blob/master/examples/documentation.html#L330. It obviously looks incorrect, as it's not nested and being placed between two closed `div`s.

adnasa: I don't follow you when you say _...it rather adds a useful api information to the documentation..._ unless you meant that it toggles the _display of the api information_

but taking a look again at [your change](https://github.com/dmapper/nvd3/blob/master/examples/documentation.html#L330), I ask myself if I am totally blind and missed something that made you want to change this file more than "fixing the indentation". Because as it looks to me right now in your change of this file, it didn't make the indentation any better even if the [previous line in master branch](https://github.com/novus/nvd3/blob/master/examples/documentation.html#L330) *obviously looks incorrect*, as you say.

ovvn: Well, the commit is obviously about the `wrapLabels` parameter as it's clear from the PR description. Changing indentation appeared here just by accident and turned out to be right thing.

adnasa: oh well... [very right indeed](http://take.ms/i8RmN)

adnasa:
@ovvn :+1: 

---

Final note: Yes, I agree with you that this has been quiet annoying, but I really **wasn't** trying to block you. I want this merge to happen just as much as you. The reason I persist is that my principle for a qualified PR (for now) is:

1. Focus on the changes on code/feature/fix that fullfils the intent of the PR
2. Everything else is irrelevant (save things like indentation fixes for another PR, this helps reviewers to have a great overview and understand point 1 in this list)

--- 

good night, man :wink:

liquidpele:
This PR has the build files included, please remove them and fix any merge conflicts.  Thanks.

PR #1426, opened 2016-01-06
Fixed some bugs with the switchYAxisOrder option in linePlusBarChart

klangfarbe:
1. The x axis scale was broken in case the option was initially enabled
2. The tooltip formatting was broken and used the wrong tickFormat
3. The horizontal rules for y2 axis were not drawn in case the option is enabled and the lines had no values


liquidpele:
Cool, thanks!

Issue #1427, opened 2016-01-06
Hidden tooltip obscures anything below it


elesueur:
Because the mechanism used to hide the tooltip is opacity: 0 [1] rather than display: none, anything under the tooltip is unreachable even when the tooltip is hidden, for example, a drag handle. Hiding the tooltip should really use display: none. At the very least, a hidden tooltip should have a class, e.g. nvtooltip-hidden to allow the CSS to be overridden.

[1] https://github.com/novus/nvd3/blob/master/src/tooltip.js#L215

PR #1428, opened 2016-01-06
set display: none when hiding tooltip

elesueur:
This solves issue https://github.com/novus/nvd3/issues/1427


liquidpele:
Won't this not display the opacity transition?   I wouldn't think that it could interpolate display.   

Can you show a demo of it working?

liquidpele:
Closing this due to no response...

Issue #1434, opened 2016-01-12
Default svg.nvd3-svg width and height styles override chart.width() and chart.height() attributes 


mqklin:
From `main.css`:

    /********************
      Default CSS for an svg element nvd3 used
    */
    svg.nvd3-svg {
        ...
        width:100%;
        height:100%;
    }

And when I try to make this: `chart.height("77em")`, it sets `height` attribute to `svg` element, but classes have higher priority then `[attributes]`!

![bug](https://cloud.githubusercontent.com/assets/9456433/12258279/475811a2-b92f-11e5-9cbb-63adb7090dcb.png)

mgalloue:
You can use` height: 77em !important`, to change priority.

mqklin:
@mgalloue, had you tried it before you wrote this? Of course it's not working because `width` in this case in an `attribute`, not `style`.

mgalloue:
Oops, sorry yes I was speaking about style obviously. For all my charts I'm using chart.width() for sizing and it's works well, don't know what is the difference in you case.

mqklin:
@mgalloue, it works because of incorrect `Chrome` or `Firefox` behaviour. It's not working in `IE`. `chart.width()` sets `width` as attribute and it shouldn't rewrite `width` as style (see screen above).

mqklin:
Fix PR: https://github.com/novus/nvd3/pull/1435

PR #1435, opened 2016-01-12
fix issue #1434

mqklin:
It may be a breaking change, but I think it's better than having a bug.

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -13,8 +13,6 @@ svg.nvd3-svg {
     -moz-user-select: none;
     user-select: none;
     display: block;
-    width:100%;
-    height:100%;
 }
 
 /********************


mgalloue:
+1

mgalloue:
Hi @mqklin which chart are you using? Can you check if the fix done in [this](https://github.com/novus/nvd3/pull/1450) PR can fix your problem?

Regards

mqklin:
I use multibarchart and horizontal multibarchart, I'll check it tomorrow

liquidpele:
No, I specifically added the width and height so that it would by default take up the whole svg... this allows the svg to scale with the page easily without having to calculate height and width every time.

liquidpele:
I merged the height/width options that @mgalloue added...   if you use those then it should override the default CSS... unless that part is broken :/

mqklin:
@liquidpele , but look at this [screen](https://cloud.githubusercontent.com/assets/9456433/12258279/475811a2-b92f-11e5-9cbb-63adb7090dcb.png)! Styles from `.css` file have bigger priority than `svg[Attributes Style]`!
Look at [jsfiddle example](https://jsfiddle.net/9jn04u1d/)! 
If you want to specify `width` and `height` as `100%` you should do it through `svg[Attributes Style]` as well!

mqklin:
@liquidpele , it is possible to set `chart.width` with not only unitless? e.x. `chart.width("10em")` ?

PR #1440, opened 2016-01-12
fix multiChart.js tooltip

montrezorro:
Call to headerFormatter() is missing and tolltip header isn't formatted with the function xAxis.tickFormat()


liquidpele:

PR #1442, opened 2016-01-12
Enhancement to scatter.js: Add showLabels option to ScatterChart to show text with bubble

rinand:
Add an option to display text as a label of each shape in the scatter chart. 

```js
var chart = nv.models.scatterChart()
  .showDistX(true)
  .showDistY(true)
  .color(d3.scale.category10().range())
  .showLabels(true);

d3.select('#text svg')
  .datum(randomData(4, 10))
  .transition().duration(500)
  .attr('height', 500.0)
  .call(chart);

//  series-data should have attribute to label.
function randomData(groups, points) {
  var data = [],
    random = d3.random.normal();
  for (i = 0; i < groups; i++) {
    data.push({
      key: 'Group ' + i,
      values: []
    });
    for (j = 0; j < points; j++) {
      data[i].values.push({
        x: random(),
        y: random(),
        size: Math.round(Math.random() * 100) / 100,
        shape: 'circle',
        label: 'circle-' + j
      });
    }
  }
  return data;
}
```


liquidpele:
Please remove build files from any pull requests, thanks.

liquidpele:
Also, a demo of this would be great :)

rinand:
example has been added to examples/scatterChart.html and build files removed. :)

liquidpele:
Hmm... unit tests failed, but I don't see how they could have from that last commit, so maybe just a fluke.  

Sorry, by an example I meant like could you post a link to a chart showing your example, or just a screenshot of what it looks like?

rinand:
no clue why unit tests failed -  :( anyhow, I removed updates on example file and here is a screenshot.

scatterchart_with_label

And, thank **@liquidpele** for the maintaining awesome library.

liquidpele:
Sometimes travis CI just bombs... it's about the 8th time I've seen it do that.  

This looks pretty good...  thanks for contributing!   Please also document the 'label' data attribute that it uses on the wiki page for that:

PR #1443, opened 2016-01-13
Fix generation of lineChart

regit:
In the case of a line charts containing multiple lines, the color
of the line was white resulting in a looking non displayed graph. This
patch fixes the issue by using an approach similar to the one
used in multibarChart.


liquidpele:
I'm confused at this.. how would the line ever display as white?  Can you show a demo in a jsfiddle?

liquidpele:
Closing this due to no response.

Issue #1425, opened 2016-01-06
multiBarHorizontalChart has a bug related with forceY


jongmoon:
Hi, I'm very thankful for nvd3 these days.

But I had several problems using nvd3. One of those is forceY bug of multiBarHorizontalChart.
Although I know '0 base' is majority in barchart,  I need to change base value more than 0.

However it is not working correctly.
I made an [example](http://codepen.io/jongmoon/pen/OMpzoy). And as this example shows.. bar chart start from incorrect position.

![image](https://cloud.githubusercontent.com/assets/3255471/12142646/113b96e6-b4bd-11e5-837f-df90e5ce46ce.png)

I could find two solutions.

**1. Change the min value of domain.**

There are codes to get min range value in ```nv.models.multiBarHorizontal``` like below. 

*Bug*
```javascript
y(0)
//eg. y0 = y0 || d3.scale.linear().domain(y.domain()).range([y(0),y(0)]);
```
But forceY is forced on chart. eg. chart.forceY(10), then return value y(0) is no more '0' (min range value expected) in ```nv.models.multiBarHorizontal```. So I could change these code like below and this could solved the problem.

*Solution*
```js
var yDomainMin = y.domain()[0] || 0;
y(yDomainMin);
//eg. y0 = y0 || d3.scale.linear().domain(y.domain()).range([y(yDomainMin),y(yDomainMin)]);
```
**2. Make clip rect**

There's clip rect in ```nv.models.multiBar```. This models also have a same issue. But It does not make problem. because it has a clip rect.

![image](https://cloud.githubusercontent.com/assets/3255471/12143102/1a230448-b4c1-11e5-84e2-b9b2cd141f78.png)

So If clip rect is applied on ```nv.models.multiBarHorizontal```, It can also solve problem.

If you give me a chance to PR, I would be my pleasure.
Although If you dont, I will be happy if my job is helpful to you.

mgalloue:
Hi, 
You can just add `chart.yScale().clamp(true);`
I tried it in your example and it works fine.

Regards,

jongmoon:
Thank you very much for your response. 

I didn't consider clamp(), It works fine as you commented.
I learned a lot thanks to you.

By the way,
I think it would be more helpful if ```nv.models.multiBarHorizontal``` clamps y-scale internally in this case although forceY is not normal in this chart.

I wonder what do you think about it.

And Thank you once again :)

mgalloue:
You're welcome.
[...]

PR #1454, opened 2016-01-21
Clamp YScale on MultiBarHorizontalChart

jongmoon:
Clamp YScale on MultiBarHorizontalChart so that graph does not exceed the graph area when forceY option is enabled.

I raised this issue #1425 and fix this issue referencing @mgalloue 's comment and how other nv.models did.(eg. discreteBarChart, boxPlot Chart)

And I added test code on test/multiBarHorizontalChart.html. But this file did not work because wrong file reference, so I fix it together.


liquidpele:

PR #1456, opened 2016-01-24
[multiBarHorizontal] Pass event & element params on click. Add wrapLabels parameter description.

ovvn:


liquidpele:
looks good, thanks for contributing!

PR #1464, opened 2016-01-27
Fix dragging issue

mgalloue:
I need to define available width and height as global variables because it is used on the drag function and it can change during a chart update.

### DIFF
--- a/src/models/parallelCoordinates.js
+++ b/src/models/parallelCoordinates.js
@@ -10,6 +10,8 @@ nv.models.parallelCoordinates = function() {
     var margin = {top: 30, right: 0, bottom: 10, left: 0}
         , width = null
         , height = null
+        , availableWidth = null
+        , availableHeight = null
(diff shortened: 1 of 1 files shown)


src/models/parallelCoordinates.js
```diff
@@ -41,8 +43,8 @@ nv.models.parallelCoordinates = function() {
         renderWatch.reset();
         selection.each(function(data) {
             var container = d3.select(this);
-            var availableWidth = nv.utils.availableWidth(width, container, margin),
-                availableHeight = nv.utils.availableHeight(height, container, margin);
+            availableWidth = nv.utils.availableWidth(width, container, margin),
```
liquidpele: end of this line should be a semicolon then.

mgalloue:
Thanks @liquidpele , I fixed it.

liquidpele:
cool beans.

PR #1469, opened 2016-01-30
Fix for forceY on stackArea.scatter always getting set to [0]

mikeyoon:
I noticed that stackedArea.scatter.forceY can't ever be user set because it always get hard set to [0]. Since there is code to default it to [0] right above I figured this must be an error.

### DIFF
--- a/src/models/stackedArea.js
+++ b/src/models/stackedArea.js
@@ -114,7 +114,6 @@ nv.models.stackedArea = function() {
                 .y(function(d) {
                     if (d.display !== undefined) { return d.display.y + d.display.y0; }
                 })
-                .forceY([0])
                 .color(data.map(function(d,i) {
                     d.color = d.color || color(d, d.seriesIndex);
                     return d.color;


liquidpele:
No, it's not a  bug because this is stacked area, and so you always want the y axis to display zero so you can see the full stack even if all the points are above zero.

mikeyoon:
Ok, in that case, should I make a PR to remove the bit right above?

```
// If the user has not specified forceY, make sure 0 is included in the domain
// Otherwise, use user-specified values for forceY
if (scatter.forceY().length == 0) {
  scatter.forceY().push(0);
}
```

That's what threw me off. I assumed that since there's code to preserve any user-specified forceY, it had a reason. If there's no need to ever preserve their forceY, then I think this bit can be removed.

liquidpele:
oooooh, okay, so we were handling it 2 different ways.   okay, that seems valid then, accepting this PR

PR #1478, opened 2016-02-02
Do not display "NaN" in tooltip

mgalloue:
- User can customize text displayed for NaN values in a tooltip
- Undefined is displayed by default.

### DIFF
--- a/examples/documentation.html
+++ b/examples/documentation.html
@@ -968,6 +968,11 @@ as well as used to print out examples using those example inputs.
             desc: "Sets the font-size CSS style on axis labels.",
             default: "undefined",
             examples: ["12px", "1em"]
+        },
+        nanValue: {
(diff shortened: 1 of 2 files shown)


liquidpele:
Can you explain a little why this is necessary?   It seems odd to want to display undefined values in the tooltip...

mgalloue:
To be consistant with what we see on the chart we can't just hide a dimension with undefined values on the tooltip. The tooltip show all the available dimensions with the corresponding value, taking into account the sorting. Hide this dimension could be misleading the user.

mgalloue:
@liquidpele I don't understand why checks failed, do you know what's happening?

liquidpele:
Travis CI has been fairly unstable for a while now, so you have to go look at the output of the tests to see if it failed or bombed.  In this case, firefox disconnected so the tests didn't actually fail.

Issue #1475, opened 2016-02-02
If axis labels are rotated, on chart.update, the label positions are miscalculated


mikeyoon:
I noticed that if the axis labels are rotated, on the first chart render, they are placed as expected. On subsequent chart.updates, they will be misplaced due to the rotated labels affecting the textHeight calculations.

Here's a jsfiddle with the issue. You can trigger if by resizing your window after it renders. https://jsfiddle.net/L7jndb81/

The code at issue I think is the following:

```javascript
xTicks.each(function(d,i){
  var box = this.getBoundingClientRect();
  var width = box.width;
  textHeight = box.height;
  if(width > maxTextWidth) maxTextWidth = width;
});
rotateLabelsRule = 'rotate(' + rotateLabels + ' 0,' + (textHeight/2 + axis.tickPadding()) + ')';
```

On the first pass, the textHeight is set correctly, but after that since the axis persists and the height of the box is affected by the initial rotation. Not sure what the best solution is right now. An easy one might be to reset the rotation before doing the calculations, but that might be visible to the user.

liquidpele:
Updated jsfiddle to use master branch code:

https://jsfiddle.net/L7jndb81/1/

to reproduce, look at the labels under the x-axis, adjust the height, and you'll see the jump to the right just a bit as soon as you trigger the chart.update (even if you only adjust height and not width!).

PR #1479, opened 2016-02-02
Reset tick transform attribute before rotateLabel calculations

mikeyoon:
A quick fix for #1475. This is the safe and simple solution of resetting the transform attribute on xTicks before doing the textHeight calculation. I didn't notice any rendering issues because the transform reset didn't have any visual impact when I stepped through with a debugger.

I also played around with an implementation that took advantage of the persistence of the transform attribute, but it seemed like it could easily regress down the road since a lot of attributes on other elements get reset on every update.


liquidpele:
Not ideal, but I think it's better than what it is right now.

PR #1480, opened 2016-02-02
Add fontSize option for axis labels

mikeyoon:
Simple font-size override option for axis labels. I think in the future it could also allow a function override for more responsive behavior.

### DIFF
--- a/examples/documentation.html
+++ b/examples/documentation.html
@@ -963,6 +963,11 @@ as well as used to print out examples using those example inputs.
             desc: "For parallel coordinate: Label displayed under the undefined values line.",
             default: 'undefined values',
             examples: ["values not defined"]
+        },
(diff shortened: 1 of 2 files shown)


liquidpele:
This seems like something better handled with CSS.

mikeyoon:
I think CSS would solve the case where there are a limited set of font-sizes to choose from, but our situation involves the user picking from any number of font-sizes, or typing it in, kind of like formatting in excel or similar.

liquidpele:
I guess this does make sense for axis..   please document the new option in documention.html then though.

mikeyoon:
Thanks, does documentation go in the nvd3-community repo?

liquidpele:
No, I just use that for the doc site...  long story ;)

PR #1488, opened 2016-02-04
Enhancement - MultiBarWithFocusChart feature in multibarChart model

liquidpele:
recreating this to point to master... closing old PR https://github.com/novus/nvd3/pull/778

### DIFF
--- a/src/models/multiBarChart.js
+++ b/src/models/multiBarChart.js
@@ -1,4 +1,3 @@
-
 nv.models.multiBarChart = function() {
     "use strict";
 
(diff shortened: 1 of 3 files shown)


src/models/multiBarChart.js
```diff
@@ -351,15 +499,98 @@ nv.models.multiBarChart = function() {
...
+                    });
+            }
+
+
+            function onBrush() {
+                brushExtent = brush.empty() ? null : brush.extent();
+                var extent = brush.empty() ? x2.domain() : brush.extent();
+
+                //The brush extent cannot be less than one.  If it is, don't update the multibar chart.
+                if (Math.abs(extent[0] - extent[1]) <= 1) {
```
billderose: this won't paint the chart if it has a very small domain, e.g. if we're dealing with ratio and all values are between 0 and 1

src/models/multiBarChart.js
```diff
@@ -314,13 +460,15 @@ nv.models.multiBarChart = function() {
                 d.disabled = false;
 
                 switch (d.key) {
-                    case 'Grouped':
-                    case controlLabels.grouped:
+                    case 'Grouped':                    case controlLabels.grouped:
```
billderose: what's doing on with the whitespace here?

liquidpele:
This is too out of date, and doen't use the refactored focus object, closing... 

I'd like to see a focus item for multibar though so if someone can redo it then that'd be great :)

kresimirfijacko:
what is current status on this one? 
i managed to get it working, but i have problems with tooltip, it doesn't show up - i think it's because of different versions. which version of nvd did you use?
i would actually need interactive guideline if it's possible

PR #1493, opened 2016-02-05
Add support for line chart to place legend on bottom

slavede:


liquidpele:
Looks good!  Thanks for contributing!

slavede:
Glad to help, any idea when this will be tagged in some of the following versions?

liquidpele:
I built latest and pushed it up, so it's in the 1.8.2-dev version now.

slavede:
What about official version (not -dev)? Because I need that one for my build system...

liquidpele:
will probably be a while before another version comes out, we don't do it for every PR.

slavede:
Of course, no problem, it's not so urgent, it was just matter of finding out rough number, a week, a month, 2-3 months?

liquidpele:
Few months probably.

PR #1494, opened 2016-02-05
Remove unnecessary duplicate object

mgalloue:
DimensionsName is a duplicate of DimensionData so I remove it to clean the code. Also the loop for currentPosition and originalPosition is not necessary.

### DIFF
--- a/src/models/parallelCoordinatesChart.js
+++ b/src/models/parallelCoordinatesChart.js
@@ -16,7 +16,6 @@ nv.models.parallelCoordinatesChart = function () {
 		, color = nv.utils.defaultColor()
         , state = nv.utils.state()
         , dimensionData = []
-        , dimensionNames = []
         , displayBrush = true
         , defaultState = null
         , noData = null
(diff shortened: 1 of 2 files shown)


src/models/parallelCoordinatesChart.js
```diff
@@ -99,21 +98,6 @@ nv.models.parallelCoordinatesChart = function () {
...
-                            dimensionNames.splice(i, 0, k);
-                        } else {
-                            var gap = dimensionNames[idx].currentPosition - dimensionNames[idx].originalPosition;
-                            dimensionNames[idx].originalPosition = k.originalPosition;
-                            dimensionNames[idx].currentPosition = k.originalPosition + gap;
-                        }
-                });
-                //Remove old dimensions
-                dimensionNames = dimensionNames.filter(function (d) { return newDimensions.indexOf(d.key) >= 0; });
-
```
liquidpele: What were we keeping the originalPosition for here, and why is that not necessary now?

mgalloue:
@liquidpele, If the original and current positions are not set on the chart creation, default values are given according to DimensionData order. Else, if you want a specific order, you can set it outside of nvd3, by setting currentPosition and originalPosition directly. Also, the only method changing the currentPosition is the dragging function. Before, the removed code was in charge of re-ordering dimension when a new one was inserted but you have to set it outside of NVD3. NVD3 doesn't know where the dimension should be display.

mgalloue:
Hi @liquidpele, any update on this PR?
Thanks

PR #1528, opened 2016-02-22
more data for more flexible custom tooltip in multiBarChart

torstenloos:
like in stackedAreaChart and lineChart


src/models/multiBarChart.js
```diff
@@ -369,7 +369,7 @@ nv.models.multiBarChart = function() {
                                 key: series.key,
                                 value: chart.y()(point, pointIndex),
                                 color: color(series,series.seriesIndex),
-                                data: series.values[pointIndex]
+                                data: point
```
liquidpele: but point is set to the exact same thing above, so what's the point?

liquidpele:
closing as no logic changed

PR #1536, opened 2016-02-23
fix change color on scatter chart and update

UlTriX:


UlTriX:
By setting the colour on the data itself if we don't change the data but call set colour and then updateChart it will continue to use the previous set coluor palette

PR #1539, opened 2016-02-24
Format MultiChart's interactive tooltip's header with headerFormatter

joelmertanen:
Otherwise the header has the raw data, unlike with the "old" tooltip

Steps to reproduce:
- Add `chart.useInteractiveGuideline(true);` to `examples/multiChart.html`
- Add formatting to x-axis, for example `chart.xAxis.tickFormat(d3.format(',.1f'));`

In the current `master` branch you can see different formatting on tooltip's header and x-axis. With this fix, the formatting is the same in the interactive tooltip and the x-axis.

### DIFF
--- a/src/models/multiChart.js
+++ b/src/models/multiChart.js
@@ -419,6 +419,9 @@ nv.models.multiChart = function() {
 
                     interactiveLayer.tooltip
                     .chartContainer(chart.container.parentNode)
+                    .headerFormatter(function(d, i) {
+                        return xAxis.tickFormat()(d, i);
+                    })
                     .valueFormatter(function(d,i) {
                         var yAxis = allData[i].yAxis;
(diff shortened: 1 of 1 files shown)


src/models/multiChart.js
```diff
@@ -419,6 +419,9 @@ nv.models.multiChart = function() {
 
                     interactiveLayer.tooltip
                     .chartContainer(chart.container.parentNode)
+                    .headerFormatter(function(d, i) {
+                        return xAxis.tickFormat()(d, i);
+                    })
```
liquidpele: The problem with this (and the valueFormatter below) is that if someone does:

    chart.interactiveGuideline.tooltip.headerFormatter(function(){ return 'a'; });    

I think you should be able to do ._headerFormatter(...) and ._valueFormatter(...), if you start an option with an underscore then it tries to only use the passed value if it wasn't overwritten already

PR #1529, opened 2016-02-22
fixed tooltip bug for stackedAreaChart

Snuggert:
Fixed a bug that was troubling me,
stackedValue is defined nowhere, this seemed the correct way of fixing it.

PR #1540, opened 2016-02-24
interactive legend fix on stackedAreaChart

UlTriX:


Snuggert:
There already are several pull requests for this bug: #1508, #1529, ...
Yet the team seems reluctant merging them.

So this is the perfect opportunity to show users suffer this bug.! :+1:

liquidpele:
merged a diff one

PR #1543, opened 2016-02-25
Sparkline additional spots

asakalou:
Allow users to pass additional array of points 'spots' to sparkline.

Spots are rendered on the sparkline area. You can pass cssClass attr within a spot object to apply some style.

## Preview
screen shot 2016-02-25 at 19 03 09

## Example configuration

Styles:
`
.nvd3.nv-sparkline .nv-spot.odd {
   stroke: green;
   fill: #green;
}
`
`
.nvd3.nv-sparkline .nv-spot.even {
   stroke: blue;
   fill: blue;
}
`

```javascript
nv.addGraph({
        generate: function () {

            var spots = [];
            for (var i = 0; i < 20; ++i) {
                spots.push({
                    x: i * 5,
                    y: Math.cos(i * 2),
                    cssClass: i % 2 ? 'odd' : 'even'
                });
            }

            var chart = nv.models.sparkline()
                    .width(400)
                    .height(30)
                    .spots(spots);

            d3.select(chartId)
                    .datum(sine())
                    .call(chart);

            return chart;
        }
    });
```

### DIFF
--- a/examples/sparkline.html
+++ b/examples/sparkline.html
@@ -20,6 +20,15 @@
         #chart1 {
             height: 30px;
         }
+        .odd {
+            stroke: #801010;
+            fill: #802638;
+        }
+
+        .even {
+            stroke: #555880;
+            fill: white;
+        }
     </style>
 </head>
 <body>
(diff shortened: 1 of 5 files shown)


liquidpele:
I'm not sure I like this... sparkline is supposed to be super-minimal. what you're doing looks more like just a line and scatter chart together without the legend or the axis, which you can do with multichart...

liquidpele:
That is to say, I don't think this belongs in main nvd3, but that's why it's open source, so people can adjust it to their own needs as necessary!

PR #1544, opened 2016-02-26
Add reduceXTicks option to discrete bar chart

mikeyoon:
Just copying the functionality from multibar chart, but setting as false by default. One minor optimization is the selector on xTicks to match the xTicks selector on multibar. Seemed like the original one was not at the right level and the filter didn't work. Not sure if there's a reason to make xTicks one level higher on discrete bar.


sezgi:
Fixed bug where bar charts show bars for 0 and null values. Here's a jsfiddle showing the bug:

https://jsfiddle.net/sezgi/9kLzgxbw/

sezgi:
Fixed bug where bar charts show bars for 0 and null values. Here's a jsfiddle showing the bug:

https://jsfiddle.net/sezgi/xg1b2zz2/

arturochian:
Please add this pull request :octocat:

PR #1549, opened 2016-03-04
Added support for axisLabelDistance to right aligned axis

robinv:
Added the same logic as for left aligned axes.


liquidpele:
cool beans

PR #1550, opened 2016-03-05
Fixed bug where bar charts show bars for 0 and null values

sezgi:
Here's a jsfiddle showing the bug:
https://jsfiddle.net/sezgi/xg1b2zz2/


smills2929:
It should show bars for 0 values, however it should not for null values.

liquidpele:
IIRC the novus guys did that on purpose so that you could see a tiny bit of the chart, otherwise it looked like it was just missing.

smills2929:
I think null values should not show, but it makes sense to show 0 values.
To me null should be interpreted as missing.
On 27 Apr 2016 06:36, "liquidpele" <notifications@github.com> wrote:

IIRC the novus guys did that on purpose so that you could see a tiny bit of
the chart, otherwise it looked like it was just missing.

—
You are receiving this because you commented.
Reply to this email directly or view it on GitHub
<https://github.com/novus/nvd3/pull/1550#issuecomment-214908325>

liquidpele:
You say it makes sense to show 0 values... but if you have a height of zero then it's effectively not shown...

PR #1556, opened 2016-03-08
allow boolean options for showing min/max points and current point

brchristian:
I noticed that the `sparkline` model helpfully shows the minimum and maximum points, as well as the current point. For users wanting the ability to toggle the display of these points, I've added simple boolean setter/getter methods that allow that option.

So for instance, someone wanting to show the current point (which defaults to `true`), but not the min/max points, could simply call:

````
 var chart = nv.models.sparkline()
                    .width(400)
                    .height(30)
                    .showMinMaxPoints(false)
````

It seems helpful to allow users this option if they choose.

### DIFF
--- a/build/nv.d3.js
+++ b/build/nv.d3.js
@@ -12450,8 +12450,6 @@ nv.models.sparkline = function() {
         , yDomain
         , xRange
         , yRange
-        , showMinMaxPoints = true
-        , showCurrentPoint = true
         , dispatch = d3.dispatch('renderEnd')
         ;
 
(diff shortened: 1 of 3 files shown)


liquidpele:
This includes build files in the PR, please only include the src files

brchristian:
@liquidpele :+1:

liquidpele:
Looks okay to me... since you're adding new options though, please document those in the documentation.html file in the big json object at the bottom that's used to build the doc page.

brchristian:
@liquidpele I wasn’t totally sure about the formatting so simply added them to the bottom. Let me know if that looks okay to you.

liquidpele:

Issue #1558, opened 2016-03-09
SparklinePlus hover value not updating with new data


jeremywestwood:
I have an NVD3 SparklinePlus that I am updating with new data when it comes through. The line itself updates as expected but the hover values are as if the original data is still there. Example code is below and here is a link to a [jsfiddle](http://jsfiddle.net/jpwestwood/3bq65z4t/) with a working example of the problem:

```
 var chart
  nv.addGraph(function() {
    chart = nv.models.sparklinePlus()
    chart.showLastValue(true)
    d3.select("#spark svg").datum(sine(10, 20)).call(chart)
    return chart
  })

  setTimeout(function() {
    d3.select("#spark svg").datum(sine(5, 30)).call(chart)
  }, 1000)
```

Am I doing something incorrect in the update function, in terms of how I replace the data, or is this a bug in the SparklinePlus?

jeremywestwood:
Have created a potential fix for the issue in this pull request: https://github.com/novus/nvd3/pull/1561

PR #1561, opened 2016-03-10
Pass the data through to sparklineHover and updateValueLine

jeremywestwood:
A sparklinePlus would keep using the original data for the hover values once the data was updated. Due to scoping. This fixes this by passing through the latest data to the sparklineHover and updateValueLine functions.
This is a fix for  [issue 1558](https://github.com/novus/nvd3/issues/1558)

### DIFF
--- a/src/models/sparklinePlus.js
+++ b/src/models/sparklinePlus.js
@@ -95,7 +95,7 @@ nv.models.sparklinePlus = function() {
             }
 
             gEnter.select('.nv-hoverArea').append('rect')
-                .on('mousemove', sparklineHover)
+                .on('mousemove', function() {return sparklineHover} (data))
                 .on('click', function() { paused = !paused })
                 .on('mouseout', function() { index = []; updateValueLine(); });
 
@@ -105,7 +105,7 @@ nv.models.sparklinePlus = function() {
                 .attr('height', availableHeight + margin.top);
 
             //index is currently global (within the chart), may or may not keep it that way
-            function updateValueLine() {
+            function updateValueLine(data) {
                 if (paused) return;
 
                 var hoverValue = g.selectAll('.nv-hoverValue').data(index);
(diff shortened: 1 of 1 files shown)


liquidpele:
Why are you needing this?   Those functions are using the main "data" variable that gets passed into the chart object, so I'm not seeing what this is supposed to accomplish.

jeremywestwood:
This was a fix for the issue: https://github.com/novus/nvd3/issues/1558

There is a [jsfiddle](http://jsfiddle.net/jpwestwood/3bq65z4t/) that demonstrates the problem. It could very well be that I am not updating the graph correctly in which case it would be helpful to know what I'm doing wrong.

liquidpele:
Ok, I'll look at those

liquidpele:
I still don't think this is the proper fix for the issue..  updating the data should change the main data variable, passing it around in methods shouldn't be required...

jeremywestwood:
Yes, I wasn't sure if this was the correct fix either but couldn't figure out another way to fix it. What is strange is that the main data variable is changed, it's just the reference to it in the updateValueLine and sparklineHover functions seem to still link to the original data.

liquidpele:
That sounds like a problem with javascript scoping.  Are you using just plain nvd3, or something like angular-nvd3?   If just regular, if you can create a jsfiddle showing the problem (linking to the master branch's code via rawgit.com) then I can take a look...

jeremywestwood:
Thanks!
Here is a jsfiddle that demonstrates the problem. I am using plain nvd3.

http://jsfiddle.net/jpwestwood/3bq65z4t/1/

It creates a sin graph and then after 2 seconds changes to a dfferent sin graph but the hover still shows values from the original.

liquidpele:
Well, the real issue is you're passing in a whole other data object instead of updating the original object.  I modified your jsfiddle to make it work with current master. 

https://jsfiddle.net/3bq65z4t/2/

jeremywestwood:
Ah, thanks! I thought there was probably something I was doing wrong.

PR #1562, opened 2016-03-10
fixed xDomain calculating in multiChart

gkucmierz:


src/models/multiChart.js
```diff
@@ -95,7 +95,7 @@ nv.models.multiChart = function() {
                     })
                 });
 
-            x   .domain(d3.extent(d3.merge(series1.concat(series2)), function(d) { return getX(d) }))
+            x   .domain(d3.extent(d3.merge(series1.concat(series2)), function(d) { return d.x }))
```
liquidpele: I don't think this is right, the x value should always bet retrieved via getX because people can override that to get the x value from a differently named attribute.

gkucmierz: Becaouse this code is so much messed up. Look at this this value is taked before using getX and it is stored in x property.
If you will store x value in other than x property, actual code will never work correctly.

I am actual fixing this nvd3 in our internal project in corporation where I work.
I can easily point much, much more errors in nvd3

gkucmierz:
Nice, you just closed my pull request, and you are still not correct.

And I need to fixing this shitty library in my corporation, sounds great…

src/models/multiChart.js
liquidpele: Oh, I see, it rebuilds the two series with x and y and then uses those.

liquidpele:
You sound like you need some coffee bro ;)

gkucmierz:
I am sorry for my comment, this is one of the best libraries for charts.
But there are so much errors I need to fix every day, and company policy of my firm dont allow me to develop this code here.

PR #1565, opened 2016-03-11
Add duration option for scatterChart model

the-fool:
The scatterChart model's duration setter was not causing any effect in the chart.
Now the setter updates the duration value for the scatter model, which conditions the duration for the scatterChart.

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -41,7 +41,8 @@ nv.models.scatter = function() {
         , useVoronoi   = true
         , duration     = 250
         , interactiveUpdateDelay = 300
-        , showLabels    = false 
+        , showLabels    = false
+        , fillOpacity   = 0.5
         ;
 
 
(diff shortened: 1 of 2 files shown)


src/models/scatter.js
```diff
@@ -410,7 +411,7 @@ nv.models.scatter = function() {
                 .style('fill', function(d,i) { return color(d, i) })
                 .style('stroke', function(d,i) { return color(d, i) })
                 .style('stroke-opacity', 1)
-                .style('fill-opacity', .5);
+                .style('fill-opacity', fillOpacity);
```
liquidpele: you can already override this with CSS.

src/models/scatterChart.js
```diff
@@ -375,6 +374,11 @@ nv.models.scatterChart = function() {
             rightAlignYAxis = _;
             yAxis.orient( (_) ? 'right' : 'left');
         }},
+        duration: {get: function(){return duration;}, set: function(_){
+            duration=_;
+            renderWatch.reset(duration);
+            scatter.duration(duration);
+        }},
```
liquidpele: It was already inheriting duration from scatter

PR #1567, opened 2016-03-15
Allows legend to update properly even with different symbol shapes.

galtieri:
Allows legend to update properly even with different symbol shapes.

### DIFF
--- a/src/models/legend.js
+++ b/src/models/legend.js
@@ -61,7 +61,7 @@ nv.models.legend = function() {
                     .attr('class','nv-legend-symbol')
                     .attr('r', 5);
 
-                seriesShape = series.select('circle');
+                seriesShape = series.select('.nv-legend-symbol');
             } else if (vers == 'furious') {
                 seriesEnter.append('rect')
                     .style('stroke-width', 2)


liquidpele:
okay... but it still only ever adds circles, am I missing something?  How are you using a different shape?

galtieri:
I changed the simbols using this kind of code ...

"shapesmod" tell me if the series has to be its simbol changed
"shapes" has the required symbols: "circle", "triangle", etc...

Without modification of legend's code the symbol does'nt update properly.

// ********************************************
d3.select('#chart0 svg').selectAll('.nv-series')[0].forEach(function(d,i) {
  if(shapesmod[i])
  {
      var group = d3.select(d),
          circle = group.select('circle');
      var ufill = circle.style('fill');
      var ufillop = circle.style('fill-opacity');
      var ufillstr = circle.style('stroke');
      var ufillstrw = circle.style('stroke-width');
      circle.remove();
      var usym = group.append('path');
        usym.attr('class','nv-legend-symbol')
        .attr('d', d3.svg.symbol().type(shapes[i]))
        .style('fill-opacity', ufillop)
        .style('stroke-width', ufillstrw)
        .style('stroke', ufillstr)
        .style('fill', ufill);
    }
});
// ********************************************

Issue #1406, opened 2015-12-25
multichart tooltip bug


hendyadys:
multichart is supposed to look at the tickFormat for the appropriate yAxis when rendering tooltip. However, only yAxis1 is ever triggered. This can be reproduced by playing with the nvd3 examples 
https://github.com/novus/nvd3/tree/master/examples

also there is not headerFormatter for the xAxis, which has been mentioned by another user

cesare-montresor:
experiencing the same issue here a screenshot:

![screen shot 2016-03-31 at 11 37 42](https://cloud.githubusercontent.com/assets/447853/14172112/ba2b2ac6-f736-11e5-9ab3-53b894baec5d.jpg)

PR #1583, opened 2016-03-31
bug fix title tooltip multiChart model

cesare-montresor:
resolves bug: https://github.com/novus/nvd3/issues/1406


liquidpele:
fixed by https://github.com/novus/nvd3/pull/1539

PR #1589, opened 2016-04-06
Enhancement - More modular implementation of focus charts.

Snuggert:
Used the focus chart base from the lineChart model and adapted the code so the StackedAreaChart model could focus as well.

Should probably include some unit-testing in the future.


liquidpele:
Oh wow, this is cool.... looks like you have a conflict with master though, you need to merge in upstream master and fix the conflict in your branch.

Snuggert:
I merged my branch with the current master.

liquidpele:
wtf... github is showing stuff in your PR that is already merged to master now.   

Guess I'll wait and see if it fixes itself...

Snuggert:
Could it be that I broke everything in my attempt to merge with upstream?

liquidpele:
Not sure... I mean, I merged in other PRs yesterday, and I see those in your changeset, so I'm confused why github would show them here since the diff with master shouldn't show them.  Something weird like this happened a while back and I contacted github support and it fixed itself magically before they responded (of course) so I dunno. 

Is this the process you used?  If not, this is the proper way:

1. First create the upstream:     https://help.github.com/articles/configuring-a-remote-for-a-fork/ 
2. Then merge latest upstream into your branch:    https://help.github.com/articles/syncing-a-fork/

Snuggert:
Yup. That's the guide I used.

Snuggert:
I think I unfucked everything. :)

liquidpele:
Ah, that looks better!   

I'll grab this and test it a bit first when I get a chance since it's kind of core functionality, but I'd love to break out focus to be reusable like this!

Issue #474, opened 2014-03-19
Ability to zoom chart


dev-tim:
Are you going to add some zoom/pan features to nvd3, or there is a way how I can  add this features to existing models?

Vineeth-Mohan:
I am also looking for the same.
This is exactly what i am looking for - http://www.highcharts.com/demo/line-time-series

liquidpele:
The blocks example there is pretty cool... something to look into.   Keeping marked as Feature.

Untit1ed:
+1. Really need this feature for every graph.

Vineeth-Mohan:
@DavidSouther  -Kindly let us know the plan to go ahead for this plan. I badly need this one and moving to some other charting library just for this would be hard to implement.

DavidSouther:
@liquidpele drawing your attention to this.

liquidpele:
@Vineeth-Mohan David isn't working for novus anymore, so I'm doubtful he'll want to invest the time ;)

Regarding this feature (http://bl.ocks.org/mbostock/3892919) it's a cool feature, something I could look into at some point for the scatter chart at least... however it doesn't seem like it makes as much sense for other charts.  Regardless, I probably wouldn't work on it for a while, so if you need it soon I suggest you fork the project and have a go at!  Be sure to send us a pull request if you get it working so we can merge the new feature back in :)

Vineeth-Mohan:
@liquidpele  - Sure , i will give a shot. I am not very familiar to D3 , if i get stuck anywhere i will get back to you. If there is any pointers from NvD3 , on where to star , that would be great

PBrockmann:
nvd3 would gain a lot to get zoom feature on axis, single or both.
Look at: http://jsfiddle.net/PBrockmann/oh7wjs45/
same as http://bl.ocks.org/jgbos/9752277, just added changes of cursor to col-resize and row-resize when mouse is over an axis.

liquidpele:
Oh cool, thanks for the extra examples :)

sudotliu:
Any guess on when this might be implemented?

liquidpele:
Well, I'm not working on it so unless someone else is then it'll be a while ;).  

If anyone does it be sure to send us a pull request!

jaypinho:
[...]

PR #1591, opened 2016-04-08
Add drag-to-zoom feature to lineChart

imbolo:
#474 
support drag-to-zoom operation,  just like highcharts

zoom


DmitryEfimenko:
awesome work!

PR #1601, opened 2016-04-14
Implement "reversed" property for legend

falbrech-hsdg:
For some chart types, especially stackedAreaChart, it can be useful to "reverse" the legend's items to match the more natural way of looking at the chart (from top to bottom, while series are stacked from bottom to the top).

This commit implements this feature in a very straightforward way. It currently only works when "adjust" is set to true.

### DIFF
--- a/src/models/legend.js
+++ b/src/models/legend.js
@@ -17,6 +17,7 @@ nv.models.legend = function() {
         , updateState = true   //If true, legend will update data.disabled and trigger a 'stateChange' dispatch.
         , radioButtonMode = false   //If true, clicking legend items will cause it to behave like a radio button. (only one can be selected at a time)
         , expanded = false
+        , reversed = false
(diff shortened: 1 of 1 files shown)


liquidpele:
Not sure about this... it seems like an odd/specific way to control the legend...

falbrech-hsdg:
Do you mean the implementation or the feature itself which is odd? For the implementation, I cannot judge (but implementation is more consistent in #1612 anyway).

For the feature, it is common in other charting libraries, e.g. Highcharts:
http://jsfiddle.net/gh/get/jquery/1.7.2/highslide-software/highcharts.com/tree/master/samples/highcharts/legend/reversed/

And as pointed out, it looks (for our charts) much more natural... see referenced #1612 for a screenshot how it looks with "reversed" turned on.

PR #1605, opened 2016-04-15
Implement alignAnchor feature and clean layout strategy for legends

falbrech-hsdg:
Renames "align" property to "columnize", replaces "rightAlign" property
with "alignAnchor" property (valid values: left, right, center).

All combinations of columnize and alignAnchor are supported.

This Pull Request includes the changes of Pull Request #1601, thus can replace that Pull Request.

This is how legend would look like with legend.alignAnchor('center') and legend.columnize(false):

image


falbrech-hsdg:
Will add some more tests for legend with a new PR.

Issue #1606, opened 2016-04-15
Line Chart interactive guideline does not highlight hover point when some series are disabled


dinkydani:
Disabling a series breaks some of the lines highlight hover points.

[Plunkr] (http://plnkr.co/edit/YnpLhGoXqCNvg2MCGJdU?p=preview) demonstrating the issue with the latest code from the master branch.
Disable series "A" and observe how series "B" no longer has the highlight point on hover. This continues to occur as other series are enabled/disabled.

I identified the issue was caused by filtering out disabled series before highlighting the point using the series index, resulting in an 'off by 1' error.

PR #1610, opened 2016-04-19
 Fixed issue #1606: Line Chart interactive guideline does not highlight hover point when some series are disabled

dinkydani:
Fixed the issue by assigning the series index inside the `forEach` rather than filter and using the data index rather than series index for the tooltip color.

I have checked in the build files so I can use this commit in my project today. Advise if you'd like me to remove them before you merge.

Cheers, Danielle

### DIFF
--- a/src/models/lineChart.js
+++ b/src/models/lineChart.js
@@ -383,11 +383,17 @@ nv.models.lineChart = function() {
                 var singlePoint, pointIndex, pointXLocation, allData = [];
                 data
                     .filter(function(series, i) {
-                        // Assign the index here instead of in the filter
-                        // as the DOM removes disabled series
-                        return !series.disabled && !series.disableTooltip;
+                        // Assign a color index to the series if it's not disabled
+                        // to use for the tooltip colors rather than the series index
+                        if(!series.disabled && !series.disableTooltip) {
+                            series.colorIndex = i;
+                            return series
+                        }
                     })
                     .forEach(function(series,i) {
+                        // Assign the index here instead of in the filter
+                        // as the DOM removes disabled series
+                        series.seriesIndex = i;
                         var extent = focusEnable ? (brush.empty() ? x2.domain() : brush.extent()) : x.domain();
                         var currentValues = series.values.filter(function(d,i) {
                             return lines.x()(d,i) >= extent[0] && lines.x()(d,i) <= extent[1];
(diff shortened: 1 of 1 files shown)


liquidpele:
Please remove build files from the PR

dinkydani:
@liquidpele done

src/models/lineChart.js
```diff
@@ -383,7 +383,8 @@ nv.models.lineChart = function() {
                 var singlePoint, pointIndex, pointXLocation, allData = [];
                 data
                     .filter(function(series, i) {
-                        series.seriesIndex = i;
+                        // Assign the index here instead of in the filter
+                        // as the DOM removes disabled series
```
liquidpele: this comment says your'e assigning the index here, but you moved it away from here...  ?

src/models/lineChart.js
```diff
@@ -401,10 +402,12 @@ nv.models.lineChart = function() {
                         if (point === undefined) return;
                         if (singlePoint === undefined) singlePoint = point;
                         if (pointXLocation === undefined) pointXLocation = chart.xScale()(chart.x()(point,pointIndex));
+                        // Use the original data index rather than filtered series index for the color
+                        var colorIndex = data.findIndex(function(d) { return d.key === series.key });
```
liquidpele: This is basically looping over the whole series list for every series.   It would be better to do this above in the filter function, and instead of using i, just use a counter that only increments when it's not disabled.  That way you're not having to loop over the series list again.

src/models/lineChart.js
dinkydani: Whoops, sorry I messed up here when removing the build files it should be doing `series.seriesIndex = i;` inside the forEach. Will fix.

src/models/lineChart.js
dinkydani: That makes sense, will do

dinkydani:
@liquidpele Updated, let me know if you have any more change suggestions :)

AlexandreDenis:
Thanks @dinkydani for your fix. I added it in my project and it seems to work.
However, there is another bug about the hover points. Indeed they aren't "completes" when some series are disabled.

To show you an example, below is a screenshot that illustrates the issue :
Exemple of incomplete hover points

So as you can see above, hover points on the plots are incomplete. Have you any idea ?

src/models/lineChart.js
```diff
@@ -383,10 +383,17 @@ nv.models.lineChart = function() {
...
+                        // to use for the tooltip colors rather than the series index
+                        if(!series.disabled && !series.disableTooltip) {
+                            series.colorIndex = i;
+                            return series
+                        }
                     })
                     .forEach(function(series,i) {
+                        // Assign the index here instead of in the filter
+                        // as the DOM removes disabled series
+                        series.seriesIndex = i;
```
liquidpele: Isn't this going to make two different series have the same seriesIndex if you start off with some disabled and enable things since this is after it's been filtered?    I think this should still be in the .filter(...) but before the if block right?

liquidpele:
This branch appears to have conflicts now too...

liquidpele:
ping... I can't merge this unless you merge in latest master and resolve any conflicts.

src/models/lineChart.js
liquidpele: ping.

liquidpele:
Closing this PR as abandoned.

PR #1613, opened 2016-04-19
Add option to show percentage in pieChart tooltip

mrijke:
This commit adds an option to the pieChart model which allows the user
to choose whether to show the percentage taken by the hovered slice
in the tooltip.


liquidpele:
neat

Issue #707, opened 2014-10-24
bar chart fill-opacity defined in CSS is overwritten when data is reloaded


micheldanjou:
Hi,

I have noticed that the fill-opacity CSS attribute used to represent bars in bar charts is over-written to 0.75 each time new data are loaded.

- I would like the fill-opacity to be 1 by default as the colors look a bit washed out with .75, especially on dark backgrounds.
- I have changed every single reference of "fill-opacity" in the css to 1 and the issue still happens.
- I can change the "fill-opacity" when live-editing the CSS in chrome but it gets overwritten as soon as new data is loaded.
- I can successfully change other opacity properties such as the stroke-opacity using the CSS file. This proves that the CSS file I am editing is the one being used.

I believe this is a bug but if anybody as any suggestions about it I would like to hear from you.
Regards
Michel

liquidpele:
Feel free to send us a pull request with a fix... to get you started, someone did similar with lines in this commit:

https://github.com/nvd3-community/nvd3/commit/14b686ece509154ef5e8573e0aca8570936387ef

liquidpele:
Closing this since it's not really a bug or anything, you can override inline CSS by using the !important css flag, or you can create a pull request that improves things.  Cheers!

PR #1614, opened 2016-04-19
Hand over fillOpacity for bars in multiBar and multiBarHorizontal charts

FatsoTheFatCat:
As #707 discusses, it'd be nice to have control over the fill opacity for the bar charts


liquidpele:
Looks fine, thanks!

PR #1615, opened 2016-04-20
fixing tooltip floating-point number precision, fixes #1385

### LINKED ISSUE #1385: Blurred tooltip
When tooltip translate css property has floating-point number it's borders are blurred on Ubuntu Chrome.

ywplee:
crediting @AdamWaz, fixing #1385


liquidpele:
Thanks!

PR #1623, opened 2016-04-26
StackedAreaChart tooltip improvements

mrijke:
This commit improves the behavior of the StackedAreaChart tooltip by handling
null values consistently with the other charts (i.e. display "N/A" when the value
is null).

Additionally, the "TOTAL" in the tooltip is only shown if there are non-null values.


liquidpele:
sweet, thanks for contributing!

PR #1629, opened 2016-04-28
Bugfix in initTooltip() function

mehranshakeri:
consol
js

When there are several charts, they share the same tooltip! In some exceptional cases, DOM element is removed but NDV3 tooltip object still exist! This causes initTooltip function to not create new tooltip. As the result, NVD3 tooltip object will be updated but since it doesn't exist in DOM elements it won't be shown anywehre.

As you can see in attached files, when it happens that the tooltip DOM elements is removed (I don't know who!), NVD3 tooltip object still contains it's value but if you try to find it via JQuery or d3.select, there is no such element in DOM.

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -247,7 +247,7 @@ nv.models.tooltip = function() {
 
     // Creates new tooltip container, or uses existing one on DOM.
     function initTooltip() {
-        if (!tooltip || !tooltip.node()) {
+        if (!tooltip || !tooltip.node() || d3.select('.nvtooltip#'+id)[0][0] === null) {
             var container = chartContainer ? chartContainer : document.body;
             // Create new tooltip div if it doesn't exist on DOM.
 
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -247,7 +247,7 @@ nv.models.tooltip = function() {
 
     // Creates new tooltip container, or uses existing one on DOM.
     function initTooltip() {
-        if (!tooltip || !tooltip.node() || d3.select('.nvtooltip#'+id)[0][0] === null) {
+        if (!tooltip || !tooltip.node() || d3.select('.nvtooltip#'+id).empty()) {
             var container = chartContainer ? chartContainer : document.body;
             // Create new tooltip div if it doesn't exist on DOM.
 


EvaSDK:
Would this be related to issue #1680 ?

mehranshakeri:
It could be the case. I just checked our charts and I can confirm that we have stacked area chart with "useInteractiveGuideline: true" and bar chart without "useInteractiveGuideline: false". And this bug happens only when we have these charts together. So maybe one is removing the others tooltip and other one doesn't know it's tooltip is removed from DOM.

liquidpele:
This can't be merged due to conflicts so that'll have to be fixed... 

but regardless, I'm confused how they could share a tooltip since the tooltip is created with a randomly generated id (unless you used the id() option to set all the charts to use the same id??)

src/tooltip.js
```diff
@@ -247,7 +247,7 @@ nv.models.tooltip = function() {
 
     // Creates new tooltip container, or uses existing one on DOM.
     function initTooltip() {
-        if (!tooltip || !tooltip.node()) {
+        if (!tooltip || !tooltip.node() || d3.select('.nvtooltip#'+id)[0][0] === null) {
```
liquidpele: should do:    d3.select(....).empty() to test if the selection returns nothing.

mehranshakeri: You are right.

mehranshakeri:
I checked our code. We don't set "id" anywhere. I debugged again and if you don't check id (what is happening in current release) it uses already existing instance and just changes its "id".

liquidpele:
Ohhhhh, it's setting a random ID, but then it's doing a select on just .nvtooltip so it finds the same one every time and changes the id...  that's the bug I think.

mehranshakeri:
I was trying to say the same thing :)))) It seems I was lost in too much details

mehranshakeri:
@liquidpele 
It seems proper solution would be:

`tooltip = d3.select(container).select('#'+id).data(data);`

Shall I create another pull request since this branch is old and merging has conflict?

liquidpele:
sure

mehranshakeri:
New pull request is created #1762

hajdbo:
This sometimes creates multiple tooltips, even for a single chart. In the following example, a single chart leads to 4 tooltips.
image

PR #1635, opened 2016-05-04
Add transitionDuration option to multiChart

strax:
Adds a `transitionDuration` option to multiChart model that affects the transition duration of the axes and lines of the chart.

### DIFF
--- a/src/models/multiChart.js
+++ b/src/models/multiChart.js
@@ -574,7 +574,7 @@ nv.models.multiChart = function() {
             }
         }},
 
-        transitionDuration: {get: function(){return duration;}, set: function(_) {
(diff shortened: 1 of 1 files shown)


src/models/multiChart.js
```diff
@@ -571,6 +572,13 @@ nv.models.multiChart = function() {
                 scatters1.interactive(false);
                 scatters2.interactive(false);
             }
+        }},
+
+        transitionDuration: {get: function(){return duration;}, set: function(_) {
```
liquidpele: The option should be named the same... just "duration", so that it matches other charts.

strax: Changed

liquidpele:
looks good, thanks for contributing!

PR #1636, opened 2016-05-04
Improve Sunburst margin & show percentage on tooltip

mrijke:
This commit fixes the margin on the Sunburst chart.
Since the sunburstChart.js has no longer a wrapper container,
the margin options are useless. Instead, the margin is now
shared between the sunburstChart and the sunburst objects.
Furthermore, the sunburst.js did not take into account the
margin when calculating the middle of the container.
Also, when the update() function was called the container
did not apply the newly calculated translate() arguments.

Finally, the sunburstChart.js now has the option to show
the percentage of a node in the tooltip, similar to the
pieChart.

### DIFF
--- a/src/models/sunburst.js
+++ b/src/models/sunburst.js
@@ -58,6 +58,12 @@ nv.models.sunburst = function() {
         return centerAngle;
     }
 
+    function computeNodePercentage(d) {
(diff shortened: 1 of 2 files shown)


liquidpele:
Seems okay.

PR #1444, opened 2016-01-14
Enhancements to Multichart

matthewpbyrne:
I required stackable bar charts in a multichart in one of my projects, and I have added an implementation, along with tests here.

PR #1637, opened 2016-05-04
Bar stacking in multichart.

matthewpbyrne:
An updated version of the earlier abandoned PR at  https://github.com/novus/nvd3/pull/1444

Bar-stacking capability was required in one of my multicharts, hence I've added this feature here.

In response to earlier comments: 
- The use case for showY1AxisZero & showY2AxisZero is where the user wishes to explicitly show a value of 0 one of their chart's Y-axes.
- "stackBars" was changed to "chart.bars.stacked(true)" as suggested.

### DIFF
--- a/src/models/multiChart.js
+++ b/src/models/multiChart.js
@@ -19,7 +19,12 @@ nv.models.multiChart = function() {
         useVoronoi = true,
         interactiveLayer = nv.interactiveGuideline(),
         useInteractiveGuideline = false,
-        legendRightAxisHint = ' (right axis)'
+        legendRightAxisHint = ' (right axis)',
+        showY1Axis = true,
+        showY1AxisZero = false,
+        showY2Axis = true,
+        showY2AxisZero = false,
+        bars = {stacked: false}
         ;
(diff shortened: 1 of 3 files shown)


src/models/multiChart.js
```diff
@@ -19,7 +19,12 @@ nv.models.multiChart = function() {
         useVoronoi = true,
         interactiveLayer = nv.interactiveGuideline(),
         useInteractiveGuideline = false,
-        legendRightAxisHint = ' (right axis)'
+        legendRightAxisHint = ' (right axis)',
+        showY1Axis = true,
+        showY1AxisZero = false,
+        showY2Axis = true,
+        showY2AxisZero = false,
```
liquidpele: It looks like multiChart is missing the forceX and forceY options that exist on most of the other charts, it should match this same style instead of inventing this separate (and more limited) option I think.

liquidpele:
Ping... please update this PR or we can just close it.

liquidpele:
closing due to no reply.

Issue #853, opened 2015-03-06
arcTween for donuts not working


nomego:
When updating a piechart "donut" slice's datum with
```javascript
d3.select(svgElement).datum(newData);
```
and then calling
```javascript
chart.update();
```
I sometimes get
```
Uncaught TypeError: Cannot use 'in' operator to search for 'amount' in -3860.92
d3.js:5139 d3_interpolateObject
d3.js:5230 d3.interpolators
d3.js:5225 d3_interpolate
d3.js:5133 d3_interpolateObject
d3.js:5230 d3.interpolators
d3.js:5225 d3_interpolate
d3.js:5235 d3_interpolateArray
d3.js:5230 d3.interpolators
d3.js:5225 d3_interpolate
d3.js:5133 d3_interpolateObject
d3.js:5230 d3.interpolators
d3.js:5225 d3_interpolate
d3.js:5133 d3_interpolateObject
d3.js:5230 d3.interpolators
d3.js:5225 d3_interpolate
nv.d3.js:8943 arcTween
d3.js:8022 attrTween
d3.js:8166 (anonymous function)
d3.js:282 d3_class.forEach
d3.js:8165 start
d3.js:8159 (anonymous function)
d3.js:2085 d3_timer_mark
d3.js:2065 d3_timer_step
```

with the current arcTween() function:

```javascript
            function arcTween(a) {
                a.endAngle = isNaN(a.endAngle) ? 0 : a.endAngle;
                a.startAngle = isNaN(a.startAngle) ? 0 : a.startAngle;
                if (!donut) a.innerRadius = 0;
                var i = d3.interpolate(this._current, a);
                this._current = i(0);
                return function(t) {
                    return arc(i(t));
                };
            }
```

Replacing it with an approach like https://gist.github.com/mbostock/5100636 however, rewritten as:

```javascript
            function arcTween(a) {
                a.endAngle = isNaN(a.endAngle) ? 0 : a.endAngle;
                a.startAngle = isNaN(a.startAngle) ? 0 : a.startAngle;
                if (!donut) a.innerRadius = 0;
                var i = d3.interpolate(a.startAngle, a.endAngle);
                this._current = i(0);
                return function(t) {
                    a.endAngle = i(t);
                    return arc(a);
                };
            }
```

works.

Can you see anything that might cause this based on my usage or does the arcTween() function need improvement?

liquidpele:
What version of d3 are you on?  Can you provide a jsfiddle that demonstrates the error?

[...]

PR #1638, opened 2016-05-05
arcTween fix

nomego:
Fixes tween (#853), support for arbitrary properties

### DIFF
--- a/src/models/pie.js
+++ b/src/models/pie.js
@@ -339,14 +339,15 @@ nv.models.pie = function() {
                 return a > 90 ? a - 180 : a;
             }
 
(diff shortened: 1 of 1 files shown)


jessenieminen:
I tested this.

I liked the fact that with this approach the pieChart would have an initial load animation but I think that when implemented this way, the tween doesn't then look good for the transitions where you toggle certain data points on/off.

How about implementing a parameter for choosing the desired tween?

liquidpele:
Looks like there is some contention here.... I'm hesitant to merge in tween changes without a lot of testing as tweens are notoriously hard to get right...

Issue #1585, opened 2016-04-01
Some chart: Set top margin is not working


thienhv295:
I try to set top margin for some chart, but it's not working (it work fine with right, bottom, left exclude top margin). Please check link demo below:

[https://jsfiddle.net/thienhv295/wouy3qLx/](https://jsfiddle.net/thienhv295/wouy3qLx/)

Thank you.

ovvn:
Same thing. tested in multibar chart. Margin.top isn't set up properly

scottlowe:
Same problem here. I'm using version 1.8.2, and setting the top margin appears to have no effect.

zag2art:
+1

DeWenS:
+1

fctsvirus:
+1

zheoreh:
+1

Tarlord:
Does not working in 1.8.0, 1.8.1 versions also(probably old bug)

MaestroSc:
+1

art-emm:
+1

Yska:
+1

ovvn:
I've found that if there's a legend on a chart then margin.top is being overwritten with `legend.height()` in many charts. That's why user specified top margin is ignored.
https://github.com/novus/nvd3/blob/master/src/models/cumulativeLineChart.js#L224
https://github.com/novus/nvd3/blob/master/src/models/historicalBarChart.js#L119
https://github.com/novus/nvd3/blob/master/src/models/lineChart.js#L191
https://github.com/novus/nvd3/blob/master/src/models/linePlusBarChart.js#L236
https://github.com/novus/nvd3/blob/master/src/models/multiBarChart.js#L171
https://github.com/novus/nvd3/blob/master/src/models/multiBarHorizontalChart.js#L162
https://github.com/novus/nvd3/blob/master/src/models/pieChart.js#L121
https://github.com/novus/nvd3/blob/master/src/models/scatterChart.js#L169
https://github.com/novus/nvd3/blob/master/src/models/stackedAreaChart.js#L169
https://github.com/novus/nvd3/blob/master/src/models/discreteBarChart.js#L118
@liquidpele should the top margin be overwritten in all charts with legend?

ovvn:
There're two approaches
1. **Not to overwrite the top margin & availableHeight depending on a legend's height**. In this case user has a full control over the look of the chart. So he can set a top margin in a way that will prevent the chart and legend from overlapping.
[...]

PR #1640, opened 2016-05-06
Not to overwrite top margin & availableHeight depending on a legend's height

ovvn:
Fix for #1585

### DIFF
--- a/src/models/cumulativeLineChart.js
+++ b/src/models/cumulativeLineChart.js
@@ -220,11 +220,6 @@ nv.models.cumulativeLineChart = function() {
                     .datum(data)
                     .call(legend);
 
-                if ( margin.top != legend.height()) {
-                    margin.top = legend.height();
-                    availableHeight = nv.utils.availableHeight(height, container, margin);
-                }
-
                 g.select('.nv-legendWrap')
                     .attr('transform', 'translate(0,' + (-margin.top) +')')
             }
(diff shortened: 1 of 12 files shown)


liquidpele:
I don't understand why this is a fix... does this not make the chart and the legend overlap?

ovvn:
There're two approaches
1. Not to overwrite the top margin & availableHeight depending on a legend's height. In this case user has a full control over the look of the chart. So he can set a top margin in a way that will prevent the chart and legend from overlapping.
2. Overwrite the top margin & availableHeight only in case the legend's height is bigger than the top margin. In this case user's specified top margin will be overwritten when the legend's height is bigger than the margin in order to prevent the overlapping.

I made two pull request for each of the case respectively.

liquidpele:
I think this is fixed by https://github.com/novus/nvd3/pull/1808 please confirm.

ovvn:
I don't think the PR #1808 is a right thing to do as it's actually rolls back this PR https://github.com/novus/nvd3/pull/1642
and bring us back to the situation where setting a top margin doesn't work at all https://github.com/novus/nvd3/issues/1585

liquidpele:
Yea, it does roll it back, but the original PR caused an issue.   We really need to check if the margin was overridden or not and only have it adjust if it wasn't specifically set.

ovvn:
That's right, but now we adjust a top margin anytime it's different from legend, i.e. overwrite it with the legend's height. That's why no top margin can be specified at all.

ovvn:
> We really need to check if the margin was overridden or not and only have it adjust if it wasn't specifically set.

Created a PR #1821 which implements the concept.

PR #1641, opened 2016-05-06
Fix lineWithFocusChart rendering issues

mrijke:
This commit fixes a rendering issue wit the lineWithFocusChart.
When a lineWithFocusChart initially has no data, then gets data and
update() is called, only the focus part is rendered. Only when data is
set again and update() is called again does the 'normal' linechart
show correctly.
Listening for the 'onBrush' event of the focus before initializing the
chart seems to fix the issue.

### DIFF
--- a/src/models/lineChart.js
+++ b/src/models/lineChart.js
@@ -45,7 +45,7 @@ nv.models.lineChart = function() {
     }).headerFormatter(function(d, i) {
         return xAxis.tickFormat()(d, i);
     });
-    
+
     interactiveLayer.tooltip.valueFormatter(function(d, i) {
         return yAxis.tickFormat()(d, i);
     }).headerFormatter(function(d, i) {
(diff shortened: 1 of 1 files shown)


liquidpele:
Can you explain why this works?   I don't see how re-setting the onBrush callback to the same thing it was already on every update changes anything...

mrijke:
The reason why I moved up the subscribing for the onBrush event is because the callback is not called the first time my chart renders (in https://github.com/novus/nvd3/blob/master/src/models/lineChart.js#L394) and thus my 'main' chart area stays blank. If I then refresh the data object and call chart.update(), it does work.

I believe it works now because with my fix the subscribing for events is done before the call to render the focus is made on line 256. (https://github.com/novus/nvd3/blob/master/src/models/lineChart.js#L256) 
That would explain why it works on the second call to update()

liquidpele:
Interesting, okay.

PR #1642, opened 2016-05-06
Overwrite the top margin & availableHeight only in case the legend's height is bigger than the top margin.

ovvn:
Fix for #1585


liquidpele:
Oh nice fix, thanks for contributing!

PR #1650, opened 2016-05-11
linePlusBarChart: tooltip and ticks have are flipped axis (Reverting #1518)

tlercher:
The Change reverts #1518.

The result of this change is that the `tooltip` and `ticks` are always flipped. (`y1Axis.tickFormat` used by the `tooltip` in `y2Axis` and vice versa)

Because `getBarsAxis` und `getLinesAxis` are only used by the tooltip it shouldn't affect anything else.

Example (`nvd3-community`-Example against `1.8.3`)
http://plnkr.co/edit/aJ0BLI8QPBLprXTtpokV?p=preview

Example with fix applied:
http://plnkr.co/edit/MTsl4q9fiobyHUZVUaOm?p=preview

Alternatively just revert the commit. ;)

### DIFF
--- a/src/models/linePlusBarChart.js
+++ b/src/models/linePlusBarChart.js
@@ -70,13 +70,13 @@ nv.models.linePlusBarChart = function() {
     //------------------------------------------------------------
 
     var getBarsAxis = function() {
-        return !switchYAxisOrder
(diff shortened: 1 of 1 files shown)


AltxSF:
+1 this should be merged

PR #1398, opened 2015-12-14
Use chart.interactiveLayer.tooltip.valueFormatter in lineChart [fixes #303]

apaatsio:
Fixes #303 for lineCharts.

Issue #1249, opened 2015-09-11
valueFormatter being ignored


cheapsteak:
reproduction: http://codepen.io/cheapsteak/pen/eppGrd

```js
chart.tooltip.valueFormatter(d => 'yyyyyyasdf', d => 'nothing shows up');
chart.interactiveLayer.tooltip.valueFormatter(d => 'this is getting ignored', d => 'also ignored');
```

I'm using version 1.8.1, same bug for lineChart and lineWithFocusChart

RenatoUtsch:
Same as #1247. @liquidpele please close this.

Issue #1247, opened 2015-09-09
interactive guideline tooltip custom valueFormatter


pgrm:
I've seen dozens of issues about this, but I'm not entirely sure if this part has been already mentioned:

The `valueFormatter` can be overwritten via `chart.interactiveGuideline.tooltip.valueFormatter(newCustomFormatter)` or better it could be overwritten, because the valueFormatter is being set inside the function. What that means:

after deleting the following lines:

    .valueFormatter(function(d,i) {
        return d == null ? "N/A" : yAxis.tickFormat()(d);
    })

from [lineChart.js](https://github.com/novus/nvd3/blob/master/src/models/lineChart.js#L256) and [lineWithFocusChart.js](https://github.com/novus/nvd3/blob/master/src/models/lineWithFocusChart.js#L333) I can now set the tooltip as described above. Or actually I'm using [angular-nvd3](https://github.com/krispo/angular-nvd3) for that and that needs also some more fixes, but it works.

Now I don't know what could break by removing those lines, but now it works es expected and I can define my custom format also for interactive guidelines.

So, could this be implemented? Or does it break terribly something else?

RenatoUtsch:
Does it work as expected? I thought that specifying a new value formatter would just replace the old function.

If the default value formatter functions aren't specified, what happens? How is the data formatted?

metysj:
The 'old' function is reset every time because it's inside the 'elementMousemove' handler

liquidpele:
Oh, yea that's weird... it has it setting that here:

https://github.com/novus/nvd3/blob/master/src/models/lineChart.js#L415

but, it's already being set up here:

https://github.com/novus/nvd3/blob/master/src/models/lineChart.js#L64

so I think the place inside of the mousemove handler should just be deleted?

pgrm:
it seems to be working now for me, without the fix

Issue #1210, opened 2015-08-25
Issue with Tooltip in Multichart


uenz:
The key (Name) of a line is only shown in multichart tooltip if the line is of type area. See the [MultiChart](https://raw.githubusercontent.com/novus/nvd3/master/examples/multiChart.html) sample. The keys are only shown for the first two Lines (item0 and item1). Putting a console.log(d) in keyFormatter shows undifined for all other types than area. User specific valueFormatter is not called. I tried a console.log("valueFormatter") in the valueFormatter in the offical source. It was also not called.

liquidpele:
I think I just merged in a fix for that this morning, can you check the latest code in master?

uenz:
In the master branch the files ( nv.d3.js / nv.d3.min.js) in the build 
folder are still the old ones. Unfortunately i don't now how to 
"rebuild" the build folder or
how to test the fix without a new nv.d3.js

liquidpele:
I built latest and pushed it up.

uenz:
Now it works with type area and line. Not with bar and scatter.
As a quick hack i also added the "key: evt.series.key" to 
mouseover_scatter and mousover_bar.
It works with scatter, but not with bar which have a different 
definition of evt.series evt['series'].

cheniel:
Works for me on a lineChart.

liquidpele:
Someone fixed it for the rest and I merged that, closing this.

PR #1655, opened 2016-05-12
Fixed #1156: multiChart tooltip valueFormatter is always overriden

### LINKED ISSUE #1156: multiChart tooltip valueFormatter is always overriden
Example of the relevant code from `models/multiChart.js`:

```javascript
                tooltip
                    .duration(100)
                    .valueFormatter(function(d, i) {
                        return yaxis.tickFormat()(d, i);
                    })
                    .data(evt)
                    .position(evt.pos)
                    .hidden(false);
```

The `valueFormatter` is always set on the mouseover events, which makes it impossible to set a custom `valueFormatter`. Calling `chart.tooltip.valueFormatter()` has no effect, because of the code above.

The code above is required because it determines which yAxis to get the formatter from. However, it would be possible to make this behaviour optional, or use custom formatters which accept the yAxis as an argument and return a function. Are you open to a pull request implementing the latter?

juhamust:
Fixes issue #1156 by using similar approach as in #1398 
Related to #1247, #1249, #1210

### DIFF
--- a/src/models/multiChart.js
+++ b/src/models/multiChart.js
@@ -419,15 +419,17 @@ nv.models.multiChart = function() {
                         });
                     });
 
+                    var defaultValueFormatter = function(d,i) {
+                        var yAxis = allData[i].yAxis;
+                        return d == null ? "N/A" : yAxis.tickFormat()(d);
+                    };
+
                     interactiveLayer.tooltip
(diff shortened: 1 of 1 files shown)


liquidpele:
Thanks!

Issue #1532, opened 2016-02-23
Tooltip Position Wrong when the chart is in a dialog


rockydd:
The tooltip will have a wrong offset when the chart is in a bootstrap dialog. This issue is found in 1.8.2. While It works well in 1.8.1.

<img src="http://i.imgur.com/xVdrL1J.jpg?1">

rockydd:
OK, found the root cause. Now the tooltip was translated based on the position returned by function position() use d3.event.clientX. However, when the parent container has 'transform' style, then the start point for the fixed position will be the left-top corner of the parent container instead of the browser window.

laxu:
Same issue happens when using Foundation dialogs. Is there anything we can do to fix this until a new version is out aside from forking the project and changing the code there?

smills2929:
I have the same issue as I am simply using translations on the chart containers. Has anyone figured out a work-around?

liquidpele:
Can you submit a jsfiddle that demonstrates the problem, and link to the latest version of nvd3 in the master branch using rawgit.com ?

smills2929:
Here you go. Works fine with a translate on body in 1.8.1. Is broken on 1.8.2 or latest commit.
1.8.1: https://jsfiddle.net/oxvygewj/
latest commit (c2ea2cb): https://jsfiddle.net/yxn42srr/

nickcoutsos:
Is this or a related issue being looked at? I just spent an hour trying to clean up some CSS hacks my boss used to work around this issue and it seems like the alternative is just downgrading to 1.8.1, except apparently 1.8.2 introduces a fix for something else.

liquidpele:
Updated jsfiddle to show latest master code, still looks to be an issue.  

https://jsfiddle.net/yxn42srr/1/

nickcoutsos:
I did some digging in the history for the tooltip module, and the biggest red flag was #1190. There are a lot of changes for a newcomer to try to make sense of, but (if I've done this correctly), here's are jsfiddles using builds closest to that merge:

* [before](https://jsfiddle.net/ub1ranrr/2/) (3ba48c02b96d5fdc395beb0cd36102b271a2bf0c)
* [after](https://jsfiddle.net/ub1ranrr/1/) (fbdcb94c7b54fd22f52e7d354f67526af8a1eb05)

[...]

PR #1673, opened 2016-05-21
Fix for #1532.

RenatoUtsch:
Stumbled upon a good fix for this in a side project.

Fiddle demonstrating the fix: https://jsfiddle.net/genhmq3v/

More testing for the tooltip positioning would be a good idea.


RenatoUtsch:
Fixes #1532

RenatoUtsch:
Ugh this broke the scroll tests I wrote... let me work some more on this.

PR #1674, opened 2016-05-22
Fix bug with tooltip positioning.

### LINKED ISSUE #1532: Tooltip Position Wrong when the chart is in a dialog

RenatoUtsch:
FIxes #1532.

This fix corrects the position of the tooltip based on the translate property of the parents.

https://jsfiddle.net/genhmq3v/1/

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -27,7 +27,6 @@ nv.models.tooltip = function() {
         ,   distance = 25 // Distance to offset tooltip from the mouse location.
         ,   snapDistance = 0   // Tolerance allowed before tooltip is moved from its current position (creates 'snapping' effect)
         ,   classes = null  // Attaches additional CSS classes to the tooltip DIV that is created.
-        ,   chartContainer = null // Parent dom element of the SVG that holds the chart.
         ,   hidden = true  // Start off hidden, toggle with hide/show functions below.
         ,   hideDelay = 200  // Delay (in ms) before the tooltip hides after calling hide().
         ,   tooltip = null // d3 select of the tooltip div.
@@ -37,42 +36,6 @@ nv.models.tooltip = function() {
         ,   headerEnabled = true // If is to show the tooltip header.
         ,   nvPointerEventsClass = "nv-pointer-events-none" // CSS class to specify whether element should not have mouse events.
     ;
-
-    /*
-     * If a parent of the container has a translate transformation, fix the
-     * positioning. The fixed position of the tooltip will be based starting
-     * at this element.
-     */
-    var fixTranslation = function(pos) {
-        var obj = chartContainer;
-
-        while(obj && obj.style) {
-            if(getComputedStyle(obj).transform != 'none') {
-                var rect = obj.getBoundingClientRect();
-                pos.left -= rect.left;
-                pos.top -= rect.top;
-                break;
-            }
-
-            obj = obj.parentNode;
-        }
-
-        return pos;
-    };
-
-    /*
-     Function that returns the position (relative to the viewport) the tooltip should be placed in.
-     Should return: {
-        left: <leftPos>,
-        top: <topPos>
-     }
-     */
-    var position = function() {
-        return fixTranslation({
-            left: d3.event !== null ? d3.event.clientX : 0,
-            top: d3.event !== null ? d3.event.clientY : 0
-        });
-    };
 
     // Format function for the tooltip values column.
     var valueFormatter = function(d, i) {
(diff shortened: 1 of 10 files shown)


RenatoUtsch:
travis seems to be failing because of the environment the tests are running, not because of the PR.

workgena:
I have the same issue. And this PR does not solve my problem.

Also you have this commit
https://github.com/RenatoUtsch/nvd3/commit/f1754e35e958658974cbf7261c80ec9a01dcf52c

In my case, this commit fixes issue with tooltip positioning.

RenatoUtsch:
@workgena the problem is that that commit breakes other things.

Can you please create a jsfiddle to show your issue, so that I can work on it?

workgena:
Found strong dependence on bootstrap version.

Boostrap 2.3.1
http://jsfiddle.net/pL9sdgLu/1/

Boostrap 3.3.6
http://jsfiddle.net/bg5gsm2a/7/

This is CSS issue with bootstrap v3.3.6. Tooltip position has exact offset as has modal. To see it more clearly - resize demonstration window.

RenatoUtsch:
That jsfiddle worked for me. What browser are you using?

workgena:
Same browsers on OS X
good: http://awesomescreenshot.com/0845wb86a8

with error: http://awesomescreenshot.com/03e5wb7t2c

RenatoUtsch:
Okay, I saw it. Will take a look at it later.

RenatoUtsch:
Thanks, I made an even better fix because of what you said. This one now will probably work in all cases. Your example is working now.

I can't find any more problems with this PR, have everyone's problems been solved?

RenatoUtsch:
Again, all checks have passed in my machine, this is a problem with the travis environment.

@liquidpele

liquidpele:
Cool, will look at some PRS tonight.

workgena:
Applied this PR on my project and it works fine. 
Checked StackedArea and MultiBar charts.
Thx @RenatoUtsch

It would be good to have nvd3 1.8.4 with this fix.

src/tooltip.js
```diff
@@ -39,17 +39,39 @@ nv.models.tooltip = function() {
...
+
+        while(obj && obj.style) {
+            if(getComputedStyle(obj).transform != 'none') {
+                var rect = obj.getBoundingClientRect();
+                pos.left -= rect.left;
+                pos.top -= rect.top;
+                break;
+            }
+
+            obj = obj.parentNode;
```
liquidpele: This is always going to walk up the whole object tree every time it adjusts the tooltip position... which is a lot...  What was getBoundingClientRect().top and left doing that was wrong exactly?  From what I can tell it is supposed to take translate into account, there's even a bugzilla bug about it that they fixed at one point.

RenatoUtsch: It is always going to walk the object tree, but it will stop in the case of
a translation. The problem is that if there is a parent that has a css
translation, the fixed positioning of children will be based on the
position of this parent, not the screen. And as d3.event.client* is based
on the position of the screen, it has to take into account the translation
of this parent. The complexity of walking up the tree is not good, I agree.

I had a good idea.  We can just look at the parent of the tooltip, and if
it doesn't have a translation property, we add a translate(0,0). We would
only have to take into account the immediate parent of the tooltip and it
would always work.

I'll do this later today, thanks for the good idea @liquidpele.

Em Qui, 26 de mai de 2016 11:23 AM, liquidpele <notifications@github.com>
escreveu:

RenatoUtsch: A less invasive approach would be to put the nvd3 chart and the tooltip
inside a new div, and add a translate(0,0) to the new div.  This would
ensure that adding the additional translation wouldn't break the rest of
the code of the user that depends on fixed positioning.

What do you think?

Em Qui, 26 de mai de 2016 12:04 PM, Renato Utsch <renatoutsch@gmail.com>
escreveu:

RenatoUtsch:
From what I've researched, we have only two choices to avoid having to walk up the DOM tree:

1. Add a 'translate(0,0)' to the tooltip parent if it doesn't have a translate transformation, so that we only need to look at it instead of the whole tree. The problem with this is that we're changing a div that we didn't create, and this can (and probably will) break user code.
2. Make the tooltip div a direct child of document.body, so that we only have to look if the document.body has a transformation. The problem with this is that we're not putting the tooltip together with the SVG.

I personally prefer the 2nd choice. First, it won't break any user code (except code that called chartContainer()? But I don't see a good use for changing it except for internal nvd3 code). Second, NVD3 originally worked like that, with the tooltip directly inside document.body, so we'll just be regressing to the original behavior (but much better now, the old code was a mess).

I'll implement the 2nd choice and update this PR okay? Anybody against it? @liquidpele?

workgena:
I prefer putting tooltip into document.body. Many similar things(popup, modals, etc.) in other js-libs works this way.

RenatoUtsch:
Done. Now the tooltip is directly into document.body. All the tests passed in my machine.

The performance is not a problem anymore @liquidpele.

liquidpele:
Awesome, good job!

RenatoUtsch:
Maybe a new 1.8.4 release would be good now, to finally put an end to all these tooltip-related bugs =)

PR #1677, opened 2016-05-23
Pass more data to color method to allow greater customization and consistency

manbearwiz:
This fix passes the entire data object and index into the color function rather than just the name. In addition to to allowing more intelligent color functions, this change makes the API more consistent with the pie chart.

The motivation for this change is to be able to use an attribute of the data passed in to generate the fill colors for the sections in order to signify intensity.

BREAKING CHANGE:

Any existing sunburst chart that uses a custom color function will need to be updated to the new API; however, the name is still passed to this function as an attribute of the object. Migration is as simple as changing 

```
function(name) {
  // ...
}
```
to
```
function(data) {
  var name = data.name;
  // ...
}
```

I would also like to note that in the current state, it seems unlikely anyone would be using a node name to do any sort of color calculation/generation.

### DIFF
--- a/src/models/sunburst.js
+++ b/src/models/sunburst.js
@@ -235,15 +235,15 @@ nv.models.sunburst = function() {
 
             cGE.append("path")
                 .attr("d", arc)
-                .style("fill", function (d) {
+                .style("fill", function (d, i) {
                     if (d.color) {
                         return d.color;
                     }
                     else if (groupColorByParent) {
-                        return color((d.children ? d : d.parent).name);
+                        return color((d.children ? d : d.parent), i);
                     }
                     else {
-                        return color(d.name);
+                        return color(d, i);
                     }
                 })
                 .style("stroke", "#FFF")


src/models/sunburst.js
```diff
@@ -235,15 +235,15 @@ nv.models.sunburst = function() {
...
                     if (d.color) {
                         return d.color;
                     }
                     else if (groupColorByParent) {
-                        return color((d.children ? d : d.parent).name);
+                        return color((d.children ? d : d.parent), i);
                     }
                     else {
-                        return color(d.name);
+                        return color(d, i);
```
liquidpele: The default color function is nv.utils.defaultColor, which uses nv.utils.getColor which has a definition of just function(color).  While it might work, it's going to hash a whole object every time, which is going to end up generating different colors if the data structure is modified.  

Why would you be changing/duplicating the name attribute anyway?  If you really have to do that, then I suggest putting a unique number on the end of the name and then strip it off with a custom tooltip.headerFormatter option

manbearwiz:
This change make it consistent with the pie chart. Correct me if I'm wrong, but any disadvantage to this implementation would already be evident on the pie chart then. From [pie.js](https://github.com/novus/nvd3/blob/master/src/models/pie.js#L205):

```
slices.attr('fill', function(d,i) { return color(d.data, i); });
```

My general use case is to pass in objects like 

```
{
  name: "Wisconsin",
  count: 5758,
  temp: 67
}
```

and then use a d3 linear color scale to color segments by the temp field. Any suggestions on a better way to achieve this then passing that object into the color function?

liquidpele:
I'll have to think about this one, I don't much care for passing in objects as keys but it would be nice to be able to reference the full object too...

PR #1686, opened 2016-06-06
Fix #1685 - Added min/max barWidth to the discreteBar.

### LINKED ISSUE #1685: DiscreteBar: Add support for min/max bar width

mpetkov:
Proposed solution to #1685

### DIFF
--- a/src/models/discreteBar.js
+++ b/src/models/discreteBar.js
@@ -9,6 +9,8 @@ nv.models.discreteBar = function() {
     var margin = {top: 0, right: 0, bottom: 0, left: 0}
         , width = 960
         , height = 500
+        , maxBarWidth
+        , minBarWidth
         , id = Math.floor(Math.random() * 10000) //Create semi-unique ID in case user doesn't select one
         , container
         , x = d3.scale.ordinal()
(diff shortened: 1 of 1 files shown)


src/models/discreteBar.js
```diff
@@ -49,6 +51,14 @@ nv.models.discreteBar = function() {
...
                     point.series = i;
                 });
+
+                // Changes the availableWidth to allow for a minimum bar width
+                if(minBarWidth) {
+                    var minWidth = series.values.length * minBarWidth * 1.1;
+                    if(availableWidth < minWidth) {
+                        availableWidth = minWidth;
+                    }
+                }
```
liquidpele: availableWidth is used by more than the bar calculation, this should only be changing the actual bar position and width.

liquidpele: Oh wait, I see what this is doing...  but it should still not change availableWidth I don't think...   since the bar charts currently make the bars as thick as they can be given the width, why would you ever want to force them to be bigger like this?

nqbao: agree, the minBarWidth should be constraint by the availableWidth

liquidpele:
closing this as-is, will reopen if concerns are addressed

azlenko:
Hi, if it's ok with everyone, I would include maxBarWidth option. It's a very desired feature, because when there is only one bar on discrete bar chart it looks too thick.
So my proposal is to remove all code related to minBarWidth option and leave maxBarWidth.
I can do it myself in a new PR just copying the @mpetkov code if he/she doesn't follow this thread anymore.

mpetkov:
Hey all, so the minBarWidth could have been done several different ways and it is just the way I decided to implement it at the time.  As @azlenko pointed out, it is a needed feauture because when you have one bar it is very thick.

PR #1693, opened 2016-06-10
Fixes  #1692 (isMSIE bugfix breaks tooltip in IE11)

### LINKED ISSUE #1692: isMSIE bugfix breaks tooltip in IE11
**Description:**
The tooltip displayed is allways for the left-most column of the bar chart. Even tho the tooltip itself is
displayed where the cursor actually is, the data is gathered from the wrong column.

image

**To reproduce:** 
-Using browser IE11 (same goes for debugging as IE 9, 10 and edge)
-Have barchart with multiple bars and tooltip
-Hover over with mouse
-Mouseover will allways display tooltip-data for the first bar

**Mitigating the problem:**
-Disabling the isMSIE check, by setting isMSIE to false in nv.d3.js (line 201 for version 1.8.3) makes the problem dissapear. 

It seems that the bugfix for IE was made for IE10, and that IE11 still triggers the bugfix even if it is not affected by the error. And as such it only messes things up

I would suggest changing the check to one that excludes IE11

studiefredfredrik:
Check for activeXobject on window object, will return false for IE11

### DIFF
--- a/src/interactiveLayer.js
+++ b/src/interactiveLayer.js
@@ -17,7 +17,7 @@ nv.interactiveGuideline = function() {
         ,   showGuideLine = true
         ,   svgContainer = null // Must pass the chart's svg, we'll use its mousemove event.
         ,   tooltip = nv.models.tooltip()
-        ,   isMSIE = "ActiveXObject" in window // Checkt if IE by looking for activeX.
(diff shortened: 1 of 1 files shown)


liquidpele:
Interesting, thanks!

PR #1694, opened 2016-06-13
linePlusBarChart: need to pass along "strokeWidth" while rendering line

mrvicadai:
@liquidpele while trying to customize the line width in a linePlusBarChart, I noticed the `stroke-width` of the line is unaffected although the property `strokeWidth` is set on the chartData. This PR should fix the bug.


liquidpele:
ah cool, thanks for contributing!

liquidpele:
Also, could you add that info to the wiki page about data attributes? 

mrvicadai:
sure, it's [done](https://github.com/novus/nvd3/wiki/Data-Attributes#set-the-width-of-a-line)

PR #1718, opened 2016-07-06
Make multiBarChart use the same tooltip regardless of whether or not interactiveGuideline is enabled.

mackler:
It seems that when `useInteractiveGuideline` is true in a multiBarChart, the tooltip belongs to the interactive layer, rather than the chart, causing customizations made to the chart to be missing from the tooltip as it is displayed.

At least that's how it seems to me, but I am very new to this library and probably might be wrong about everything.  In any event, this pull request is my suggestion for a fix that, again, seems to work to me.

**Warning** this is only a partial solution.  I know nothing about _JavaScript_ testing so I only made the changes to the multiBarChart.  The other charts use also the tooltip, so I'm sure they would be broken by this change.  Maybe this should be an issue rather than a pull request.

I strongly encourage someone who knows more that I do to understand what is actually causing the phenomenon that this pull request attempts to address and to implement an actually correct solution.

### DIFF
--- a/src/interactiveLayer.js
+++ b/src/interactiveLayer.js
@@ -211,7 +211,12 @@ nv.interactiveGuideline = function() {
     }
 
     layer.dispatch = dispatch;
-    layer.tooltip = tooltip;
+
+    layer.tooltip = function(_) {
+        if (!arguments.length) return tooltip;
+        tooltip = _;
+        return layer;
+    };
 
     layer.margin = function(_) {
         if (!arguments.length) return margin;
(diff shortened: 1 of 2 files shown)


liquidpele:
That was mostly intentional, the interactive layer has it's own tooltip object that it controls since the data is going to be displayed differently between one item and many items and it can be customized differently that way as well.   You can customize the layer's tooltip by just doing:   chart.interactiveLayer.tooltip.someOption()

PR #1699, opened 2016-06-20
Add markerLines and markerLineLabels to bullet chart

olivermue:
Adds the ability to show a simple vertical line as marker on a bullet chart.

PR #1721, opened 2016-07-07
Move marker line in bullet chart on redraw

balmma:
The new marker line code merged in from pull request #1699 some days ago does not properly handle redraws of the marker line. The selectAll tries to find a path element although the line is a line element.

Furthermore setting the transform attribute does not make any sense here. Setting x1 and x2 works far better.


liquidpele:
Looks okay to me, thanks!

PR #1725, opened 2016-07-11
Fix translate exception in multiChart

balmma:
The current code throws an Exception when setting the transform attribute:

d3.min.js:1Error: <g> attribute transform: Expected transform function, "NaN,5)".

This seems to be caused by a missing + in the string concatenation.

### DIFF
--- a/src/models/legend.js
+++ b/src/models/legend.js
@@ -34,7 +34,7 @@ nv.models.legend = function() {
             var g = wrap.select('g');
 
             if (rightAlign)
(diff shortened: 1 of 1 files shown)


balmma:
Don't really get why the tests are failing. Any pointers?

balmma:
It seems that Travis has a problem. Other pull requests are failing as well (e.g. #1726).

rlrosa:
@balmma , i think the "NaN,5)" is fixed by my pull   #1760

liquidpele:
Fixed in another PR.

PR #1729, opened 2016-07-14
Implement Sankey charts for NVD3

ksgy:
This PR is a simple implementation of the Sankey Diagram by 
- https://bost.ocks.org/mike/sankey/
- https://github.com/soxofaan/d3-plugin-captain-sankey

The PR has:
- Example page with a simple and an advanced version of Sankey chart
- Basic `mocha` tests for `svg` structure and attributes
- Sankey model
- Sankey chart with error handling for "no data" and data format check


src/models/sankeyChart.js
```diff
@@ -0,0 +1,239 @@
...
+    };
+
+
+    //============================================================
+
+    function chart(selection) {
+        selection.each(function(data) {
+
+            //============================================================
+            // Error handling
```
liquidpele: What's with these random comments that don't belong?     Please just remove any unhelpful comments, including just random lines of equal signs :)

src/models/sankey.js
```diff
@@ -0,0 +1,343 @@
...
+        size = [1, 1],
+        nodes = [],
+        links = [],
+        sinksRight = true;
+
+    sankey.nodeWidth = function(_) {
+        if (!arguments.length) return nodeWidth;
+        nodeWidth = +_;
+        return sankey;
+    };
```
liquidpele: It looks like the options on this file aren't using the chart._options format, which makes it so they can't be inherited automatically, please make this fit the same format as the rest of nvd3.

src/models/sankeyChart.js
ksgy: Seen in https://github.com/novus/nvd3/blob/master/src/models/line.js - thought make the code more readable, but will remove :)

src/models/sankey.js
ksgy: Missed that one, will do that

src/models/sankeyChart.js
liquidpele: I tried to remove them all, guess I missed some  :p 

originally there were like 20 of those per file with no real reasoning as to where they were, so it actually impeded readability.

src/models/sankey.js
ksgy: @liquidpele Refactored using the `_options`but not sure if that's the right way to pass and call functions, I haven't seen any kind of similar "callback" function to start the rendering for a chart.
See the last 3 properties: https://github.com/ksgy/nvd3/blob/master/src/models/sankey.js#L309-L311

ksgy:
Hey @liquidpele any news on this pull request?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "nvd3",
-  "version": "1.8.4-dev",
+  "version": "1.8.5-dev",
```
liquidpele: Wait... why is this changing the version?

ksgy: I might be wrong, but as far as I remember, all previous new features increased the version, but I can change it back :)

liquidpele: Nope, please change it back.

PR #1741, opened 2016-07-23
fix #1740: Wrong interactiveGuideline tooltip header with multiple datasets that span different domains in lineChart

### LINKED ISSUE #1740: Interactive guideline tooltip header getting stuck on last element of first dataset when multiple datasets are added dynamically with different domains
**Intro**: I initialize a line chart with the interactive guideline and some data that goes from x = [0, 4], then I add some data with a bigger domain, x = [0, 8]

**Bug**: The interactive tooltip header will still show x = 4 even when I mouseover in the range between x = [4, 8]

**Steps to reproduce**: http://plnkr.co/edit/5tvHKdNvRfUzncvR8z2s?p=preview

Click on "Add data" button, mouseover in the region between x = 4 and x = 8, you will see that the tooltip header always shows 4 and doesn't update. The y values are correct, though.

If the data sets are reversed (ie I add the bigger dataset first) then there's no problem.

marenaud:
Fix for issue #1740 

I don't quite have time to write tests but I just wanted to submit the PR to show how I fixed the bug on my own build in case it inspires anyone. Basically the code iterates through each series but only sets the "singlePoint" variable once if it's undefined. If there's many data series and the later ones have a larger domain, the singlePoint variable from the earlier series will be passed to the tooltip (on line 325) and the wrong value of "X" will be shown on the header.

The issue may exist for other types of charts, I will verify that later.

### DIFF
--- a/src/models/lineChart.js
+++ b/src/models/lineChart.js
@@ -278,6 +278,7 @@ nv.models.lineChart = function() {
             interactiveLayer.dispatch.on('elementMousemove', function(e) {
                 lines.clearHighlights();
                 var singlePoint, pointIndex, pointXLocation, allData = [];
+                var maxSeriesSize = -1;
                 data
                     .filter(function(series, i) {
                         series.seriesIndex = i;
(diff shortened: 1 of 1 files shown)


marenaud:
A quick search shows the similar behaviour in cumulativeLineChart, multiChart, stackedAreaChart, multiBarChart and historicalBarChart, though I'm not sure if the notion of multiple datasets with different domains makes sense in all those chart types.

liquidpele:
interactive guideline requires all the series to have a value for each index of the array and for them to have the same number of items, othewise it can't calculate the correct correct element to highlight on each series.

liquidpele:
To be more clear, while your solution fixed a particular part of that limitation, it was not a complete fix for all cases.   Try adding null as the value for any place a series doesn't have a value.

Issue #1525, opened 2016-02-19
Line chart example does not highlight some series points


mlawry:
In regards to the [lineChart.html](https://github.com/novus/nvd3/blob/master/examples/lineChart.html) example, use the legend to disable the middle series "Random Points". Then when you mouse over the chart, the "Random Cosine" series will not have its data points highlighted (there is no circle dot like the others). Try disabling other series using the legend and this problem still appears.

It looks like a problem with [lineChart.js](https://github.com/novus/nvd3/blob/master/src/models/lineChart.js#L399):

     lines.highlightPoint(series.seriesIndex, pointIndex, true);

[It used to be](https://github.com/novus/nvd3/blob/e2cd4d4e6bcc4745ed5546a5af6525ade194b6b0/src/models/lineChart.js#L366)

    lines.highlightPoint(i, pointIndex, true);

and the `i` was changed to `series.seriesIndex`, which doesn't look to be correct because when there are disabled series, `i` covers only the enabled series whereas `series.seriesIndex` covers all enabled and disabled series. As a result, for some series, `i` will have a value different to `series.seriesIndex`.

Changing `series.seriesIndex` back to `i` seems to fix the problem.

almike:
yeah same here

maitreyavvm:
+1

PR #1743, opened 2016-07-25
fix: revert highlight point index value

almike:
this is to fix https://github.com/novus/nvd3/issues/1525


liquidpele:

PR #1745, opened 2016-07-26
Enable label support in StackedArea and Line

### LINKED ISSUE #479: How to add Annotation to the graph ?
Is this possible to add annotation to nvd3 graphs ?

zipengwu:
This commit tries to resolve #479

- Label will be automatically added to scatter point, when values have label data
- For example {x: 2, y: 3, label: "label"}


liquidpele:
I don't think that's a good default, but you can set it already by doing chart.scatter.showLabels(true);

PR #1758, opened 2016-07-30
Fix chart resize bug in scatter model optimisation

ThomasBrierley:
Add a sizeDiff bool that explicitly checks for changes in chart size. The scaleDiff bool used to trigger a full redraw in the optimisation from PR #1378 is not always true when chart size changes.


ThomasBrierley:
Full details of bugs and example are in PR #1378

PR #1760, opened 2016-08-02
fix - d3.min.js:1Error: attribute transform: Expected transform function, "NaN,5)".

rlrosa:
this fixes legend translate error: string-int is NaN, should be string+(-int).


liquidpele:
Thanks !

PR #1762, opened 2016-08-04
Bugfix: Select unique tooltip by id

mehranshakeri:
Selecting tooltip with class `selectAll('.nvtooltip')` returns all existing tooltips which leads to sharing the same instance despite having unique id per each tooltip instance.
This can be fixed by selecting its own id per tooltip. `select('#'+id)`

- new pull request for #1629


liquidpele:
Awesome, thanks!

RenatoUtsch:
Why do you want to have different tooltips between charts? The behavior was as expected: the tooltips were reused between charts because you can only mouseover at one chart at the same time. You clutter the body if you create one tooltip per chart.

The bug here was not that the tooltip is reused between charts, but that one of the two different types of chart (interactive guidelined vs not) is deleting the tooltip element.

This should be reverted and the correct bug fixed. This would also solve #1886.

PR #1786, opened 2016-08-19
Multichart Line/Scatter points with focus/context functionality

nscalf561:
We found a use case for lines and scatters to have 2 y axes and focus/context functionality, a combination which we didn't see in the individual charts, so we built it into the multichart model.  This still has bars and stacks in it and they still work as expected in the standard multichart model, but they don't work with the context.  There are todo's marking what still needs to be added to make all components of multichart have full functionality with focus/context, but once tweaks are done to this (which are outside of our current scope), this would had some key functionality to multichart.  

This was originally branched off of v1.8.2 multichart.  
@liquidpele let me know what you think.

### DIFF
--- a/examples/multiChartWithFocus.html
+++ b/examples/multiChartWithFocus.html
@@ -1,0 +1,75 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <link href="../build/nv.d3.css" rel="stylesheet" type="text/css">
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.2/d3.min.js" charset="utf-8"></script>
+    <script src="../build/nv.d3.js"></script>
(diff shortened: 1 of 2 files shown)


liquidpele:
Was there not a way to use the new focus object like lineChart does?

nscalf561:
So when we were building this we were building for v1.8.2 because we were hitting some bugs in a later version.  Because of this, we wanted to build based off what was available in v1.8.2, which didn't have the new focus object.

liquidpele:
Well... I'd prefer we get bugs in current master fixed.

liquidpele:
If you refactor this to be based off master and the new focus object then I'd be glad to merge it, but as-is it's not inline with the current code...

liquidpele:
Closing this as it doesn't utilize the new focus object.

PR #1791, opened 2016-08-23
Fix renderEnd not being rendered when showLabels is false 

niconorsk:
When showLabels is false, there would no longer be any watchTransition calls and as a result renderEnd would never be triggered.

### DIFF
--- a/src/models/pie.js
+++ b/src/models/pie.js
@@ -210,7 +210,7 @@ nv.models.pie = function() {
             });
 
             slices.select('path')
-                .transition()
+                .watchTransition(renderWatch, 'pie labels')
                 .duration(duration)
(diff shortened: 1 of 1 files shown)


niconorsk:
Ok, I have tried and failed to fix the test failures and I'm finding myself stumped.  Possibly a misunderstanding of how the different components are supposed to interact together

liquidpele:
Well... I can't merge if it causes failing tests...

liquidpele:
You might want to merge in latest upstream master to your branch, I think someone fixed some stuff with the tests.   That might resolve the issue here too?

liquidpele:
ping

liquidpele:
Closing due to no reply.

Issue #1804, opened 2016-09-05
scatter chart - shapes with no fill?


alinegui:
I want to use different shapes and colors in the chart - both with filling and without.
Tried something like this but it doesn't work:
{
 key: 'test',
 color: 'red',
 values: [...],
 area: false
} 
Also tried to look for a solution of different border and filling color but didn't find (like fill option in d3)...

Is this possible?
Thank you.

PR #1805, opened 2016-09-06
Added support for scatter chart to change border color

alinegui:
#1804

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -10,7 +10,7 @@ nv.models.scatter = function() {
         , width        = null
         , height       = null
         , color        = nv.utils.defaultColor() // chooses color
-        , borderColor  = null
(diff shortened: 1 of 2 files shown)


src/models/scatter.js
```diff
@@ -10,6 +10,7 @@ nv.models.scatter = function() {
         , width        = null
         , height       = null
         , color        = nv.utils.defaultColor() // chooses color
+        , borderColor  = null
```
liquidpele: This option is not descriptive of what it does, it makes it sound like the whole chart border color.  Other options on the points are prefixed with point, so this should probably be something like "pointBorderColor"

You'll also want to add a description and example for this new option into documentation.html in the big json object at the bottom, so that it shows up on the docs page.

liquidpele:
Looks good, thanks!

PR #1806, opened 2016-09-06
fixing multiBar tooltip formatting

emirhanmarlali:
allowing users to set valueFormatter for interactiveLayer.tooltip (it was previously being overwritten)

This change should also be implemented on other charts that use interactiveLayer.tooltip


emirhanmarlali:
I added this part according to this [commit](https://github.com/novus/nvd3/pull/1001/commits/662b46ce6737839009f9477d5a2aabef7102a48b)

liquidpele:
Thanks for contributing!

PR #1810, opened 2016-09-08
fix: Line chart interactive guideline error when extent is not in ascending order

arfsb:
Fixes a “TypeError: Cannot read property 'x' of undefined” error when
using the interactive guideline for line charts if the extent is not in
ascending order.

In our use case, we reversed the x-axis on the line chart by reversing
the x domain.  This caused series.values.filter to filter out all our
values and result in the TypeError when calling interactiveBisect  A
check was added to ensure the extent is in ascending order.

### DIFF
--- a/src/models/lineChart.js
+++ b/src/models/lineChart.js
@@ -285,11 +285,13 @@ nv.models.lineChart = function() {
                     })
                     .forEach(function(series,i) {
                         var extent = focusEnable ? (focus.brush.empty() ? focus.xScale().domain() : focus.brush.extent()) : x.domain();
-                        if (extent[0] > extent[1])
-                            extent.reverse();
-
                         var currentValues = series.values.filter(function(d,i) {
-                            return lines.x()(d,i) >= extent[0] && lines.x()(d,i) <= extent[1];
+                            // check if point is between the extent
+                            if(extent[0] <= extent[1]) {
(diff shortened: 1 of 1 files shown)


liquidpele:
Why wouldn't it be in ascending order?   All the charts require that the passed in data has the x values in ascending order...

arfsb:
As an example, we have the following data points:
`[{x: 0, y:10}, {x:1, y:20}, {x:2, y:30}]`

The x-axis would normally plot from 0 to 2 (left to right).  We want it to plot from 2 to 0.  We were able to do this by setting the x-axis's domain to [2,0].  The line chart renders correctly but with useInteractiveGuideline set to 'true', we get the TypeError when we mouseover the chart.

This actually worked ok in v1.8.1 but we recently upgraded to 1.8.4 and ran into this issue.

I'm definitely open to suggestions if there is another way to reverse the x-axis without manually manipulating the domain.

liquidpele:
I thought about this for a while and decided that while it's very much an edge case, the change is small enough that I'm not very worried about it.  However, it needs comments there explaining why it is necessary so that someone doesn't just remove what looks like code without a point.

arfsb:
@liquidpele I changed the implementation.  Hopefully, it's clearer.

liquidpele:
Yea that change is fine, but please also add comments before your changes explaining why they were necessary.

PR #1817, opened 2016-09-17
Fix point update bug in scatter model

### LINKED ISSUE #1814: Stacked Area Chart 'expanded' highlights not updated on data change or series toggling
If I change the data for the chart when in expanded mode the highlighted points will still be for the initial data. The text percentages are correct though. I believe that 'point.display.y' in stackedAreaChart.js line 422 is not being updated.

Here is the initial data:
image

And here it is after changing the data:

image

ThomasBrierley:
When determining if point positions need updating the optimisation in the scatter model checks for differences with "x" and "y" point keys. Alternative keys "0" and "1" appear to be used in some cases too.

Fixes #1814

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -57,17 +57,14 @@ nv.models.scatter = function() {
         , needsUpdate = false // Flag for when the points are visually updating, but the interactive layer is behind, to disable tooltips
         , renderWatch = nv.utils.renderWatch(dispatch, duration)
         , _sizeRange_def = [16, 256]
-        , _caches
+        , _cache = {}
         ;
 
     function getCache(d) {
-        var cache, i;
-        cache = _caches = _caches || {};
-        i = d[0].series;
-        cache = cache[i] = cache[i] || {};
-        i = d[1];
-        cache = cache[i] = cache[i] || {};
-        return cache;
+        var key, val;
+        key = d[0].series + ':' + d[1];
+        val = _cache[key] = _cache[key] || {};
+        return val;
     }
 
     function getDiffs(d) {
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -67,6 +67,12 @@ nv.models.scatter = function() {
         return val;
     }
 
+    function delCache(d) {
+        var key, val;
+        key = d[0].series + ':' + d[1];
+        delete _cache[key];
+    }
+
     function getDiffs(d) {
         var i, key, val,
             cache = getCache(d),
@@ -445,7 +451,7 @@ nv.models.scatter = function() {
                     .type(function(d) { return getShape(d[0]); })
                     .size(function(d) { return z(getSize(d[0],d[1])) })
             );
-            points.exit().remove();
+            points.exit().each(delCache).remove();
             groups.exit().selectAll('path.nv-point')
                 .watchTransition(renderWatch, 'scatter exit')
                 .attr('transform', function(d) {
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -452,12 +452,14 @@ nv.models.scatter = function() {
                     return 'translate(' + nv.utils.NaNtoZero(x(getX(d[0],d[1]))) + ',' + nv.utils.NaNtoZero(y(getY(d[0],d[1]))) + ')'
                 })
                 .remove();
+            // Update points position only if "x" or "y" have changed
             points.filter(function (d) { return scaleDiff || sizeDiff || getDiffs(d, 'x', getX, 'y', getY); })
                 .watchTransition(renderWatch, 'scatter points')
                 .attr('transform', function(d) {
                     //nv.log(d, getX(d[0],d[1]), x(getX(d[0],d[1])));
                     return 'translate(' + nv.utils.NaNtoZero(x(getX(d[0],d[1]))) + ',' + nv.utils.NaNtoZero(y(getY(d[0],d[1]))) + ')'
                 });
+            // Update points appearance only if "shape" or "size" have changed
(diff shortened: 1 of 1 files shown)


src/models/scatter.js
```diff
@@ -455,7 +455,7 @@ nv.models.scatter = function() {
                     return 'translate(' + nv.utils.NaNtoZero(x(getX(d[0],d[1]))) + ',' + nv.utils.NaNtoZero(y(getY(d[0],d[1]))) + ')'
                 })
                 .remove();
-            points.filter(function (d) { return scaleDiff || sizeDiff || getDiffs(d, 'x', 'y'); })
+            points.filter(function (d) { return scaleDiff || sizeDiff || getDiffs(d, 'x', 'y') || getDiffs(d, 0, 1); })
```
liquidpele: Actually this is still not correct because x and y can really be any attribute, and should be controlled with the getX and getY functions...

ThomasBrierley: Ah yes, good shout. I think I was trying to prevent layering up unnecessary function calls when I first added this (after all it's supposed to be an optimisation). But I did not notice the scoped getX/Y has a public setter on 607/8... besides I've now added another function call anyway.

ThomasBrierley: Actually it's the same for "size" and "shape", I think I chose the direct approach because it makes the cache function simpler, i'm going to have to re-think that part a bit to allow for those custom getters.

liquidpele: You could just pass in the methods to the sizeDiff function, and have it use the methods if the argument type is function, otherwise use it as the attribute of the data object.

liquidpele: Oh, well if nothing would use it then whatever, no need to add functionality that isn't used :)

ThomasBrierley:
Ok that should do it, bit paranoid about this now so went through all the examples manually again :P

This relies on one assumption: all the publicly settable getters used by getDiffs accept the same d[0], d[1] parameters regardless of whether they need them all (I noticed getShape doesn't need d[1] for example).

ThomasBrierley:
One more thing, the cache object could be considered a source of memory leaks if one wanted to add and remove lots of points continuously. So this removes entries on exit.

This can be tested roughly with the "LineWithFocusChart" example by taking and comparing heap snapshots while expanding and contracting the focused area. It's hard to get much difference in size because the cache objects are so small compared to everything else...

The important thing is the heap does not grow unbounded will adding and removing.

src/models/scatter.js
```diff
@@ -448,20 +451,22 @@ nv.models.scatter = function() {
...
+            points.exit().each(delCache).remove();
             groups.exit().selectAll('path.nv-point')
                 .watchTransition(renderWatch, 'scatter exit')
                 .attr('transform', function(d) {
                     return 'translate(' + nv.utils.NaNtoZero(x(getX(d[0],d[1]))) + ',' + nv.utils.NaNtoZero(y(getY(d[0],d[1]))) + ')'
                 })
                 .remove();
-            points.filter(function (d) { return scaleDiff || sizeDiff || getDiffs(d, 'x', 'y'); })
+            // Update points position only if "x" or "y" have changed
+            points.filter(function (d) { return scaleDiff || sizeDiff || getDiffs(d, 'x', getX, 'y', getY); })
```
ThomasBrierley: It would be nice and more atomic if I could just pass the getter functions on their own, but there needs to be a way to identify the function uniquely to store it's value in the cache for comparison next time.

The getDiffs function is generic and takes n key + getter pairs in any order, the key can actually be anything, so long as the same key is used for referring to the same value in any subsequent calls.

I did look at the options for using functions as keys: ES6 has a nice reliable function.name property which refers to the _first_ reference assigned to the function regardless of scope and var vs function name... ES6 and under is unreliable without just turning the whole function source into a string which seems a bit excessive... Sorry long explanation.

liquidpele: You could use function.toString() as the key.

liquidpele: Or if you really want them named, just pass it as an object, so like:

{x: getX, y: getY}, so that there is a clear relation instead of the next dev having to parse through the code to figure out how those arguments relate.

ThomasBrierley: > You could use function.toString() as the key.
Yes this is the ES5 option I meant but it's it's more performance sensitive not only calling yet another function to get a potentially huge string but also using huge strings as keys in a hash table.

> Or if you really want them named, just pass it as an object, so like:
> 
> {x: getX, y: getY}, so that there is a clear relation instead of the next dev having to parse through the code to figure out how those arguments relate.
I agree that would be clearer but adding (point * attributes) of throwaway objects every update is a lot of work for the GC. objects are particularly expensive to create and destroy in large numbers, where as the scope and other primitives inside the function calls are cheap. (only objects are currently the cache entries which persist across updates).

I'm on board with making the code clearer but I think it needs to not undermine the purpose of it. I'm open to suggestions if you can think of any better ways, alternatively I can comment the gettDiffs function to make it's purpose and interface more explicitly apparent.

If I had to choose one of your current options I'd go for function.toString() because that probably has less overhead and would be smoother than GC which operates intermittently.

liquidpele: Since that function is only within the internals of scatter, I think it's okay to keep it how it is if you feel the performance would suffer otherwise.   Just comment it to explain why you chose to do it that way (including the other options and your concerns with those).

liquidpele:
I'm just merging this.  Please add a comment explaining that method whenever you get a chance, otherwise I'm afraid someone else coming by will tinker with it and break it.   Or unit tests, that would be good too :)

liquidpele:
rebuilt master with this change.

ThomasBrierley:
Sorry, I've been a bit busy recently.

I will do another PR soon with comments / unit tests and the suggested improvements from #1814

Issue #1819, opened 2016-09-18
Missing parameter declarations


ehartford:
in the force directed graph model, parameters getX and getY are not declared, will this not create global variables?

https://github.com/novus/nvd3/blob/master/src/models/forceDirectedGraph.js#L163

I would expect to see something here

https://github.com/novus/nvd3/blob/master/src/models/forceDirectedGraph.js#L27

like 

    ,getX = d3.functor(0.0)
    ,getY = d3.functor(0.0)

liquidpele:
Yea, that looks like a bug.  That's a fairly new graph so it was probably an oversight by the author.   Send us a pull request with the fix if you'd like :)

PR #1828, opened 2016-09-23
add declarations for getX and getY in force directed graph

ehartford:
https://github.com/novus/nvd3/issues/1819

### DIFF
--- a/src/models/forceDirectedGraph.js
+++ b/src/models/forceDirectedGraph.js
@@ -24,6 +24,8 @@ nv.models.forceDirectedGraph = function() {
         // These functions allow to add extra attributes to ndes and links
         ,nodeExtras = function(nodes) { /* Do nothing */ }
         ,linkExtras = function(links) { /* Do nothing */ }
+        , getX=d3.functor(0.0)
+        , getY=d3.functor(0.0)
         ;
 
 


src/models/forceDirectedGraph.js
```diff
@@ -24,6 +24,8 @@ nv.models.forceDirectedGraph = function() {
         // These functions allow to add extra attributes to ndes and links
         ,nodeExtras = function(nodes) { /* Do nothing */ }
         ,linkExtras = function(links) { /* Do nothing */ }
+        , getX=d3.functor(0.0)
+        , getY=d3.functor(0.0)
```
liquidpele: getX and getY don't appear to actually be used anywhere?

ehartford: https://github.com/ehartford/nvd3/blob/dc81385b97ee48a7f8a9d8988f997867dc93b646/src/models/forceDirectedGraph.js#L165

liquidpele: ping

liquidpele:
going to merge this just so calling x() and y() won't throw errors, but it looks like really getX and getY need to be either used or the whole options need to be removed entirely.

PR #1834, opened 2016-09-30
Fix Issue 1816: Boxplot with 0 values for either high whisker or low whisker do not display

billderose:


liquidpele:
cool, thanks!

PR #1835, opened 2016-10-01
Implement autoprefixer & use consistent formatting. Fixes #1833

### LINKED ISSUE #1833: In the CSS, it's best-practice to list vendor prefixes before the real, unprefixed property
It's to safeguard in case the prefixed property was buggy or the unprefixed one changed how it works somehow.

For instance, [in lines.css L21](https://github.com/novus/nvd3/blob/master/src/css/lines.css#L21), `transition` should come last.

It would be even better if the prefixes were added with Autoprefixer on build though.

stowball:


liquidpele:
Unit tests didn't run now:

Running "postcss:dist" (postcss) task
Warning: Promise is not defined Use --force to continue.

stowball:
Hmmm the tests all ran for me. I ran all the gulp tasks successfully

liquidpele:
Hmm, weird... maybe commit something minor, like a comment, to have travis CI re-run and see if it has the same issue..

stowball:
Ah, [postcss needs node `>=0.12`](https://github.com/postcss/postcss/blob/master/package.json#L7), Travis is using `v0.10.47` 😢

Issue #1243, opened 2015-09-07
LineWithFocus: chart doesn't update if the focus area is less than 1 even if chart is all decimals


roksys:
Hello,

I am using line chart with focus. I faced one problem, when the difference between max x and min y value are less than 1, only focus bar is appearing. Example - http://i.imgur.com/iiGnuky.png.

Chart code:
```javascript
function line_chart(coordinates, x_axis_label, y_axis_label){
    var width = 900, height = 490;
    var chart = nv.models.lineWithFocusChart()
        .margin({left: 130, bottom: 50})  //Adjust chart margins to give the x-axis some breathing room.
        .duration(350)  //how fast do you want the lines to transition?
        .showLegend(false)    //Show the legend, allowing users to turn on/off line series.
        .width(width)
        .height(height);


    chart.xAxis.axisLabel(x_axis_label).tickFormat(d3.format('d'));
    chart.yAxis.tickFormat(d3.format('.12f'));
    var myData = [
        {
            color: '#2ca02c',
            values: coordinates
        }    
    ];

    d3.select('#chart svg').datum(myData).call(chart).style({'width': width, 'height': height}); 
    nv.utils.windowResize(function() { chart.update() }); 

    return chart;
}
```

roksys:
I found where is the problem 7544 line
//The brush extent cannot be less than one.  If it is, don't update the line chart. 
```javascript
if (Math.abs(extent[0] - extent[1]) <= 1) {
    return;
}
```
But I don't get, why brush extent cannot be less than one? Could anyone explain it?

liquidpele:
Hmm, that doesn't seem right.  Perhaps they wanted to prevent an update too often from small values, but assumed the chart would be using larger numbers?   Send up a pull request with a fix and we will review it :)

roksys:
I just commented those lines, and it's working

liquidpele:
We should check the range of the chart and see if the extent range is less than a small percent of that instead of assuming that the chart is all large numbers...

billderose:
@liquidpele What's curious is that the comment right before this line reads `"//The brush extent cannot be less than one.  If it is, don't update the line chart."`. So maybe we just need to change the condition to: 
```
if (!brush.empty() && Math.abs(extent[0] - extent[1]) <= 1) {
    return;
}
```
[...]

Issue #955, opened 2015-04-13
Line with focus chart small scale not rendering


dovidkopel:
With this fiddle:
http://jsfiddle.net/jr3c3v5s/1/

````
nv.addGraph(function() {
    var chart = nv.models.lineWithFocusChart();

    d3.select('#chart svg')
    .datum([
        {
            key: 'a',
            values: [
                {
                    x: 0,
                    y: 0
                },
                {
                    x: 2,
                    y: 2
                }
            ]
        }
    ])
        .call(chart);
    nv.utils.windowResize(chart.update);
    return chart;
});
````
I have no problem rendering the graph. However, with this fiddle: http://jsfiddle.net/w9go6s2s/3/

````
nv.addGraph(function() { var chart = nv.models.lineWithFocusChart();

    d3.select('#chart svg')
    .datum([
        {
            key: 'a',
            values: [
                {
                    x: 0,
                    y: 0
                },
                {
                    x: 1,
                    y: 1
                }
            ]
        }
    ])
        .call(chart);
    nv.utils.windowResize(chart.update);
    return chart;
});
````
I can get the bottom chart to render, but not the focus chart. Any help is appreciated!

liquidpele:
Well that's interesting... it seems that it doesn't render if all the x values are within 1...  even just making the second x 1.00001 makes it work.  

breaks:   x is 55 and 56
works:    x is 55 and 56.00001
breaks:  x is 55 and 55.3 and 55.7 and 55.9

SUBHAM-ROY:
Facing the same issue. Is there any solution available yet ?

PR #1837, opened 2016-10-03
Fix 1243: LineWithFocus: chart doesn't update if the focus area is less than 1 even if chart is all decimals

billderose:
@liquidpele See discussion [here](https://github.com/novus/nvd3/issues/1243) and [here](https://github.com/novus/nvd3/issues/955). This is a "quick and dirty" fix in that it simply removes the offending lines.


billderose:
Tagging @zipengwu who recently touched this code.

liquidpele:
I'm fine with this change, since the syncBrushing option someone added takes care of the performance issues involved in a much better way.   Any objections to it?

Issue #1859, opened 2016-10-31
Toggling focusEnable behaves differently between lineChart and linePlusBarChart


MikeMatusz:
In the following example using linePlusBarChart, toggling focusEnable does not clear the selected range and works the way I would expect it to:
http://nvd3-community.github.io/nvd3/examples/linePlusBarChart.html

If you modify that example to use lineChart:
https://plnkr.co/edit/3PMkJeOvCsYoO81L3Xyz?p=preview

Now, toggling focusEnable does clear the selected range, and also fails to actually hide the focus chart.  That latter part is definitely an issue, but the former is less clear.  Some people might actually be using the chart expecting that behavior, but in my case, I'd like it to behave the same as linePlusBarChart.  I would be willing to submit a PR for this, but should I add an additional option for lineChart to opt-in to not clearing the brush extent when the focus chart is hidden so the default behavior remains unchanged?

liquidpele:
It's not supposed to clear the focus range.  Also, your plunker doesn't remove the focus if you toggle it again.

MikeMatusz:
Maybe I didn't explain myself properly, but based on your response, I think this _is_ a valid issue.

It's not my plunker that isn't removing the focus chart, it's a problem with the library.  I've done quite a bit of debugging before submitting the issue here.  That's one half of this issue.

Also, it _is_ clearing the focus range, so it sounds like you're agreeing with me, and I'm not sure why this issue was closed.

Try this:
Toggle the focus chart ON
Select a subset of the full range on the focus chart.
Toggle the focus chart OFF

Expected Behavior:
The domain of the main chart should remain filtered and not reset
Focus chart is hidden

The linePlusBarChart example exhibits these behaviors, which you just agreed are the correct ones.
The lineChart fails on both.  The domain of the main chart is resetting when focusEnable is set to false.

Please reopen this issue, or provide either a working example or a better explanation of why it was closed.

liquidpele:
Well, I'm in the office now and have had my coffee, and I'm not sure what I was thinking this morning :p

MikeMatusz:
Haha, thanks!  So I have modified the code locally (I'll create a fork soon), but it's basically removing a few blocks of code from lineChart and adding the display none/initial that linePlusBarChart uses.  My concern is that someone might currently be using it and expecting the domain to reset (i.e. this would be a breaking change) so I was wondering if I should create a setting to opt-in to the behavior.

[...]

PR #1860, opened 2016-11-02
Line chart focus enabled toggle

MikeMatusz:
Fixing issues where focus chart did not hide properly when toggled off, and selected range would reset.


MikeMatusz:
Relates to #1859

liquidpele:

PR #1864, opened 2016-11-11
Halved the pie margin

zinkethaking:


liquidpele:
You need to provide some description about what you changed, and more importantly *why* you changed it.

zinkethaking:
picture1
picture2

This is a zoomed snap of one of many pies I have on the screen. I just thought the amount of margin (20%) was a bit too much so I halved it (10%)

liquidpele:
Thanks for contributing!

PR #1880, opened 2016-12-01
Fix point update bug in scatter model

### LINKED ISSUE #1865: useInteractiveGuideline points missing or mispositioned
I upgraded from 1.8.1 to 1.8.4 yesterday, to fix a bug with the interactive guideline, but now I have another one

I use the example from https://nvd3-community.github.io/nvd3/examples/site.html (LineWithFocusChart in particular, but the problem is in any graph that use useInteractiveGuideline)
When you don't have all series selected, the points of some series are missing or mispositioned.

captura de pantalla de 2016-11-11 18-29-41
Even when it's only one left:
captura de pantalla de 2016-11-11 18-30-08

When all series are selected works OK:
captura de pantalla de 2016-11-11 18-29-52

Tested in: _Mozilla Firefox 45.4 ESR_ and _Chrome 54.0_ (Ubuntu 16.04)

ThomasBrierley:
When series are toggled on and off the axis are able to dynamically resize, scaleDiff and sizeDiff do not capture this change. domainDiff added and used to determine point x,y update pruning.

Fixes #1865

This is the third bug this optimisation has caused... I'm beginning to think it wasn't such a good idea. If any more pop up I'm tempted to remove it, or at least have someone more familiar with the guts of nvd3 to check over the considerations when culling points from updates.

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -51,6 +51,7 @@ nv.models.scatter = function() {
     //------------------------------------------------------------
 
     var x0, y0, z0 // used to store previous scales
+        , xDom, yDom // used to store previous domains
         , width0
         , height0
         , timeoutID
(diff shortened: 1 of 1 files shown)


liquidpele:
neat, thanks for contributing!

Issue #1662, opened 2016-05-16
multichart same x-axis scale


kimyu92:
How could I force line chart to be rendered on the  scatter plot x-axis scale?

<img width="330" alt="screen shot 2016-05-16 at 10 37 56 am" src="https://cloud.githubusercontent.com/assets/2663509/15294556/54e1bdd4-1b52-11e6-9e9c-19977f9aa580.png">

kimyu92:
@liquidpele is there a way to allow scatter plot and line to be render on the same x-axis scale with the latest source code?

frozenshadow:
I solved it by replacing this line https://github.com/novus/nvd3/blob/master/build/nv.d3.js#L9747 with the following code:
```javascript
var ex = d3.extent(d3.merge(series1.concat(series2)), function(d) { return d.x }); // Allow extent to be used multiple times
for(var i = 0; i < charts.length; i++){ // Loop through all charts except bars1 and bars2 (because it's ordinal)
	if(!(i === 4 || i === 5)){ charts[i].xDomain(ex); } // Set xDomain for each chart individually
}

x.domain(ex).range([0, availableWidth]);
```

Please, let me know if it helps.

liquidpele:
Send a pull request if you fix a bug, thanks!

frozenshadow:
@liquidpele Sure! I just don't know if it really fixed the bug in all situations.
I only tested it for my use case (which is the same as that of @kimyu92) and in the example.

So do you still want me to make a PR?

liquidpele:
If you think it's a fix, then yes, but it's really up to you.   I rely on pull requests from the community for updates/fixes.

frozenshadow:
@liquidpele I've just made the pull request. Hope it helps :)

cmshirley:
Thank you @frozenshadow for taking the time to submit this! We are having issues with not being able to set a domain on multicharts.

PR #1881, opened 2016-12-04
MultiChart X-axis scale fix

frozenshadow:
X-axis scale fix for issue #1662

### DIFF
--- a/src/models/multiChart.js
+++ b/src/models/multiChart.js
@@ -97,8 +97,12 @@ nv.models.multiChart = function() {
                     })
                 });
 
-            x   .domain(d3.extent(d3.merge(series1.concat(series2)), function(d) { return d.x }))
-                .range([0, availableWidth]);
+            var ex = d3.extent(d3.merge(series1.concat(series2)), function(d) { return d.x }); // Allow extent to be used multiple times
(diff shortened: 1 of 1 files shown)


src/models/multiChart.js
```diff
@@ -97,8 +97,12 @@ nv.models.multiChart = function() {
                     })
                 });
 
-            x   .domain(d3.extent(d3.merge(series1.concat(series2)), function(d) { return d.x }))
-                .range([0, availableWidth]);
+            var ex = d3.extent(d3.merge(series1.concat(series2)), function(d) { return d.x }); // Allow extent to be used multiple times
+            for(var i = 0; i < charts.length; i++){ // Loop through all charts except bars1 and bars2 (because it's ordinal)
+                if(!(i === 4 || i === 5)){ charts[i].xDomain(ex); } // Set xDomain for each chart individually
```
liquidpele: Instead of using the index, this should be checking if the charts are ordinal... otherwise someone adding/changing that array of charts will break this in a non-obvious way.

frozenshadow: Thanks for the response.
Sounds like you've got a good point there. I will look into it later.

frozenshadow: @liquidpele 
I'm truly sorry for this late reply.
It has been quite a long time since I have seen and touched this code. And unfortunately I don't really understand what you mean. Do you mean the for loop?

PR #1890, opened 2016-12-10
Fix #1889: Proper dimensions for bottom legend on line chart

### LINKED ISSUE #1889: Line Chart Legend "bottom" Does Not Scale Correctly
NVD3 v1.8.5-dev

```
            var chart = nv.models.lineChart()
                .interpolate("monotone")
                .legendPosition("bottom")
```

When resizing the browser, the legend items sometimes gets cut off if there are too many keywords in the legend.  The reason is because unlike the `bottom` position of the legend in the other nvd3 models, the line chart does not adjust the chart's bottom margin based on the legend height.

KidA78:
Fix includes:
- Accounting for chart X axis on legend bottom position
- Adjusts available height according to adjusted bottom margin
- Added example of line chart w/ bottom margin

Usage:
```
        chart = nv.models.lineChart().legendPosition("bottom");
```


liquidpele:
I'd prefer not to merge in the example html... you could include this option in the existing example html page somehow, but I'd like to avoid having too many example pages as it becomes kind of difficult to maintain and browse.

KidA78:
@liquidpele I hear you.  Just made the change.  Removed the extra example I added, and put in a toggle button to toggle the legend between top and bottom. Give it a shot and let me know what you think

examples/lineChart.html
```diff
@@ -29,6 +29,8 @@
 <div style="position:absolute; top: 0; left: 0;">
     <button onclick="randomizeFillOpacity();">Randomize fill opacity</button>
     <button onclick="expandLegend();">Expand/Contract Legend</button>
+    <button onclick="toggleLegend();">Toggle Legend Top/Bottom</button>
```
liquidpele: This is going to make the buttons overlap on the legend when it's at the top, at least in the default layout where it's in an iframe with a set width... but that's an issue that's beyond the scope of this PR so it's fine for now.

KidA78:
@liquidpele Thanks!

Issue #1653, opened 2016-05-11
evt.seriesIndex is wrong in multichart


EliSnow:
Referring to [this](https://github.com/novus/nvd3/blob/254a7b3f65d83129c8600876055058ffc9fd1d4e/src/models/multiChart.js#L293) in the multichart code, the `seriesIndex` is wrong for the point being hovered over and causes the wrong `tickFormat` function to be used.

Interestingly enough the event object also has a `series` propery on it which could be used. For example:
```js
var yaxis = evt.series.yAxis === 2 ? yAxis2 : yAxis1;
```

An example of this breaking can be found [in this plunker](http://plnkr.co/edit/tWNZOFKGB13lDBGFKPhr?p=preview). I have used the non-minified version of nvd3 here so you can more easily view the source in developer tools, add break points and inspect. When I put a breakpoint on line 9614 (at least that's the correct line # in Chrome's dev tools on my system), which is the equivalent line linked to above in the multiChart code, and hover over a point which belongs to yAxis2, the `seriesIndex` points to the wrong position in the data array, thus grabbing the wrong `tickFormat` function.

PR #1911, opened 2017-01-02
Fix multiChart tooltip series yAxis (#1653)

sylvinus:
This is a pull request with the fix suggested in #1653, with some more code to make it work with multiBar. We're running this in production.


liquidpele:
Seems fine.

PR #1916, opened 2017-01-05
Fix for xScale in multiChart

krmarien:
With this change the xScale can be set in the multiChart:

    var chart = nv.models.multiChart();
    chart.lines1.xScale(d3.time.scale());

Example: [http://plnkr.co/edit/TOcSRfEEp5YnnfiJ6Afd?p=preview](http://plnkr.co/edit/TOcSRfEEp5YnnfiJ6Afd?p=preview)

### DIFF
--- a/src/models/multiChart.js
+++ b/src/models/multiChart.js
@@ -44,7 +44,7 @@ nv.models.multiChart = function() {
         stack1 = nv.models.stackedArea().yScale(yScale1).duration(duration),
         stack2 = nv.models.stackedArea().yScale(yScale2).duration(duration),
 
-        xAxis = nv.models.axis().scale(x).orient('bottom').tickPadding(5).duration(duration),
+        xAxis = nv.models.axis().orient('bottom').tickPadding(5).duration(duration),
         yAxis1 = nv.models.axis().scale(yScale1).orient('left').duration(duration),
         yAxis2 = nv.models.axis().scale(yScale2).orient('right').duration(duration),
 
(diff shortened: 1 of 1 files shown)


src/models/multiChart.js
```diff
@@ -213,6 +213,9 @@ nv.models.multiChart = function() {
                 return a.map(function(aVal,i){return {x: aVal.x, y: aVal.y + b[i].y}})
             }).concat([{x:0, y:0}]) : [];
 
+            x = lines1.xScale();
```
liquidpele: The x variable has stuff done to it above this, why would you overwrite it here?

krmarien: X was initialized to a linear scale and that scale was applied to the xAxis. However changing it afterwards didn't work. 

So the initialization was removed and x is only assigned here and used afterwards. 

It is possible I missed something, but this was the only way I got a different xAxis scale working.

liquidpele: yea, this doesn't seem right... this seems like it'll just throw away the:

    x   .domain(d3.extent(d3.merge(series1.concat(series2)), function(d) { return d.x }))
                 .range([0, availableWidth]);		                 .range([0, availableWidth]);

And even then, you're not changing the xScale of anything but lines1.   I think what you probably want to do is add a new option on multiChart named xScale and have it then control the xScale method on all subcharts (or at least the ones that have that option).

jufemaiz: "add a new option on multiChart named xScale and have it then control the xScale method on all subcharts"

YES! THIS IS EXACTLY RIGHT!

Sorry I've been tearing my hair out trying to get this work to allow for multilingual support and the D3 timeFormat.multi capability ><

Robsonsjre: This is my try on solving this issue (#2039)
Its really simple, just add a new option xScale on your chart (like the lineChart)

liquidpele:
Closing this as abandoned.

PR #1923, opened 2017-01-18
Update tooltip.js

BetterYan:
add d.index args so that chart model can get index information. 
(eg: existed line chart code snapshot:
        interactiveLayer.tooltip.valueFormatter(function (d, i) {
            return yAxis.tickFormat()(d, i);
        }).headerFormatter(function (d, i) {
            return xAxis.tickFormat()(d, i);
        });
)

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -72,7 +72,7 @@ nv.models.tooltip = function() {
                 .attr("colspan", 3)
                 .append("strong")
                 .classed("x-value", true)
-                .html(headerFormatter(d.value));
+                .html(headerFormatter(d.value,d.index));
         }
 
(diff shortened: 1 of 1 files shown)


src/tooltip.js
```diff
@@ -72,7 +72,7 @@ nv.models.tooltip = function() {
                 .attr("colspan", 3)
                 .append("strong")
                 .classed("x-value", true)
-                .html(headerFormatter(d.value));
+                .html(headerFormatter(d.value,d.index));
```
liquidpele: I think I'd rather just pass d itself here... that way we don't have to keep adding more and more to the function call every time someone needs another thing off d.

liquidpele:
closing as abandoned.

PR #1933, opened 2017-01-20
Scatter Model Maintainability

ThomasBrierley:
- **Comments** - For implementation decisions and optimisations (Requested by @liquidpele in #1817)
- **Formatting** - Conservative tidying of concerned code for legibility
- ~~**Performance** - Remove duplicate point enter/exit code (credit to @PeterVermont from #1814)~~
- **Fix** - A subtle edge case with invalid caches

The removal of the duplicate code warrants some thorough testing, it does appear to be unnecessary but it's possible it was there for a reason. The change passes all unit tests and at a glance everything seems ok. I'm gradually going through the examples in painful detail, (other testers are welcome).

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -473,14 +473,33 @@ nv.models.scatter = function() {
                     return 'translate(' + nv.utils.NaNtoZero(x(getX(d[0],d[1]))) + ',' + nv.utils.NaNtoZero(y(getY(d[0],d[1]))) + ')'
                 })
                 .remove();
-            // Update points position only if "x" or "y" have changed
+
+            //============================================================
+            // Point Update Optimisation Notes
+            //------------------------------------------------------------
+            // The following update selections are filtered with getDiffs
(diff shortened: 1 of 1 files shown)


ThomasBrierley:
Heh... the test *is* against the HEAD commit of the source branch but git has revealed my re-basing shenanigans and shown the commits in date order.

ThomasBrierley:
Don't merge this yet... 3fc7ee4 (The removal of duplicate code) causes some initialisation issues.

I think it can still be removed but the update selection has to pick up the slack in a more unified way.

ThomasBrierley:
I've spliced out 3fc7ee4 for now, the duplicate code seems to be necessary for transitions to work properly. The remaining functional changes are minor, the example pages all appear to work properly, should be safe to merge.

liquidpele:
looks like the only real change was to call getDiffs before "scaleDiff || sizeDiff || domainDiff"... seems fine.

PR #1934, opened 2017-01-20
Fixes Issue #330 - Scatter chart error when multiple points have same value

KidA78:
The following pull request takes into consideration that adding a jitter to all the points on a line chart will reduce the chance of overlapping points being sent to voronoi, however it is not 100% going to prevent this from happening.  Randomly errors described in issue #330 still occur.

The following fix keeps the jitter, but in case there are still any duplicates, dedupe them.

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -222,13 +222,12 @@ nv.models.scatter = function() {
                                 .map(function(point, pointIndex) {
                                     // *Adding noise to make duplicates very unlikely
                                     // *Injecting series and point index for reference
-                                    /* *Adding a 'jitter' to the points, because there's an issue in d3.geom.voronoi.
-                                     */
+                                    // *Adding a 'jitter' to the points, because there's an issue in d3.geom.voronoi.
                                     var pX = getX(point,pointIndex);
                                     var pY = getY(point,pointIndex);
 
-                                    return [nv.utils.NaNtoZero(x(pX)),
-                                            nv.utils.NaNtoZero(y(pY)),
+                                    return [nv.utils.NaNtoZero(x(pX)) + Math.random() * 1e-4,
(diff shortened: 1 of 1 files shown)


src/models/scatter.js
```diff
@@ -256,6 +255,18 @@ nv.models.scatter = function() {
                         [width + 10,-10]
                     ]);
 
+                    // delete duplicates from vertices - essential assumption for d3.geom.voronoi
+                    var epsilon = 1e-4; // Uses 1e-4 to determine equivalence.
+                    vertices = vertices.sort(function(a,b){return ((a[0] - b[0]) || (a[1] - b[1]))});
```
liquidpele: I'm not sure I understand why this sort here is necessary.  nvd3 already requires that the data provided has the x scale ordered.

KidA78: @liquidpele, I got inspiration from this P.R. https://github.com/novus/nvd3/pull/1934
It looks like the sort is used to make sure we dedupe properly below.

liquidpele:
Cool thanks for contributing!

AgDude:
Given the abandoned state of this repository I am not going to open a pull request, but there is a bug in this implementation. The sorting does not account for values which differ by less than epsilon on the x axis.
This can be fixed with the following sort function:
```
vertices = vertices.sort(function(a,b){
                      return Math.abs(a[0] - b[0]) >= epsilon ? 
                      a[0] - b[0] : 
                      a[1] - b[1];
                    });
```

liquidpele:
Well, it's not abandoned... just forked.  I'm still accepting pull requests at nvd3/nvd3

AgDude:
Opened https://github.com/nvd3/nvd3/pull/10
Thanks @liquidpele  for your initiative in updating this to more modern d3 versions. We will happily upgrade when that fork is stable.

PR #1937, opened 2017-01-26
Issue 1666

aszlacheta:
Many users were asking for possibility to disable only double click events on legend. For the time being there is only flag resposible for disabling all kind of click events. Expected behavior can be achieved by adding new flag `enableDoubleClick`.


liquidpele:
I'll go ahead and merge this too, but if you can also make a PR to add the description/default/example for this new option into the documentation.html json structure then that'd be great!

PR #1938, opened 2017-01-27
Multichart

memee:
Multichart was added options `barModel` and `xScale`. Now one can use `historicalBar` in multichart which plays well with lines and additional bars: grids are aligned with bars.

Besides I've added example server grunt task.


liquidpele:
This PR contains changes to historicalBar.js too.  Please keep each PR containing only a single feature or bugfix, thanks!

memee:
I'm not sure what options I have left. If one would like to have two bars and lines showing historical data he must use `historicalBar`. And It can't be done without those changes to `historicalBar` model.

The only option I see is to prepare completely new bar model inspired by `historicalBar` and used only in `multiChart`.

liquidpele:
I guess I don't understand why the changes to historicalBar.js were needed then.  Can you explain it more?

memee:
In my case I wanted to use `multiBarChart` with 2 bars, 1 line. By default it used `multiBar` model for bars. In didn't work in combination with lines - it wasn't aligned. So I provided an option to change scale to time and bar model to historicalBar which has its first and last element placed exactly in the middle of a first and the last tick.

Unfortunately historicalBar was designed to be the only one on the chart. When two historicalBars were set there were errors with tooltips. My modification hopefully doesn't break anything else and provide an option to use `historicalBar` for `multiChart`.

Only other option I see is to create `historicalMultiBar` model which may need some code duplication.

PR #1940, opened 2017-01-27
Add grunt-cli to package.json for local build

ThomasBrierley:
Make NVD3 development more build system agnostic. There are many node
build systems, not every project uses the same one, we shouldn't have to
remember which one or need to globaly install it's CLI.

npm adds ./node_modules/.bin to the run $PATH, so we can add grunt-cli
to the local dependencies. Contributors are free to use generic "npm run
build" commands without knowing or remembering grunt.


liquidpele:
k

Issue #1924, opened 2017-01-18
Sunburst chart "gaps"


m1pum:
I face a problem with the sunburst chart. It is also in your demo: http://nvd3-community.github.io/nvd3/examples/site.html

if you observe the path "flare" -> "data" -> "converter" (code line 106 when opening the source), "data" is not displayed. 
If I switch to version 1.8.2 of nvd3 it's there but things like labels and elementClick listeners are not working...

bgth:
Yeah right. If we use "data" string in name, its actually ignoring it. Please try changing it to some other string like "devicedata". It should work.

liquidpele:
merged PR that should fix this.

m1pum:
Thank you. I also found out that since 1.8.3 there is a big clash if I use the same <name> twice in different branches of the sunburst. This wasn't a problem in versions <1.8.2. But it's avoidable now that I found what the issue was :-)

m1pum:
Also, doubleClick event doesn't work as far as I can judge.

PR #1943, opened 2017-01-31
Fix for string "data" not showing in sunburst chart.

bgth:
Fix for [1924](https://github.com/novus/nvd3/issues/1924)


liquidpele:
ah, there are two "data" nodes I guess.   cool, thanks!

PR #1954, opened 2017-02-07
fix interactive guideline offset for Firefox, fixes #1953

### LINKED ISSUE #1953: Firefox InteractiveGuideline offset for css transform: translate
Using style `transform: translate` creates a problem in Firefox with the interactive guideline. (Does not happen in Chrome.)

Reproduce with test/translateTest.html or this Fiddle: https://jsfiddle.net/tcf957kq/1/

When you hover over the center point, Firefox does not highlight it. But it gets highlighted when you are in the right quarter of the graph. (Works fine in Chromium.)

(Additional issue: the tooltip appears too far away from the cursor, for both Chromium and Firefox.)

martinxyz:
See linked issue. Not sure if this is a clean solution, but it does fix the problem.

### DIFF
--- a/src/interactiveLayer.js
+++ b/src/interactiveLayer.js
@@ -40,9 +40,9 @@ nv.interactiveGuideline = function() {
             }
 
             function mouseHandler() {
-                var d3mouse = d3.mouse(this);
-                var mouseX = d3mouse[0];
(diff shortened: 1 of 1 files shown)


liquidpele:

PR #1957, opened 2017-02-10
Fix #1956

### LINKED ISSUE #1956: Memory leak causes browser tab to crash
A bug was reported in pfSense and after digging into it a developer figured out it was related to nvd3. Details are on pfSense bug:

https://redmine.pfsense.org/issues/7157

Looks like root cause is https://github.com/novus/nvd3/blob/ff012b78a3af803243dad9e975032329e63fe367/src/models/axis.js#L383

```js
nv.utils.inheritOptionsD3(chart, scale, ['domain', 'range', 'rangeBand', 'rangeBands']);
```

bgth:
[Fix #1956 ](https://github.com/novus/nvd3/issues/1956) change will keep the list of d3 options unique so that even if the same name is added again, it will append only if unique.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -510,6 +510,8 @@ d3.rebind makes calling the function on target actually call it on source
 Also use _d3options so we can track what we inherit for documentation and chained inheritance
 */
 nv.utils.inheritOptionsD3 = function(target, d3_source, oplist) {
+    // Find unique d3 options (string) and update d3options
+    target._d3options = (target._d3options || []).filter(function(item, i, ar){ return ar.indexOf(item) === i; });
     target._d3options = oplist.concat(target._d3options || []);
     oplist.unshift(d3_source);
     oplist.unshift(target);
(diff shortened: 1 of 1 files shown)


src/utils.js
```diff
@@ -510,6 +510,8 @@ d3.rebind makes calling the function on target actually call it on source
 Also use _d3options so we can track what we inherit for documentation and chained inheritance
 */
 nv.utils.inheritOptionsD3 = function(target, d3_source, oplist) {
+    // Find unique d3 options (string) and update d3options
+    target._d3options = (target._d3options || []).filter(function(item, i, ar){ return ar.indexOf(item) === i; });
```
liquidpele: shouldn't this be done after the concat below, so that it never has duplicates?

liquidpele:

src/utils.js
bgth: Sorry. I have corrected.

PR #1961, opened 2017-02-15
[pie-chart] Fix label overlap

ovvn:
Not to display overlapping labels

Before
labels-overlap

After
no-overlap

Credits
http://stackoverflow.com/a/19801529/2697673

### DIFF
--- a/src/models/pie.js
+++ b/src/models/pie.js
@@ -338,6 +338,39 @@ nv.models.pie = function() {
                         }
                         return label;
                     })
+                    .each(function (d, i) {
+                      var bb = this.getBBox(),
+                        center = labelsArc[i].centroid(d);
+                      console.log({center: center})
+                      console.log({bb: bb})
+                      var topLeft = {
+                          x : center[0] + bb.x,
+                          y : center[1] + bb.y
+                      };
+
+                      var topRight = {
+                          x : topLeft.x + bb.width,
+                          y : topLeft.y
+                      };
+
+                      var bottomLeft = {
+                          x : topLeft.x,
+                          y : topLeft.y + bb.height
+                      };
+
+                      var bottomRight = {
+                          x : topLeft.x + bb.width,
+                          y : topLeft.y + bb.height
+                      };
+
+                      d.visible = nv.utils.pointIsInArc(topLeft, d, arc) &&
+                        nv.utils.pointIsInArc(topRight, d, arc) &&
+                        nv.utils.pointIsInArc(bottomLeft, d, arc) &&
+                        nv.utils.pointIsInArc(bottomRight, d, arc);
+                    })
+                    .style('display', function (d) {
+                      return d.visible ? null : 'none';
+                    })
                 ;
             }
 
--- a/src/models/pie.js
+++ b/src/models/pie.js
@@ -341,8 +341,6 @@ nv.models.pie = function() {
                     .each(function (d, i) {
                       var bb = this.getBBox(),
                         center = labelsArc[i].centroid(d);
-                      console.log({center: center})
-                      console.log({bb: bb})
                       var topLeft = {
                           x : center[0] + bb.x,
                           y : center[1] + bb.y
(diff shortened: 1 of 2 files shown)


src/models/pie.js
```diff
@@ -338,6 +338,38 @@ nv.models.pie = function() {
                         }
                         return label;
                     })
+                    .each(function (d, i) {
+                        if (!this.getBBox) return;
+                        var bb = this.getBBox(),
```
liquidpele: iirc, getBBox throws errors if the chart is hidden or detached from the dom or something... I remember we had to replace it with getBoundingClientRect in another location a while back.

liquidpele:
Honestly, I dislike having the labels on the pie chart at all, there is almost never enough room for at least one item.    We already allow them outside the pie, but I'd like to add in lines to the labels like this at some point:

http://bl.ocks.org/dbuezas/9306799

Dunno if you're interested in that as well, but thought I'd throw it out there :)

src/models/pie.js
```diff
@@ -338,6 +338,38 @@ nv.models.pie = function() {
...
+                        };
+
+                        d.visible = nv.utils.pointIsInArc(topLeft, d, arc) &&
+                        nv.utils.pointIsInArc(topRight, d, arc) &&
+                        nv.utils.pointIsInArc(bottomLeft, d, arc) &&
+                        nv.utils.pointIsInArc(bottomRight, d, arc);
+                    })
+                    .style('display', function (d) {
+                        return d.visible ? null : 'none';
+                    })
```
liquidpele: I'm not sure I really like this... in most cases labels won't fit unless you have a giant section of the pie there, and we already have the labelThreshold option that lets you control when to stop showing labels as slices shrink...

ovvn: In our production apps in most cases we have short labels. The `labelThreshold` option doesn't really help, if I set `labelThreshold` to some size, no label will be shown within a slice of the size regardless of whether we have a short or long label. My approach is to show a label when it fits, and not to show when it doesn't fit.

ovvn:
Thanks for the information. Agree, the example http://bl.ocks.org/dbuezas/9306799 looks nice, but it doesn't really solve a problem with very long labels.

liquidpele:
I don't think anything really solves the problem with very long labels beyond overriding the format method to cut them off or abbreviate them.  Hiding them entirely just seems like one (non-ideal imho) option.

src/models/pie.js
liquidpele: I guess I'd be open to merging it if it was controlled by an option and off by default (for backward compatibility).   I can just see other people not wanting this and considering it a bug.

ovvn:
Another option is to wrap them like here https://bl.ocks.org/mbostock/7555321

src/models/pie.js
ovvn: Added

bgth:
hi @ovvn.  For very long labels, we will have to begin with definition of how many characters beyond which it will be called very long. Second, for pie chart, if the information is compulsorily required which is really long, you can show a smaller label and then you can update the tooltip with all the other info on mouse over. What is your thought? As @liquidpele has mentioned, if we can add the poly lines for labels, most of the issues of overlapping labels should be solved.

ovvn:
hi @bgth. the definition of a "long label" depends on whether we have labels within slices or outside of the slices. I don't think it's really about how many characters does a label have. It's rather a question of how it fits within a slice or within a canvas (for labels outside of the slices) or whether it overlaps neighbor labels.

zheoreh:
+1

Tarlord:
@liquidpele i agree with @ovvn, at least hiding labels is needed in pie chart(it's actually needed for all charts where they are exist), removing the labels completely is not good because sometimes people like to use shortcuts for chart-pieces and see it without any tooltips etc...

liquidpele:
Okay, with the addition of the option to turn it on I'm cool with this.   

Thanks!

Margaritarita:
+1

PR #1967, opened 2017-02-16
Examples update to d3js 3.5.17

bgth:
The examples are updated to use d3js 3.5.17. At present, they use d3js 3.5.2. The changes are tested on Mozilla firefox and Chrome. Problems faced are given below:

1) historicalBar.html - Error in tooltip - Renders successfully.
2) line.html - coded in old way -Renders successfully.
3) stackedArea.html error - TypeError: this[n] is undefined. Uncaught TypeError: Cannot read property 'on' of undefined -   d3.min.js:3:10105 - Renders successfully.
4) tooltip error - mousemove undefined 0 tooltip.html:78:13 TypeError: position is not a function - Error comes on the tooltip for circle. Renders successfully.
5) documentation.html - When edited, it was introducing Ctrl-M chars. So, have changed the whole file into unix line endings. Only one line changed in that - d3js 3.5.17.

Also i have cherry picked only one commit 69eb879ba942de8f032fdba9c246d987f639588d which has 45 file changes.


liquidpele:
Cool, looks okay to me.

PR #1969, opened 2017-02-17
Fix broken error bars in horizontal bar charts.

mackler:


liquidpele:
Can you explain what this fixes?  Like, with a screenshot?  It's not obvious just from the code, sorry.

mackler:
The way the error-bars position is calculated--as an offset from the size of the main data bar--is causing all parts of the error-bar with an offset less than zero to be truncated at zero.  You can see in the first screenshot that before the patch the error bars extend only to the right of the rightmost-edges of the main data bars.

error-bars-1

After the patch the negatively-offset portion of the error bars is visible to the left of the right-most edge of the main data bars

after-patch

Thanks.

liquidpele:
Ah, awesome, thanks for the screenshots that made it very clear. 

Thanks for contributing!

Issue #1980, opened 2017-03-03
Add babel transpilation step so can use ES2015+


jeznag:
Currently all code is written in either ES5 or coffeescript. It would be great to be able to write models in ES2015+

PR #1981, opened 2017-03-03
feat($js): Add babel transpilation step

jeznag:
https://github.com/novus/nvd3/issues/1980


liquidpele:
I'd like to keep this project as vanilla javascript.

jeznag:
@liquidpele would you consider having an an alpha version of NVD3 that would include a modern JS toolchain? I don't know that it's accurate to say that babel is not Vanilla JS. Babel is just used to maintain support for older browsers (<= IE 11).

liquidpele:
No.  Were this a much bigger project I'd consider it, but the size really doesn't warrant it, and most of the syntax is d3 specific anyway.  I don't want to tac on another required tech.

PR #1990, opened 2017-03-13
feat($axis): Add ability to set different tick format for max/min

jeznag:


liquidpele:
Seems all good, thanks for contributing!

Issue #1997, opened 2017-03-22
Add snapshot tests to improve coverage


jeznag:
I've been thinking about ways to achieve better test coverage. [Jest style snapshots](https://github.com/bahmutov/snap-shot) are a super easy way to get lots of coverage without a lot of test code.

I've done a proof of concept and it seems to work well: https://github.com/COzero/nvd3/blob/d541f2d471e68463c177f1f3f59d7431d0b4d762/test/mocha/lineWithFocusPlusSecondaryChart.es6#L881

It fails on major changes to chart structure. To make it work, I had to do a few things:
1. Set up benv to run the tests in node instead of browsers (dramatically speeds tests up and allows access to filesystem and therefore snapshot tests).

2. Strip out attributes that change every time a chart is rendered even with the same data: id, nv-edge-clip, clip-path, class (contains the unique ID), transform (this seems to change slightly every time and there's no good way of comparing slight differences), d (area changes slightly every time).
Here's reference code: https://github.com/COzero/nvd3/blob/d541f2d471e68463c177f1f3f59d7431d0b4d762/test/mocha/lineWithFocusPlusSecondaryChart.es6#L830

3. Use snapshot test:
```
builder.updateData(stableData);
const svgData = pretty(builder.svg.innerHTML);
snapshot(svgData);
```

PR #1998, opened 2017-03-24
test($tests): Improve testing infrastructure

jeznag:
Move karma config to config file, add stryker JS, set up benv for node based testing

#1997

### DIFF
--- a/GruntFile.js
+++ b/GruntFile.js
@@ -115,40 +115,16 @@ module.exports = function(grunt) {
                 }
             }
         },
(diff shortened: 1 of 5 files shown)


liquidpele:
stryker seems like something that is more part of your own build/testing infrastructure, so I don't think this is really appropriate for the core project.

liquidpele:
I'm not super familiar with it though, so if you think it benefits the project overall please explain how.

Issue #1988, opened 2017-03-13
Add new models: differenceChart and linePlusSecondary


jeznag:
We're working on two new models:
Line Plus Secondary chart:
![image](https://cloud.githubusercontent.com/assets/1034890/23838327/a2791a96-07e8-11e7-93c1-504b09a00fa1.png)

Difference chart:
![image](https://cloud.githubusercontent.com/assets/1034890/23838357/f7d14996-07e8-11e7-8148-59c33431ea9c.png)

We're nearly done with the code and are wondering if we should submit a PR or whether this is something that would be better in its own repo given the current focus of bug fixes. Let us know if you'd like a PR and we'll send a few through.

liquidpele:
The first one just looks like 2 charts, so I'm not sure that's a good fit for core nvd3... every full model we merge in increases the code base and it becomes difficult to maintain, so that one might be best for you to just keep as part of your own fork.   

As for the difference chart, that's interesting... I'd be willing to check that out as a PR, although I'm curious about whether the bars or a line like the example below is best... 
https://bl.ocks.org/mbostock/3894205

jeznag:
Fair enough. The first one is basically a spinoff of linePlusBar. The only special thing about it is we've set it up so you can switch between lines and bars for the secondary chart. We'll just keep it in our repo given the maintenance issue.

Cool we'll submit a PR for the difference chart by the end of the week. The reason it looks like a bar is because we've chosen step-before interpolation. We'll make sure the end user can override this.

PR #1999, opened 2017-03-24
Line with focus plus secondary

jeznag:


liquidpele:
As stated in https://github.com/novus/nvd3/issues/1988, this new model doesn't really fit into the upstream nvd3 code imho... ?

jeznag:
Yeah that's fine. Would be good to implement some of the elements here though, e.g. ESLint, node based testing, Stryker JS

We'll have this in the difference chart which is nearly ready for a PR.

PR #2002, opened 2017-03-24
modify indexify to divide by v and not by v+1

bernhard-42:
The cumulativeChart is actually an index chart, see [http://mbostock.github.io/protovis/ex/index-chart.html](http://mbostock.github.io/protovis/ex/index-chart.html) (I think, the name should be changed some day).

However, the first thing should be to correct the math: the `y` axis shows the percentage of `y_i` relative to the index value `v`. So the right formula is `(y_i - v)/v` (as also used in Mike Bostocks code) and not `(y_i - v)/(1 + v)`

### DIFF
--- a/src/models/cumulativeLineChart.js
+++ b/src/models/cumulativeLineChart.js
@@ -563,10 +563,8 @@ nv.models.cumulativeLineChart = function() {
             }
             var v = indexifyYGetter(indexValue, idx);
 
-            //TODO: implement check below, and disable series if series loses 100% or more cause divide by 0 issue
-            if (v < -.95 && !noErrorCheck) {
-                //if a series loses more than 100%, calculations fail.. anything close can cause major distortion (but is mathematically correct till it hits 100)
-
+            // avoid divide by zero
+            if (Math.abs(v) < 0.00001 && !noErrorCheck) {
                 line.tempDisabled = true;
                 return line;
             }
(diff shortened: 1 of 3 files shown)


bernhard-42:
Travis failed with
     The command "npm install -g grunt-cli" failed and exited with 1 during .
which is not related to the changes ...

liquidpele:
Looks correct to me...  thanks for contributing!

bernhard-42:
I just saw that there is another piece of code above that has the same strange logic.
Sorry, missed that before

cumulativeLineChart.js line 175

```
if (initialDomain[0] < -.95) initialDomain[0] = -.95;

return [
        (initialDomain[0] - initialDomain[1]) / (1 + initialDomain[1]),
        (initialDomain[1] - initialDomain[0]) / (1 + initialDomain[0])
];
```

I don’t know what happens there. Why is the value changed instead of making the line disabled?
And why only initialDomain[0]? And how can this happen after the filter?

If the above logic is OK (and this currently beyond my understanding), then it needs to changed to something like:

```
if (Math.abs(initialDomain[0]) < 0.00001) { 
    initialDomain[0] = (initialDomain[0] < 0) ? -0.00001 : 0.00001; 
}
return [
        (initialDomain[0] - initialDomain[1]) / initialDomain[1],
        (initialDomain[1] - initialDomain[0]) / initialDomain[0]
];
```

However I'd prefer to understand that part of the code better before issuing a PR.

Do you know what happens in (!rescaleY) case?

liquidpele:
tbh I've barely messed with cumulative line chart because I never had the need to use it... so your guess is as good as mine :)

bernhard-42:
I'll have a look at it and try to understand it. Will then come up with a PR that makes sense for review

bernhard-42:
I analyzed the behavior and created a pull request https://github.com/novus/nvd3/pull/2014
Hope this works now.

PR #1554, opened 2016-03-08
allows the use of stacked bars in multiChart

montrezorro:

PR #2006, opened 2017-03-28
Restore pre #1554 behaviour of bars that go down to 0 by default

sylvinus:
This commit introduced a subtle regression:
https://github.com/novus/nvd3/commit/9b75644ebc5c15caa2bb803e8d5612e1463caf5b#diff-b5254de5addcaf88c0900e46af95e3d7

Before it, `{x:0, y:0}` would get added to the `extraValue` only if `dataBars1.length`. The change made non-stacked bar charts behave differently.


liquidpele:
cool, thanks!

Issue #2005, opened 2017-03-28
Leftover reference to "tooltipShow" ?


sylvinus:
Hello,

In lineChart, it seems there is still a reference to `tooltipShow` and `tooltipHide` which I think are deprecated. Should they be removed?

thanks

liquidpele:
Yes, those can be removed now, it's been at least one release since they were deprecated.

PR #2010, opened 2017-04-03
removed deprecated references in linechart

torbenbru:
removed tooltipShow and tooltipHide from linechart addressing issue  #2005


liquidpele:
This changes the build file and not the source files.   PRs must only change source files.

PR #2011, opened 2017-04-04
Code Changes for Furious Check Box Issues In IE - furiousLegend.js & legend.js

gowrishankaran:
Hi Guys,

I fixed the Furious Legend Issues in IE Browser. Not tested in Safari. I hope it will work in safari also.

`.property('innerHTML')` - **InnerHTML** not working in IE. That's why path '**nv-box**' and '**nv-check**' elements are not rendering inside g '**nv-check-box**' element.

So, I updated the code like below.

**Current Code:**

```
seriesEnter.append('g')
        .attr('class', 'nv-check-box')
        .property('innerHTML','<path d="M0.5,5 L22.5,5 L22.5,26.5 L0.5,26.5 L0.5,5 Z" class="nv-box"></path><path d="M5.5,12.8618467 L11.9185089,19.2803556 L31,0.198864511" class="nv-check"></path>')
        .attr('transform', 'translate(-10,-8)scale(0.5)');
```
**Update Code:**

```
var seriesFuriousBox = seriesEnter.append('g')
   .attr('class', 'nv-check-box')
   .attr('transform', 'translate(-10,-8)scale(0.5)');

   // For Box
   seriesFuriousBox.append('path')
   .attr('d','M0.5,5 L22.5,5 L22.5,26.5 L0.5,26.5 L0.5,5 Z')
   .attr('class','nv-box');
   
   // For Check
   seriesFuriousBox.append('path')
   .attr('d','M5.5,12.8618467 L11.9185089,19.2803556 L31,0.198864511')
   .attr('class','nv-check'); 
```

Thanks,
Gowri Shankar

### DIFF
--- a/src/models/furiousLegend.js
+++ b/src/models/furiousLegend.js
@@ -63,10 +63,24 @@ nv.models.furiousLegend = function() {
 
                 seriesShape = series.select('rect');
 
-                seriesEnter.append('g')
+//                 seriesEnter.append('g')
+//                     .attr('class', 'nv-check-box')
+//                     .property('innerHTML','<path d="M0.5,5 L22.5,5 L22.5,26.5 L0.5,26.5 L0.5,5 Z" class="nv-box"></path><path d="M5.5,12.8618467 L11.9185089,19.2803556 L31,0.198864511" class="nv-check"></path>')
+//                     .attr('transform', 'translate(-10,-8)scale(0.5)');
+                
+                var seriesFuriousBox = seriesEnter.append('g')
                     .attr('class', 'nv-check-box')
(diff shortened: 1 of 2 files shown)


src/models/furiousLegend.js
```diff
@@ -63,10 +63,24 @@ nv.models.furiousLegend = function() {
 
                 seriesShape = series.select('rect');
 
-                seriesEnter.append('g')
+//                 seriesEnter.append('g')
```
jeznag: @gowrishankaran why not just delete the old code? We have git for version control if we ever need to get it back

PR #2012, opened 2017-04-06
fixed check for log scale

bernhard-42:
In https://github.com/d3/d3-scale the only scale that has a `base() `method is the log scale.
So checking whether `chart.yScale()base` is a function or not should differentiate between log scale case and any other case.

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -131,7 +131,7 @@ nv.models.scatter = function() {
             });
 
             // Setup Scales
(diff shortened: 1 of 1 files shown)


liquidpele:
This was merged as part of https://github.com/novus/nvd3/pull/2002

Please try to make sure to keep the PRs from different branches so that they dont' mix...  these were small enough that I could easily view it though so I went ahead and merged the combined one.

PR #2014, opened 2017-04-08
fixed rescaleY behaviour

bernhard-42:
The fixed behavior is like this:
- There is a new private variable `currentDomain` which holds the yDomain in the non-rescale case
- This variable will be initially set after the first call to `indexify`
- When the rescale control is disabled by the user (i.e. no rescale) then the yDomain of the currently selected index point is retrieved (by a new function `getCurrentDomain`) and stored in currentDomain.
- `currentDomain` is used in non-rescale case to set `lines.yDomain`
- Additionally tooltip don't show values of `tempDisabled` series and `rescaleY` option also sets `state.rescaleY`

### DIFF
--- a/src/models/cumulativeLineChart.js
+++ b/src/models/cumulativeLineChart.js
@@ -61,6 +61,7 @@ nv.models.cumulativeLineChart = function() {
     var dx = d3.scale.linear()
         , index = {i: 0, x: 0}
         , renderWatch = nv.utils.renderWatch(dispatch, duration)
+        , currentYDomain
         ;
 
     var stateGetter = function(data) {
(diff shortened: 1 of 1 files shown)


src/models/cumulativeLineChart.js
```diff
@@ -165,37 +166,23 @@ nv.models.cumulativeLineChart = function() {
...
-            }
 
             dx.domain([0, data[0].values.length - 1]) //Assumes all series have same length
                 .range([0, availableWidth])
                 .clamp(true);
 
             var data = indexify(index.i, data);
 
+            // initialize the starting yDomain for the not-rescale case after indexify (to have calculated point.display)
+            if (typeof(currentYDomain) === "undefined") currentYDomain = getCurrentYDomain(data);
```
liquidpele: Please use brackets even if the if statement is only one line.

src/models/cumulativeLineChart.js
```diff
@@ -428,8 +415,8 @@ nv.models.cumulativeLineChart = function() {
             controls.dispatch.on('legendClick', function(d,i) {
                 d.disabled = !d.disabled;
                 rescaleY = !d.disabled;
-
                 state.rescaleY = rescaleY;
+                if (!rescaleY) currentYDomain = getCurrentYDomain(data); // rescale is turned off, so set the currentYDomain
```
liquidpele: brackets

src/models/cumulativeLineChart.js
```diff
@@ -463,7 +450,7 @@ nv.models.cumulativeLineChart = function() {
                             color: color(series,series.seriesIndex)
                         });
                     });
-
+                console.log(JSON.stringify(allData))
```
liquidpele: remove debug code.

src/models/cumulativeLineChart.js
```diff
@@ -165,37 +166,23 @@ nv.models.cumulativeLineChart = function() {
...
                 .clamp(true);
 
             var data = indexify(index.i, data);
 
+            // initialize the starting yDomain for the not-rescale case after indexify (to have calculated point.display)
+            if (typeof(currentYDomain) === "undefined") currentYDomain = getCurrentYDomain(data);
+
+            if (!rescaleY) {
+                lines.yDomain(currentYDomain);
+                lines.clipEdge(true);
```
liquidpele: Why was the clipEdge added?  If needed shouldn't this either turn off for if rescaleY is true?

bernhard-42: This is the non-rescale case ( `!rescaleY`): If you do not rescale, then the yDomain gets locked to the one of the index when `rescaleY` was disabled and hence the lines could get out of the plot area when you select another index (learned they by using the cumulativeChart example).
We could add `lines.clipEdge(false)` to the rescaleY case. However due to `lines.yDomain(null)` this shouldn't be necessary, isn't it?

src/models/cumulativeLineChart.js
bernhard-42: Funny, I removed all brackets to be in sync with the rest of the code ;-)
Will add them ...

liquidpele: heh, yea I've been slowly trying to get it cleaned up so it'll pass jshint/jslint/whatever eventually.

src/models/cumulativeLineChart.js
liquidpele: Okay, as long as it doesn't mess up it's fine, was mostly just bringing it up for discussion.

bernhard-42: You could try examples/cumulativeLineChart.html, uncheck rescale and then select a few index points further to the right side. The lines will start to get out of plotting bounds and will be nicely clipped

bernhard-42:
requested changes should be incorporated now

liquidpele:
cool, thanks!

PR #2017, opened 2017-04-18
feat($models): Add difference chart model

jeznag:
issue #1988


package.json
```diff
@@ -6,7 +6,9 @@
   "main": "build/nv.d3.js",
   "scripts": {
     "test": "grunt",
-    "build": "grunt"
+    "test-unit": "mocha test/mocha/*.js",
+    "build": "grunt",
+    "stryker": "stryker run stryker.conf.js"
```
liquidpele: Please remove build related things from this PR and keep it just the new model and examples.

src/models/multiChart.js
```diff
@@ -398,13 +395,60 @@ nv.models.multiChart = function() {
...
             }
 
-            function highlightPoint(serieIndex, pointIndex, b){
-              for(var i=0, il=charts.length; i < il; i++){
-                var chart = charts[i];
-                try {
-                  chart.highlightPoint(serieIndex, pointIndex, b);
-                } catch(e){}
-              }
+            function highlightPoint(series, pointIndex, b, pointYValue) {
```
jeznag: This change corrects a bug with highlighting points where the wrong series got highlighted.

liquidpele:
unit tests appear to have an issue with the differenceChart.js

Running "karma:unit" (karma) task
Firefox 38.0.0 (Linux 0.0.0) ERROR
  SyntaxError: let is a reserved identifier
  at /home/travis/build/novus/nvd3/src/models/differenceChart.js:9

src/models/multiChart.js
liquidpele: I think this may have been fixed by a different PR I just merged.

jeznag: @liquidpele are you sure? The only recent change I can see related to highlighting points is https://github.com/novus/nvd3/commit/89567800f575f4df75c0bdc3a3127d1bbc1ef2b6 in scatter chart. That fix does not fix the bug here. The bug is specifically related to multi chart.

jeznag: The problem was that the existing code would highlight multiple charts.

e.g. Chart has line plus bar. User has highlighted first point in line chart: pointIndex is 0, seriesIndex is 1.

The current code would highlight point 0 in lines AND in bars. 

It shouldn't do that. It should only highlight a point in the chart it belongs to.

liquidpele:
This PR has build files in it, which are also causing a conflict.   Please only include src files.

jeznag:
@liquidpele sorry fixed.

GruntFile.js
```diff
@@ -130,10 +133,12 @@ module.exports = function(grunt) {
                     },
                     files: [
                         'bower_components/d3/d3.js',
+                        'node_modules/moment/moment.js',
                         'src/*.js',
                         'src/models/*.js',
                         'test/mocha/*.coffee',
-	                    'https://cdn.rawgit.com/Kcnarf/d3-beeswarm/fbda9b54/build/d3-beeswarm.min.js',
+	                      'https://cdn.rawgit.com/Kcnarf/d3-beeswarm/fbda9b54/build/d3-beeswarm.min.js',
```
liquidpele: You mixed tabs and spaces, AAAAA!!  ;)

examples/differenceChart.html
```diff
@@ -0,0 +1,145 @@
...
+          var value = (c.value || 0).toFixed(2);
+          if (c.key === keyForActualGreaterThanPredicted || c.key === keyForActualLessThanPredicted) {
+            var diff = Math.abs(c.data.y0 - c.data.y1);
+            if (diff === 0) {
+              value = '-';
+            } else {
+              value = diff.toFixed(2);
+            }
+          }
+          bodyhtml = bodyhtml + "<tr><td class='legend-color-guide'><div style='background-color: " + c.color + ";'></div></td><td class='key'>" + c.key + "</td><td class='value'>" + value + "</td></tr>";
```
liquidpele: This looks like it allows for XSS, you need to html escape the values you put within that html.  You can probably just copy the logic from lodash's escape method since we can't depend on jquery or something like that.  

https://github.com/lodash/lodash/blob/9d11b48ce5758df247607dc837a98cbfe449784a/escape.js

examples/differenceChart.html
```diff
@@ -0,0 +1,145 @@
...
+        };
+      }),
+      color: 'blue'
+    }];
+  }
+
+  async function main() {
+    const colors = d3.scale.category20();
+
+    await loadData();
```
liquidpele: I've tried to keep IE11 support still, which doesn't support async/await.  I know that's annoying, but it's still supported by a lot of products including the one I work on for my real job.

liquidpele: I guess since that's just the example it's okay though.

examples/differenceChart.html
liquidpele: I know this is the example, but people will just copy the examples to start their own work

src/models/multiChart.js
liquidpele: cool

examples/differenceChart.html
jeznag: I've converted it to ES5 :)

PR #2047, opened 2017-06-09
Fixed bug where tooltip would be inserted outside Body

theDiverDK:
The old code:
 tooltip = d3.select(document.body).select('#'+id).data(data);

Added the tooltip div outside the Body tag, which then led to translation tools not finding the tool tip text.
This is also not valid HTML

By using:
 tooltip = d3.select(document.body).selectAll('#'+id).data(data);

The tooltip div is once again inserted inside the body tag, where it should be.

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -265,7 +265,7 @@ nv.models.tooltip = function() {
             // Create new tooltip div if it doesn't exist on DOM.
 
             var data = [1];
-            tooltip = d3.select(document.body).select('#'+id).data(data);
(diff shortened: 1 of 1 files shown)


liquidpele:
I don't think this is correct... select and selectAll should both return a grouping, I think there is someting else going on.  

https://bost.ocks.org/mike/selection/#group

theDiverDK:
So you mean the current solution and my fix is both broken?

At least my fix, makes it usable.

liquidpele:
Okay, I tested this and you're right.  That's bizarre.

Issue #2035, opened 2017-05-23
chart.title is not a function


ConstantinoSchillebeeckx:
Is `.title()` not a function anymore? 

The [documentation](https://nvd3-community.github.io/nvd3/examples/documentation.html) seems to specify it:
>var chart = nv.models.pieChart();
chart.width(500);
chart.title('stuff').titleOffset(-10);
chart.options({height: 500, donut: true});

Yet when I call the following, I get a ".title is not a function" error:
```javascript
var chart = nv.models.scatterChart()
  .useVoronoi(true)
  .color(d3.scale.category10().range())
  .duration(300)
  .title('moo');
```

liquidpele:
Well, pieChart doesn't use a title anyway, only donutChart does because it goes in the middle of the donut...   that's an interesting but though that it inherits it from pie.js, that should probably only be in donutChart.js instead of being inherited like that...

ConstantinoSchillebeeckx:
I'm rendering multiple scatter plots on a single page - looks like I'm going to have to manually insert a title?

liquidpele:
Also your posted example appears to be using scatterChart anyway.

liquidpele:
Each chart type has different options... title isn't an option for scatterChart.   If you want to make it an option though, send us a pull request :)

PR #2051, opened 2017-06-13
added `title()` option for all chart types

ConstantinoSchillebeeckx:
Inspired by #2035, I've added a `title()` and `titleOffset()` option for all of the chart types.

- `title` simply takes a string and adds it to the top of the chart with a font size of 150%
- `titleOffset` takes an object of the form `{top: 0, left: 0}` and allows the user to adjust the positioning of the title relative to the top/center of the chart.

Default values are
```javascript
title: false,
titleOffset: {top: 0, left: 0}
```


ConstantinoSchillebeeckx:
Now that I understand the library better, I don't think this is the way to go. Instead, it should be integrated as a separate chart component á la `axis()`.

May work on this when I have time...

Issue #2024, opened 2017-04-27
Zooming breaks the interactive layer points highlighting


mvidailhet:
When you add a d3 zoom behavior to the chart, the interactiveguidelines points that are highlighted on mousemove are not the right ones.

After some research in the code, I found that the values array is cut if the points are not in the current domain. So the index of the point to highlight is not the right one.

To be more specific, the problem comes from this code, which filters the values:
https://github.com/novus/nvd3/blob/master/src/models/lineChart.js#L291

I don't really know how to change this piece of code without breaking the case described in the comment, so I won't do a PR, but when you remove this filter, everything works fine.

bgth:
The fix has been merged. Can this be closed ?

PR #2054, opened 2017-06-15
Fixed point highlighting on line chart after zoom or pan

tvbarnette999:
This is a fix for #2024.  When zooming or panning to the right, if the leftmost visible point is not the first index, the wrong points are highlighted when the interactiveGuideline is enabled. As mentioned in that issue, this fix should not affect the case described in the comment, as it checks the indexes after that case is handled. This changes the passed index from the point's index of the filtered array to its index in the full series.


liquidpele:
Thanks!

PR #2055, opened 2017-06-15
Fixed scatter point highlighting persistence on zoom

tvbarnette999:
This is a bugfix for zooming when the guideline is off. When hovering over a point, zooming in such that the mouse leaves the selection area of the point would not remove the point's highlighting. This resulted in the point being permanently highlighted:
 
duplicated_circles

This was fixed by clearing any highlighted points on mouse move, before it chooses a new one to highlight.


liquidpele:
Interesting, thanks for fixing!

Issue #416, opened 2014-01-18
Stream graph y-axis tilts due to nulls when accessed from multi-select


scaledev:
When the streamgraph alone is used, the y axis stays zeroed.  But when its used along with stacked area as an option, the y axis is tilted.

Here is a relevant SO q&a:

http://stackoverflow.com/questions/21172485/why-does-the-nvd3-streamgraph-shift-the-y-zero-axis-over-time?noredirect=1#comment31918486_21172485

If this isn't an issue, but there is a way to turn off the y-axis tilt from the selector, that would be great to know.

Thanks very very very much!!!

liquidpele:
It looks like stackedArea.js has "stream-center" and "stack_percent" as options, but stackedAreaChart's controls don't have those options, it only has "stack", "stream" and "expand" as the controls.  We should look into enabling the other options, or at least making the controls configurable.

jufemaiz:
@liquidpele do you have an update? :(

jeznag:
https://github.com/novus/nvd3/pull/2061

PR #2061, opened 2017-07-03
feat($stacked-area): Support Stream_Center as control option

jeznag:
Stream center option works better for datasets including zero

#416


liquidpele:

PR #2070, opened 2017-07-11
new plot type: Distrochart

ConstantinoSchillebeeckx:
This plot type aims at representing distributions ordinated by a categorical variables in formats such as box, violin or just the individual observations.

screen shot 2017-07-11 at 17 27 11

screen shot 2017-07-11 at 17 27 30

The following options exist:

- **x**: data accessor that defines the x-axis
- **value**: data accessor that defines the y-axis values
- **colorGroup**: data accessor that defines groups within each x-axis category [optional]
- **maxBoxWidth**: limits the maximum box width [optional, default null]
- **plotType**: specify the plot type, choices are 'box', 'violin' or false [default false]
- **observationType**: this option sets the observation x positioning; choices are 'random', 'swarm', 'line', 'centered' or false. a value of false hides all observations. note that [this](https://github.com/Kcnarf/d3-beeswarm) library must be loaded for 'swarm' to work.
- **whiskerDef**: method for calculating whisker length; choices are 'iqr', 'minmax', or 'stddev' [default 'iqr']
- **notchBox**: a value of true will [notch](https://sites.google.com/site/davidsstatistics/_/rsrc/1472777929924/home/notched-box-plots/Notched%20Box%20Plot%202.jpg?height=235&width=400) the boxes [default false]
- **hideWhiskers**: a value of true will hide the whiskers [default false]
- **showMiddle**: the type of 'middle' line to show in the distribution, choices are 'mean', 'median' or false [default false]
- **bandwidth**: heuristic for kde bandwidth calculation, can be float or str, if str, must be one of 'scott' or 'silverman' [default 'scott']
- **resolution**: resolution to use in kde calculation
- **showOnlyOutliers**: boolean whether to show only outliers in box plot - if false, all observations will be shown in box plot [default true]
- **jitter**: float value representing the percentage of box width that should be used for the random orientation of observations [default 0.7]
- **observationRadius**: integer value for radius of observations [default 3]
- **squash**: boolean for whether to squash boxes/violins together in the case where **colorGroup** is specified and there are missing colorGroups. [default true] For example, the first plot (below) has **squash**=false, where as the one below has **squash**=true

screen shot 2017-07-11 at 17 40 54

screen shot 2017-07-11 at 17 41 00


GruntFile.js
```diff
@@ -132,7 +132,8 @@ module.exports = function(grunt) {
                         'bower_components/d3/d3.js',
                         'src/*.js',
                         'src/models/*.js',
-                        'test/mocha/*.coffee'
+                        'test/mocha/*.coffee',
+	                    'https://cdn.rawgit.com/Kcnarf/d3-beeswarm/fbda9b54/build/d3-beeswarm.min.js',
```
liquidpele: What is this for?  I'm a little hesitant to depend on external items...

liquidpele:
Could you add an example page under the examples directory and link it in site.html?   Also, any options on the chart that are not used by any other chart should be added to documentation.html in the big json object at the bottom of the file.

GruntFile.js
ConstantinoSchillebeeckx: It's a small library that positions observations in a "swarm" pattern:
screen shot 2017-08-24 at 09 20 27

I hear ya on the external dependency, would it be better to simply include the source code in this chart type?

liquidpele: It's just for the tests right?  If so then it's probably fine, we also pull d3 externally, I was mostly curious what it was for.  Probably a good idea to add a comment there in the grunt file about what it is though just to keep track of it.

ConstantinoSchillebeeckx: Actually no, it's not just for the unit test - it is required for the plot option `observationType` which allows for a "swarm" layout of the observations.

In the [example](https://github.com/novus/nvd3/pull/2070/files#diff-eccc39e392f46ea2e0b3891db31c9b06R7), I manually pull it it - is there a way to automatically include it in the build? 

That question might be obvious, but I'm new to building libraries like this. 😢

liquidpele: Ah... in that case, lets make the option throw an error (nv.log.error) about the option requiring that library and a link to the github project.  Basically, I'd like to not require it outright, but if it's there then the option will work.  We do the same thing with fastdom.

ConstantinoSchillebeeckx: Sounds good; I went with a noData message instead of `nv.log.error`

src/utils.js
```diff
@@ -652,7 +652,7 @@ nv.utils.noData = function(chart, container) {
     noDataText
         .attr('x', x)
         .attr('y', y)
-        .text(function(t){ return t; });
+        .html(function(t){ return t; });
```
liquidpele: I'd rather not do .html, it's too easy to cause cross site scripting by accident.

liquidpele:
Looks good to me, thanks for contributing!

ConstantinoSchillebeeckx:
Actually, please hold off on merging - I'm still working on a bug that I can't seem to fix

liquidpele:
Oh, I already merged.  It's just in your new model I presume though, so it won't hurt anything until you fix it.  Just shoot me another PR whenever :)

PR #2074, opened 2017-07-14
Fix `atleastOnePoint not found` error.

yuhc:
The built js file is not up-to-date. The variable `atleastOnePoint` is not defined but used. The fix merges the latest [stackedAreaChart.js](https://github.com/novus/nvd3/blob/master/src/models/stackedAreaChart.js#L405) into [nv.d3.js](https://github.com/novus/nvd3/blob/master/build/nv.d3.js#L14799).

Stacked area chart then works fine with this fix.


liquidpele:
You cannot change the build files, those are created from the src files...  you'll have to modify the src files.

PR #2098, opened 2017-09-10
Distrochart

ConstantinoSchillebeeckx:
This is an update for PR #2070 that includes:

- object constancy fix during transitions
- disable legend (don't think it makes sense for this type of plot)
- add observation count info to tooltip
- whisker distance calculation now correct 😳 
- extend y-axis with `.nice()`
- bug fix for yscale when using beeswarm
- bug fix for observation opacity during updates
- disable tooltip for hidden observations


liquidpele:
cool beans, thanks!

PR #2105, opened 2017-09-27
[discreteBarChart] Fix y axis ticks number calculation.

### LINKED ISSUE #2104: Y axis ticks number calculation is incorrect
There's an issue with the y ticks number calculation at https://github.com/novus/nvd3/blob/master/src/utils.js#L455-L458

Live Example: https://jsfiddle.net/ovvn/g5qwwthx/1/embedded/result/

In the example when a 'zero' column is added, it should not change the number of y ticks, but it does.

ovvn:
Fixes #2104

### DIFF
--- a/src/models/discreteBar.js
+++ b/src/models/discreteBar.js
@@ -15,6 +15,7 @@ nv.models.discreteBar = function() {
         , y = d3.scale.linear()
         , getX = function(d) { return d.x }
         , getY = function(d) { return d.y }
+        , yAccessor
         , forceY = [0] // 0 is forced by default.. this makes sense for the majority of bar graphs... user can always do chart.forceY([]) to remove
         , color = nv.utils.defaultColor()
         , showValues = false
@@ -222,6 +223,7 @@ nv.models.discreteBar = function() {
         showValues: {get: function(){return showValues;}, set: function(_){showValues=_;}},
         x:       {get: function(){return getX;}, set: function(_){getX=_;}},
         y:       {get: function(){return getY;}, set: function(_){getY=_;}},
+        yAccessor: {get: function(){return yAccessor;}, set: function(_){yAccessor=_;}},
         xScale:  {get: function(){return x;}, set: function(_){x=_;}},
         yScale:  {get: function(){return y;}, set: function(_){y=_;}},
         xDomain: {get: function(){return xDomain;}, set: function(_){xDomain=_;}},
--- a/src/utils.js
+++ b/src/utils.js
@@ -450,11 +450,33 @@ nv.utils.calcTicksX = function(numTicks, data) {
 
 
 /*
-returns number of ticks to actually use on Y axis, based on chart data
-*/
-nv.utils.calcTicksY = function(numTicks, data) {
-    // currently uses the same logic but we can adjust here if needed later
+ returns number of ticks to actually use on Y axis, based on chart data
+ */
+nv.utils.calcTicksY = function(numTicks, data, yAccessor) {
+  if (yAccessor) {
+    // find max number of values from all data streams
+    var numValues = 1;
+    for (var i=0; i < data.length; i += 1) {
+      var values = data[i] && data[i].values ? data[i].values : [];
+      var maxValue;
+      for (var j=0; j < values.length; j += 1) {
+        maxValue = values[j] && values[j][yAccessor] ? values[j][yAccessor]: 0;
+        numValues = maxValue > numValues ? maxValue : numValues;
+      }
+    }
+    nv.log("Requested number of ticks: ", numTicks);
+    nv.log("Calculated max values to be: ", numValues);
+    // make sure we don't have more ticks than values to avoid duplicates
+    numTicks = numTicks > numValues ? numValues - 1 : numTicks;
+    // make sure we have at least one tick
+    numTicks = numTicks < 1 ? 1 : numTicks;
+    // make sure it's an integer
+    numTicks = Math.floor(numTicks);
+    nv.log("Calculating tick count as: ", numTicks);
+    return numTicks;
+  } else {
     return nv.utils.calcTicksX(numTicks, data);
+  }
 };
 
 
--- a/src/models/discreteBarChart.js
+++ b/src/models/discreteBarChart.js
@@ -183,7 +183,7 @@ nv.models.discreteBarChart = function() {
             if (showYAxis) {
                 yAxis
                     .scale(y)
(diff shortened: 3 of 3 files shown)


zheoreh:
+

zheoreh:
+1

fctsvirus:
+1

majvik:
+1

DeWenS:
+1

Yska:
+1

astranomus:
+1

diesel457:
+1

jeznag:
Seems like unit tests would be a good idea given large changes to calcTicksY function.

Also why the nv.log calls?

src/utils.js
```diff
@@ -450,11 +450,33 @@ nv.utils.calcTicksX = function(numTicks, data) {
...
+    // find max number of values from all data streams
+    var numValues = 1;
+    for (var i=0; i < data.length; i += 1) {
+      var values = data[i] && data[i].values ? data[i].values : [];
+      var maxValue;
+      for (var j=0; j < values.length; j += 1) {
+        maxValue = values[j] && values[j][yAccessor] ? values[j][yAccessor]: 0;
+        numValues = maxValue > numValues ? maxValue : numValues;
+      }
+    }
```
liquidpele: This seems super inefficient... so this is looking at every node on the chart?    Surely there is a better way to chose a tick count...

liquidpele:
This appears to use the same logic as this PR, so linking them:  https://github.com/novus/nvd3/pull/2028

src/utils.js
ovvn: Let's see. Incoming `data` for the `discreteBarChart` always has the following format:
```
[
  {
    "values": [
      {
        "label": "Label1",
        "value": 3,
        "series": 0
      },
      {
        "label": "Label2",
        "value": 123,
        "series": 0
      },
      {
        "label": "Label3",
        "value": 777,
        "series": 0
      },
        ...
    ]
  }
]
```
So complexity would be `O(n)` where `n = data[0].values.length`. The linear complexity is obviously efficient and very fast.

ovvn: Just to clarify, the idea behind this pull request is to calculate y axis ticks for the discrete bar chart based on y values, not on x values as it is now https://jsfiddle.net/ovvn/g5qwwthx/1/embedded/result/. We've been using the fix for half a year on production and it works great for us.

liquidpele: Yes, it's O(n), but for very large (millions) of points n is still sufficiently big to potentially cause problems in a browser.  And that's not just worst case, it's every time because you always have to traverse the entire array.  I see you're thinking about a simple bar chart, but this function is used by more than just that... if you want to override the function to calculate the tick count  just for one type of chart, then I'd keep all the logic in the chart itself.  

Also, what's the difference between getY and yAccessor options?  Seems like that's overlapping functionality?

src/models/discreteBar.js
```diff
@@ -15,6 +15,7 @@ nv.models.discreteBar = function() {
         , y = d3.scale.linear()
         , getX = function(d) { return d.x }
         , getY = function(d) { return d.y }
+        , yAccessor
```
jeznag: Seems like yAccessor should replace getY

src/utils.js
ovvn: Traversal in an array with millions of items is not a problem. What can be a problem is appending millions of elements into a browser's DOM, but that has nothing to do with the pull request. The functionality is not made for just one type of chart, it's generic and can be extended for other types as well. The difference between getY and yAccessor is that `getY` is a function and `yAccessor` is a string (y value property name). But that's a good point, don't remember why I made the property name to be passed explicitly. Think we can reuse the `getY` fn.

src/utils.js
```diff
@@ -450,11 +450,33 @@ nv.utils.calcTicksX = function(numTicks, data) {
...
 /*
-returns number of ticks to actually use on Y axis, based on chart data
-*/
-nv.utils.calcTicksY = function(numTicks, data) {
-    // currently uses the same logic but we can adjust here if needed later
+ returns number of ticks to actually use on Y axis, based on chart data
+ */
+nv.utils.calcTicksY = function(numTicks, data, getY) {
+  if (getY) {
+    // find max number of values from all data streams
```
jeznag: This comment is incorrect

src/utils.js
```diff
@@ -450,11 +450,33 @@ nv.utils.calcTicksX = function(numTicks, data) {
 
 
 /*
-returns number of ticks to actually use on Y axis, based on chart data
-*/
-nv.utils.calcTicksY = function(numTicks, data) {
-    // currently uses the same logic but we can adjust here if needed later
+ returns number of ticks to actually use on Y axis, based on chart data
+ */
+nv.utils.calcTicksY = function(numTicks, data, getY) {
```
jeznag: There should be unit tests for this function

src/utils.js
```diff
@@ -450,11 +450,33 @@ nv.utils.calcTicksX = function(numTicks, data) {
...
+  if (getY) {
+    // find max number of values from all data streams
+    var numValues = 1;
+    for (var i=0; i < data.length; i += 1) {
```
jeznag: This code fails linting guidelines

jeznag:
@liquidpele @ovvn I don't feel that this code was ready to be merged for the following reasons
1. It has no unit tests
2. The code is not formatted according to the NVD3 eslint config (https://github.com/novus/nvd3/blob/master/.eslintrc.json)
3. I don't think it has been tested on mobile phones with large datasets. The algorithm may perform poorly on low power CPUs. Should be tested with CPU throttling.
4. The comments in the code are misleading
5. The debugging code should be removed

I have left line comments in the PR pointing out these issues. Would be great to see a follow up commit from @ovvn addressing these points.

liquidpele:
@jeznag yea, I may have been a little too lenient on that one.    

@ovvn can you take care of the concerns brought up?

ovvn:
@jeznag
1. Yes, having unit tests is a good idea.
2. Yes, the indentation wasn't right.
3. When you say "large", what numbers are you talking about? You might need to read more about javascript performance and algorithm difficulty but in this particular case it will work fine regardless of cpu power. Do you have an experience working with d3 visualisations on desktop and mobile?
4. Yes, there was 1 outdated comment.
5. Don't see why the debugging code should be removed. The code is present in the 'calcTicksX' as well.

ovvn:
@liquidpele
Yes, fixed some issues in the pull request https://github.com/novus/nvd3/pull/2144.
I think unit tests for all util helpers is a good idea for future pull requests.

PR #2109, opened 2017-10-02
axis.orient('top') label spacing fix

ConstantinoSchillebeeckx:
No consideration was given to positioning the axis label in the `top` orientation - this fixes that by using the same label positioning as `bottom`, but in the negative direction.

### DIFF
--- a/src/models/axis.js
+++ b/src/models/axis.js
@@ -78,6 +78,7 @@ nv.models.axis = function() {
             var w;
             switch (axis.orient()) {
                 case 'top':
+                    xLabelMargin = axisLabelDistance + 36;
                     axisLabel.enter().append('text').attr('class', 'nv-axislabel');
                   w = 0;
                   if (scale.range().length === 1) {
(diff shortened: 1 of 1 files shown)


src/models/axis.js
```diff
@@ -89,7 +90,7 @@ nv.models.axis = function() {
                   };
                     axisLabel
                         .attr('text-anchor', 'middle')
-                        .attr('y', 0)
+                        .attr('y', -xLabelMargin)
```
liquidpele: This changes the default to be -36 instead of 0, is that intentional?  I'm kind of wary of changing the default value on everyone.

ConstantinoSchillebeeckx:
From how I understand things, it looks like the `.orient('top')` option is something that also has to be handled by the chart itself (similarly to `.rightAlignYAxis()`); I've implemented this in my [heatmap](https://github.com/ConstantinoSchillebeeckx/nvd3/tree/heatmap) chart with `.alignXAxis()`.

As it stands, when using `.orient('top')`, the label position cannot be adjusted through `.axisLabelDistance()` since `.attr('y', 0)`; in the case of the heatmap you get something like:

screen shot 2018-02-24 at 10 16 32 am

I understand being weary when it comes to changing things around for others, however it doesn't look like any chart handles the case of `.orient('top')`. 

In this fix, the user can adjust the label positioning with something like:
```
chart.xAxis
    .axisLabel("Hour")
    .axisLabelDistance(-20)
```

screen shot 2018-02-24 at 10 18 07 am

PR #2114, opened 2017-10-16
fix($tooltip): Fix issue with tooltip not hiding when scrolling

### LINKED ISSUE #1463: Tooltip doesn't hide when scrolling out of svg
video of the bug in action: https://youtu.be/65XYhICEcvw

Basically [`mouseHandler()`](https://github.com/novus/nvd3/blob/master/build/nv.d3.js#L222) isn't called on scroll events for the svgContainer. I've fiddled with `svgContainer.on("scroll", mouseHandler)`, similar to [here](https://github.com/novus/nvd3/blob/master/build/nv.d3.js#L361) and also `d3.select(window).on("scroll",mouseHandler)` (which, obviously, doesn't work when calculating `d3.mouse(window)`). 

I'll continue fiddling but thought I'd drop this here to see if this bug is known or if someone is already working on it.

jeznag:
Change to position: absolute instead of position: fixed

#1463

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -272,7 +272,7 @@ nv.models.tooltip = function() {
                    .attr("id", id)
                    .style("top", 0).style("left", 0)
                    .style('opacity', 0)
-                   .style('position', 'fixed')
+                   .style('position', 'absolute')
                    .selectAll("div, table, td, tr").classed(nvPointerEventsClass, true)
                    .classed(nvPointerEventsClass, true);
 


KaranbeerKaur:
I am facing same issue. Did anyone find any solution for the same?

liquidpele:
Thanks for contributing!

GeKVe:
This "fix" breaks tooltip when graph is below visible area of window.
tooltip positioning is done by translate relative to top. in fixed positioning window.top == top:0. in absolute positioning closest relative parent.top (in this case body.top) == top

PR #2116, opened 2017-10-24
Break labels on line feeds

josemf:
Implemented for DiscreteBarChart and MultiBarChart will break any x-axis label on existing "\n" character in the label, thus allowing the developers to control how labels will wrap based on supplied label content.

Only when wrapLabels=false.


liquidpele:
Dont' change build files, only source files.

liquidpele:
Any update?

josemf:
Same as the other PR.

Thanks.

PR #2117, opened 2017-10-24
Fix errors when zooming on data

joseprio:
While playing around with with the scale and range values to implement a zooming functionality, I encountered a few errors, so I added some additional checks in order to avoid them. It would be great to have them on the main branch!

### DIFF
--- a/src/models/scatter.js
+++ b/src/models/scatter.js
@@ -243,7 +243,7 @@ nv.models.scatter = function() {
                 // inject series and point index for reference into voronoi
                 if (useVoronoi === true) {
 
-                	
+
                     // nuke all voronoi paths on reload and recreate them
                     wrap.select('.nv-point-paths').selectAll('path').remove();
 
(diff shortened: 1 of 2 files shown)


joseprio:
I also noticed that the scatter graph tooltip wasn't working without "useVoronoi", so I fixed it.

src/models/scatter.js
```diff
@@ -396,63 +399,63 @@ nv.models.scatter = function() {
                         //.style('pointer-events', 'auto') // recativate events, disabled by css
                         .on('click', function(d,i) {
                             //nv.log('test', d, i);
-                            if (needsUpdate || !data[d.series]) return 0; //check if this is a dummy point
-                            var series = data[d.series],
+                            if (needsUpdate || !data[d[0].series]) return 0; //check if this is a dummy point
+                            var series = data[d[0].series],
```
liquidpele: Why the [0] all over?  d should be the data object, it shouldn't be an array...

joseprio: If memory serves right, the error when using useVoronoi "false" was that the series property didn't exist; when I debugged, I found out that the real value was inside an array, hence the change.

PR #2123, opened 2017-11-09
Fix legendPosition bottom on stackedAreaChart

vasturiano:
Bottom legend on stackedAreaChart doesn't accommodate for full X axis vertical spacing and overlaps legend with label.

This PR fixes it in a manner that is consistent with the other chart types that support bottom legend (lineChart, multiBarChart, multiBarHorizontalChart).

Before:
before

After:
after


liquidpele:
thanks for contributing!

PR #2135, opened 2017-12-28
Fix #1685 - Added new property `barWidth` to discreteBar so that the …

### LINKED ISSUE #1685: DiscreteBar: Add support for min/max bar width

albert-tse:
…width of each bar can be resized (using percents ie. 0.9)


liquidpele:
Please don't commit any changes under the build folder, only the source files should be changed in a PR.

avdeveloper:
@liquidpele I just created a new pull request using my personal github account here: https://github.com/novus/nvd3/pull/2136

Please mark this one as closed

PR #2136, opened 2017-12-28
Fix #1685 - add adjustable padding between each bar in discreteBar

### LINKED ISSUE #1685: DiscreteBar: Add support for min/max bar width

avdeveloper:
This minimal update will allow dev's to make the bars in discreteBar charts to be narrower

### DIFF
--- a/src/models/discreteBar.js
+++ b/src/models/discreteBar.js
@@ -7,6 +7,7 @@ nv.models.discreteBar = function() {
     //------------------------------------------------------------
 
     var margin = {top: 0, right: 0, bottom: 0, left: 0}
+        , barWidth = 0.9 // default to 90% so that there is spacing in between each bar in the chart
         , width = 960
         , height = 500
         , id = Math.floor(Math.random() * 10000) //Create semi-unique ID in case user doesn't select one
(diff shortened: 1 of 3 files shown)


src/models/discreteBar.js
```diff
@@ -216,6 +217,7 @@ nv.models.discreteBar = function() {
 
     chart._options = Object.create({}, {
         // simple options, just get/set the necessary values
+        barWidth:{get: function(){return barWidth;}, set: function(_){barWidth=_;}},
```
liquidpele: barWidth only makes sense to be between 0 and 1 here right?  So it should probably validate that here.

liquidpele:
for new options, please also add a description/example in the json object in the bottom of documentation.html

albert-tse:
Ok let me do that

src/models/discreteBar.js
avdeveloper: @liquidpele if it doesn't validate, should I just return the default value or do a console.error?

liquidpele: I would nv.log it (wrapper around console.log we use) and ignore the passed value.

liquidpele:
Any update?

avdeveloper:
oh yeah! let me update that on sunday XD

avdeveloper:
@liquidpele welp, I committed late, but at least it's on a Sunday XD

compwron:
I wish this were merged, it would be great for my use case.

PR #2140, opened 2018-01-30
Fix points position update on line chart

### LINKED ISSUE #2018: wrong data points positions after doing a transition
After doing a transition by changing the data, it seems that the data points are not at the right place anymore. It looks like the scale hasn't been updated for them.
If you remove the "duration" option, it works again, but there is no more transition on the line.
screen shot 2017-04-18 at 16 57 57

Here is the codepen: https://codepen.io/mvidailhet/pen/RVWyxo

vasturiano:
Fixes #2018.

Issue is that the scatter layer is watching a transition to synchronize point updates (`transform` attribute) [here](https://github.com/novus/nvd3/blob/master/src/models/scatter.js#L537).

But the `lineChart` (the parent container) is not initiating the call via a `transition` (like the [scatterChart](https://github.com/novus/nvd3/blob/master/src/models/scatterChart.js#L101) and other charts), therefore the point transform attributes aren't updated.

This PR fixes that and makes the `lineChart` behavior consistent with the other charts.

### DIFF
--- a/GruntFile.js
+++ b/GruntFile.js
@@ -163,7 +163,7 @@ module.exports = function(grunt) {
     grunt.loadNpmTasks('grunt-karma');
     grunt.loadNpmTasks('grunt-text-replace');
 
-    grunt.registerTask('default', ['concat', 'copy', 'postcss', 'karma:unit']);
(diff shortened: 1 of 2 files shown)


liquidpele:

Issue #1919, opened 2017-01-10
wrong yDomain for stacked bars in MultiChart


mszpulak:
![image](https://cloud.githubusercontent.com/assets/9830091/21803760/cdc375a4-d72b-11e6-99e2-a500cf51a181.png)

seems that max yDomain is calculated based on one bar - instead of stackeh bars. 

However it seems that it's working in regular multiBar chart.

http://plnkr.co/edit/ZNfeENKjZxaXgym0LuGl?p=preview

bgth:
Please check your tickformat. .tickFormat(d3.format(',.f')). You need to add .2f instead of .f. Its not an issue with nvd3.

mszpulak:
Hi,
It doesn't help. Still ydomain is  based on one bar.
Regards,m.,

bgth:
It works. [jsfiddle](http://jsfiddle.net/bgth/bdqtt73q/15/)

mszpulak:
![image](https://cloud.githubusercontent.com/assets/9830091/22459654/c828e008-e7a1-11e6-85c5-1ab2289a9c06.png)

As you see for the first bar, second ant third color are clipped out. It's because rigth yAxis has ydomain = 600, where it shall has 622

bgth:
Yeah. I take back. I was also able to reproduce it.

liquidpele:
Someone just allowed stacking in multichart recently, so probably a bug around that.   If you fix, please send us a pull request :)

mszpulak:
I've forked changes to https://github.com/mszpulak/nvd3/commit/2189cd5370556c47f2e02316509261c9bbd9ea64.
thought multichart is customized for my purposes, it's need to be generalized.

emanuele-f:
I'm experiencing a similar issue for stacked area charts inside the multichart. Here is my patch to fix stacked area charts. This basically sums up the values for the stacked charts and find the maximum value to correctly compute the scale.

This could probably be complemented to fix the stacked bars as well.

```
                 extraValue2BarStacked.push({x:0, y:0});
             }
-            
+
+            function getStackedAreaYs(series) {
+                return d3.transpose(series).map(function(x) {
+                    return x.map(function(g) {
+                        return g.y;
+                    });
+                }).map(function(x) {return d3.sum(x);});
+            }
+
             yScale1 .domain(yDomain1 || d3.extent(d3.merge(series1).concat(extraValue1BarStacked), function(d) { return d.y } ))
[...]

PR #2146, opened 2018-03-21
[Issue 1919] wrong yDomain for stacked bars in MultiChart 

qparis:
This is a fix proposal for the issue #1919


liquidpele:
Looks okay to me, thanks for contributing!

changliukean:
if you have a accumulated line with stacked bars, it won't fix the issue.
instead of using d3.sum(x), use d3.max(x)

PR #2170, opened 2018-06-28
Fix wrong yDomain for stacked area charts in MultiChart

emanuele-f:
As in #1919, this fixes stacked area charts as well.


emanuele-f:
More bugs now fixed

PR #2181, opened 2018-08-25
Add Scatter focus chart

zahid017:
Add Scatter with Focus chart with distribution


zahid017:
need to fix difference chart test issue

